PR #6, opened 2013-11-03
Lua 5.2 and cmake

soonhokong:
Hi leo,

I had two problems with Lua/cmake:
1. I had Lua 5.1.5, but FinaLua52.cmake recognized it as 5.2 and led a build failure (because `luaL_setfuncs` only exists in 5.2+)
2. When there is no Lua installed, we had a build failure at `src/shell/lua/leanlua.cpp` since `iostream` is not included.

I fixed them in this pull request and tested them on OSX and Ubuntu 12.04. Please take a look at the changes. Thank you.

Soonho


leodemoura:
Thanks, please go ahead a commit to the main branch.

PR #8, opened 2013-12-13
Two fixes for the problems that I had on OSX (with Lua-5.1)

soonhokong:
Hi Leo,

Please review the commits and pulled them. Please also feel free to edit them if you need.

Best,
Soonho


soonhokong:
I think @f90a9e9 also fixes the problem we have with lean_luahoo1 testcase on Windows -- http://borel.modck.cs.cmu.edu/CDash/viewTest.php?onlyfailed&buildid=6253

PR #13, opened 2014-06-03
chore(.travis.yml): add 'travis_wait' to doxygen build

soonhokong:
"If you have a command that doesn't produce output for
more than 10 minutes, you can prefix it with travis_wait,
 a function that's exported by our build environment."

http://docs.travis-ci.com/user/build-timeouts/


soonhokong:
Recent build fails (https://mail.google.com/mail/u/0/#inbox/1465f913242e97fd) because it took more than 10min to update the doxygen documentation. It tries to solve the problem by using `travis_wait`.

PR #139, opened 2014-09-05
add vector and category

fpvandoorn:
Here are my current versions of category.lean and vector.lean. There isn't much in either, but it's a start. It compiles with the current version, so it's updated to use "open" and "." (instead of "include" and "_"). It does use some superfluous sorry's and check's

### DIFF
--- a/library/data/category.lean
+++ b/library/data/category.lean
@@ -70,36 +70,35 @@ definition epi  {A B : ob} (f : mor A B) : Prop := ∀⦃C⦄ {g h : mor B C}, g
 
 end
 
+
+using unit
+
 section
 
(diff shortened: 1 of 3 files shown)


leodemoura:
I reorganized the commits (reason: some of them would not work when moved to the top of the commit stack). I also made minor edits (e.g., added Copyright, indented declarations inside of namespace). Here is the result:
https://github.com/leodemoura/lean/commits/floris

If you agree with the changes I will merge them in the blessed repository.

leodemoura:
BTW, I also deleted the Coq code in comments.

fpvandoorn:
Ok, that's fine. Thanks!

leodemoura:
The commit has been integrated.

PR #182, opened 2014-09-11
feat(library/data/category.lean) the definition of category doesn't depe...

fpvandoorn:
...nd on 'mor' anymore; make iso a class; add theorems


leodemoura:
I pushed your modifications to the main repository.

PR #211, opened 2014-09-26
changes in library

fpvandoorn:
I made changes in eq, sigma, category and minor changes is int and path.

### DIFF
--- a/library/algebra/category.lean
+++ b/library/algebra/category.lean
@@ -8,7 +8,7 @@ import data.unit data.sigma data.prod
 import algebra.function
 import logic.axioms.funext
 
-open eq
+open eq eq_ops
 
 inductive category [class] (ob : Type) : Type :=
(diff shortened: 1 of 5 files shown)


leodemoura:
@fpvandoorn The `category.lean` files is looking really nice. 
BTW, feel free to add extra shortcuts in the file `lean-input.el`.
I just added one (\==>) for long arrows (you may want to use it for natural transformations).
There are many arrows in unicode. We can make them convenient to use by extending `lean-input.el`.

fpvandoorn:
Thanks! I was indeed planning to ask how to do that. I can probably figure it out by looking at the other definitions. I was also planning to add some new shortcut for existing symbols (like \inv and -1 for ⁻¹)

leodemoura:
Here is the commit/change for the long arrow. 
https://github.com/leanprover/lean/commit/23960da532f88dd84812b41f9b4009cd231261a9
You can add new ones by just copying/modifying this one.
I will ask Soonho if he has ideas on how add new shortcuts in a more convenient way.
For now, we can just keep adding them to this file.

PR #237, opened 2014-10-09
split category.lean in different files

fpvandoorn:
Added a lot of stuff in the category library.
- Constructions: The slice functor, discrete categories
- More theorems about isos, isomorphism between objects is an equivalence relation
- Beginnings for limits and the Yoneda lemma


fpvandoorn:
Oops, apparently when I push again to my personal repository, it will automatically add it to the pull request. I will try to create another pull request from the commit I was intending.

PR #249, opened 2014-10-23
feat(library/hott): add closure properties for equivalences

javra:
Port features from the Coq Hott library


javra:
Ah, I just noticed I built with old binaries, will fix the issues...

leodemoura:
It has been integrated.

PR #254, opened 2014-10-24
feat(library/hott) add proof of 'adjointification'

javra:
This theorem establishes that is suffices to give a quasi-inverse to establish an equivalence.


javra:
Oh dear, I somehow got confused with my different branches _sigh_ No idea how to fix this

leodemoura:
Do you still have the problem? 
We can help. Which commits do you want in the master branch.

BTW, the command `git cherry-pick  [commit-hash]` is super useful.
We use it to selectively move commits from one branch to another.

javra:
I fixed it by creating another pull request for another local branch (#255)

2014-10-24 22:10 GMT-04:00 Leonardo de Moura notifications@github.com:

PR #255, opened 2014-10-25
feat(library/hott) add proof of 'adjointification'

javra:
Add proof for adjointification (only need to provide a quasi-inverse to construct an equivalence) and prove that the inverse of an equivalence is an equivalence.


leodemoura:
merged. BTW, I squashed the commit `fix strange commit` with the previous one.

PR #291, opened 2014-11-04
feat(library/hott) add theory about pre- and postcomposing being equivalences

javra:


leodemoura:
https://github.com/leanprover/lean/commits/master

PR #295, opened 2014-11-05
feat(library/hott) add theorems about pre- and postcomposing and ap being equivalences

javra:


leodemoura:
Some of theses changes have already been integrated: the first three.
Could you rebase your news commits with respect to blessed?

PR #300, opened 2014-11-06
feat(library/hott) add one new theorem, cleanup file

javra:
Equivalences.v from Coq-HoTT is completely ported now.


leodemoura:
done

PR #303, opened 2014-11-07
feat(library/hott) add univalence axiom

javra:
involves moving funext.lean to a new axioms folder


javra:
Feel free to squash some of the commits if you want

PR #307, opened 2014-11-07
additive and multiplicative structures

avigad:
I love the new structure command! I made a start on groups, and experimented with a new style for headers. (And yes, I am obsessive enough about these things that I am willing to revise all the files in the library, if we like it.)

Tomorrow I'll work on orders, and maybe adapt the function definition method used in nat.div to arbitrary well-founded orders. 

After that, I'll work on rings, fields, and ordered rings and fields -- and then use these structures for nat and int.

### DIFF
--- a/library/algebra/algebra.md
+++ b/library/algebra/algebra.md
@@ -7,4 +7,5 @@ Algebraic structures.
 * [relation](relation.lean)
 * [binary](binary.lean) : binary operations
 * [wf](wf.lean) : well-founded relations
+* [group](group.lean)
 * [category](category.lean)
--- a/library/algebra/group.lean
+++ b/library/algebra/group.lean
@@ -1,0 +1,330 @@
+/-
+Copyright (c) 2014 Jeremy Avigad. All rights reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+
+Module: algebra.group
+Authors: Jeremy Avigad, Leonardo de Moura
+
+Various  multiplicative and additive structures.
+-/
+
+import logic.eq logic.connectives
+import data.unit data.sigma data.prod
+import algebra.function algebra.binary
+
+open eq
+
+namespace algebra
+
+variables {A : Type}
+
+
+/- overloaded symbols -/
+
+structure has_mul [class] (A : Type) :=
+(mul : A → A → A)
+
+structure has_add [class] (A : Type) :=
+(add : A → A → A)
+
+structure has_one [class] (A : Type) :=
+(one : A)
+
+structure has_zero [class] (A : Type) :=
+(zero : A)
+
+structure has_inv [class] (A : Type) :=
+(inv : A → A)
+
+structure has_neg [class] (A : Type) :=
+(neg : A → A)
+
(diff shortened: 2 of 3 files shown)


fpvandoorn:
We're all overwhelming Leo with issues and pull requests :-)

Is "Module" in the header just the file name? And I'm glad I'm not the only perfectionist here :-)

leodemoura:
We may try to write a script to automate the conversion. Could you describe the changes you want?

BTW, we should probably add a `bundle` command to automate this pattern:

```
structure AddCommGroup :=
(carrier : Type) (struct : add_comm_group carrier)

coercion AddCommGroup.carrier
instance AddCommGroup.struct
```

avigad:
Leo: if the pull requests are getting excessive, let us know. We can block
things into bigger requests. There is more of a motivation to pull quickly
when we need to share data, and even then we can learn how to pull from
each others' repositories.

As far as the header style, I would like to favor /- -/ for block comments
over --. So instead of the header:

-- Copyright (c) ...
-- Released under ...
-- Author(s) : ...

I would like to make this a /- -/ block, and include the name of the module
(i.e. the full import name). Comments and a description of the file should
go in the same block underneath.

/-
Copyright (c) ...
Released under ..

Module: algebra.foo
Author(s): ...

A full development of foo theory, including bar and baz.
-/

Also, I used to use "--" for "chapter headers", e.g.

-- mult

---

-- add

---

But I think these are ugly, and now prefer

/- mult -/

/- add -/

We can use

/-
A more dramatic header
-/

and still use "--" for short or in-line comments.

Before the initial release, I would like to go through all the files in the
library (preferably with Floris and anyone else who want to join me),
review all the names of definitions and theorems, review notation and
organization into namespaces, clean up comments, etc. I don't mind making
changes like this when I do that.

On Thu, Nov 6, 2014 at 11:06 PM, Leonardo de Moura <notifications@github.com

leodemoura:
> Leo: if the pull requests are getting excessive, let us know.

It is ok. I integrate when I'm taking a break from what I'm doing.

> There is more of a motivation to pull quickly
> when we need to share data, and even then we can learn how to pull from
> each others' repositories.

I fetch data from your repositories. I used the following commands

```
git remote add jeremy https://github.com/avigad/lean.git
git remote add floris https://github.com/fpvandoorn/lean.git
...
```

Then, when I type

```
git fetch --all
```

I retrieve everything. This is how I integrate things.

fpvandoorn:
New header style sound good. 
A comment about the chapter headers, how do we distinguish them from actual multi-line comments, when we want to display them in a prettified way?
Yes, let's go over the library before initial release.

avigad:
A long time ago we raised the possibility of extracting nice document
summaries from the library files, but we have not talked about it since.

With auto completion and the find commands that have been (and will be)
added to the interface, I am not sure it is needed any more. Users can also
browse the library files through the markdown files we put in the directory.

If we ever do decide to generate document files from the library files,
though, we will have to invent some kind of "markup" language. Perhaps we
should defer thinking about that to a later date? It does not seem to be
critical.

PR #309, opened 2014-11-07
feat(library/algebra/order): begin theory of orders

avigad:
feat(library/algebra/order): begin theory of orders

I just noticed that Leo has a better definition of well-founded relations in logic.lean. I will later make this file consistent with that.

Leo, there is one comment in order.lean that shows a place where universe parameters have to be provided explicitly to make something type check. The problem is that class inference returns a solution with the wrong variables. I don't know if there is a nice fix, but it is worth taking a look.

Aside from a bundle command, I can think of other bells and whistles that might make structures easy to use. E.g. suppose I have shown that a type is a group. Right now the only way to show it is a commutative group is to use the use the constructor for commutative group. It might be nice to have another constructor simply take a group and a proof that the multiplication is commutative. (But this is not critical and can be postponed.)

### DIFF
--- a/library/algebra/group.lean
+++ b/library/algebra/group.lean
@@ -175,7 +175,7 @@ calc
     ... = 1 * b : mul_left_inv
     ... = b : mul_left_id
 
-theorem group.to_left_cancel_semigroup [instance] [s : group A] : left_cancel_semigroup A :=
+definition group.to_left_cancel_semigroup [instance] [s : group A] : left_cancel_semigroup A :=
 left_cancel_semigroup.mk (@group.mul A s) (@group.mul_assoc A s)
   (take a b c,
     assume H : a * b = a * c,
@@ -184,7 +184,7 @@ left_cancel_semigroup.mk (@group.mul A s) (@group.mul_assoc A s)
         ... = a⁻¹ * (a * c) : H
         ... = c : mul_cancel_inv_left)
 
-theorem group.to_right_cancel_semigroup [instance] [s : group A] : right_cancel_semigroup A :=
+definition group.to_right_cancel_semigroup [instance] [s : group A] : right_cancel_semigroup A :=
 right_cancel_semigroup.mk (@group.mul A s) (@group.mul_assoc A s)
   (take a b c,
     assume H : a * b = c * b,
@@ -229,7 +229,7 @@ calc
     ... = 0 + b : add_left_inv
     ... = b : add_left_id
 
-theorem add_group.to_left_cancel_semigroup [instance] [s : add_group A] : 
+definition add_group.to_left_cancel_semigroup [instance] [s : add_group A] : 
   add_left_cancel_semigroup A :=
 add_left_cancel_semigroup.mk (@add_group.add A s) (@add_group.add_assoc A s)
   (take a b c,
@@ -239,7 +239,7 @@ add_left_cancel_semigroup.mk (@add_group.add A s) (@add_group.add_assoc A s)
         ... = -a + (a + c) : H
         ... = c : add_cancel_inv_left)
 
-theorem add_group.to_add_right_cancel_semigroup [instance] [s : add_group A] : 
+definition add_group.to_add_right_cancel_semigroup [instance] [s : add_group A] : 
   add_right_cancel_semigroup A :=
 add_right_cancel_semigroup.mk (@add_group.add A s) (@add_group.add_assoc A s)
   (take a b c,
--- a/library/algebra/order.lean
+++ b/library/algebra/order.lean
@@ -1,0 +1,206 @@
+/-
+Copyright (c) 2014 Jeremy Avigad. All rights reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+
+Module: algebra.order
+Author: Jeremy Avigad
+
+Various types of orders. We develop weak orders (<=) and strict orders (<) separately. We also
+consider structures with both, where the two are related by 
+
+  x < y ↔ (x ≤ y ∧ x ≠ y)   (order_pair)
+  x ≤ y ↔ (x < y ∨ x = y)   (strong_order_pair)
+
+These might not hold constructively in some applications, but we can define additional structures
+with both < and ≤ as needed.
+-/
+
+import logic.eq logic.connectives
+import data.unit data.sigma data.prod
+import algebra.function algebra.binary
+
+open eq eq.ops
+
+namespace algebra
+
+variable {A : Type}
+
+
+/- overloaded symbols -/
+
+structure has_le [class] (A : Type) :=
+(le : A → A → Prop)
+
+structure has_lt [class] (A : Type) :=
+(lt : A → A → Prop)
+
+infixl `<=`   := has_le.le
+infixl `≤`   := has_le.le
+infixl `<`   := has_lt.lt
+
+definition has_le.ge {A : Type} [s : has_le A] (a b : A) := b ≤ a
+notation a ≥ b := has_le.ge a b
+notation a >= b := has_le.ge a b
+
+definition has_lt.gt {A : Type} [s : has_lt A] (a b : A) := b < a
+notation a > b := has_lt.gt a b
+
+
+/- weak orders -/
+
+structure weak_order [class] (A : Type) extends has_le A :=
+(le_refl : ∀a, le a a) 
+(le_trans : ∀a b c, le a b → le b c → le a c) 
+(le_antisym : ∀a b, le a b → le b a → a = b)
+
+theorem le_refl [s : weak_order A] (a : A) : a ≤ a := !weak_order.le_refl
+
+theorem le_trans [s : weak_order A] {a b c : A} : a ≤ b → b ≤ c → a ≤ c := !weak_order.le_trans 
+
+theorem le_antisym [s : weak_order A] {a b : A} : a ≤ b → b ≤ a → a = b := !weak_order.le_antisym
+
+structure linear_weak_order [class] (A : Type) extends weak_order A :=
+(le_total : ∀a b, le a b ∨ le b a)
+
+theorem le_total [s : linear_weak_order A] {a b : A} : a ≤ b ∨ b ≤ a := 
+!linear_weak_order.le_total
+
+
+/- strict orders -/
+
+structure strict_order [class] (A : Type) extends has_lt A :=
+(lt_irrefl : ∀a, ¬ lt a a) 
+(lt_trans : ∀a b c, lt a b → lt b c → lt a c)
+ 
+theorem lt_irrefl [s : strict_order A] (a : A) : ¬ a < a := !strict_order.lt_irrefl
+
+theorem lt_trans [s : strict_order A] {a b c : A} : a < b → b < c → a < c := !strict_order.lt_trans
+
+theorem lt_imp_ne [s : strict_order A] {a b : A} : a < b → a ≠ b :=
+assume lt_ab : a < b, assume eq_ab : a = b, lt_irrefl a (eq_ab⁻¹ ▸ lt_ab)
+
+structure wf_strict_order [class] (A : Type) extends strict_order A :=
+(wf_rec : ∀P : A → Type, (∀x, (∀y, lt y x → P y) → P x) → ∀x, P x)
+
+-- TODO: is there a way to make this type check without specifying universe parameters?
+-- The good news is that the error message was very helpful.
+theorem wf_rec_on.{u v} {A : Type.{u}} [s : wf_strict_order.{u v} A] {P : A → Type.{v}} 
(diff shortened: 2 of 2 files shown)


leodemoura:
> I just noticed that Leo has a better definition of well-founded relations in logic.lean. I will later make this file consistent with that.

It is part of the definitional package... There are good and bad news. The good news is that we can build the definitional package on top of `well_founded.fix` or `below_rec` (McBride's paper). The `well_founded.fix` is easy to use even manually. Here is Fibonacci using it:
https://gist.github.com/leodemoura/5f14a69c886b978b46ff

Remark: the `dcases_on` will be generated automatically too. It is a variation of `cases_on`. 

The bad news is that any definition that rely on proofs will get stuck because we define them as theorems. For example, `fib 3` gets stuck because it depends on `or.elim`, `lt_trans`, ...

I will add a `feval` command that ignores the `opaque` annotations. Then, at least, the user can test the definition using it. For example, she can type

`feval fib 5`

When we do that, then everything works.

McBride's approach does not seem to suffer from this issue, but his `below_rec` supports only structural induction (remark: I could have defined `fib` using it).

leodemoura:
> Leo, there is one comment in order.lean that shows a place where universe parameters have to be provided explicitly to make something type check.

I will investigate.

leodemoura:
> Right now the only way to show it is a commutative group is to use the use the constructor for commutative group. It might be nice to have another constructor simply take a group and a proof that the multiplication is commutative. (But this is not critical and can be postponed.)

This is quite useful. There are some issues due to field merging. For example, if the we "inherit" from more than one structure. One option is to always peek the value from the first parent, and ignore the other one. Another option is to force the user to provide a proof that the merged fields have the same value. We also need a name for this operation (e.g., `mk_from_parents`).

leodemoura:
Regarding  `wf_rec_on`, we have to provide the universes because the theorem type is elaborated independently of the theorem proof. The constraint on the universe associated with `P` is only produced in the proof. That is, the proof is restricting the meaning of the theorem.

We have two options: 

1- Provide the universes explicitly like you did

2- Use a `definition` instead of a theorem

avigad:
Very nice!

Remember, I was experimenting with a general recursion package, based on
what I did in nat.div:

  https://github.com/avigad/scratch/blob/master/lean/general_recursion.lean

It is a slight generalization of what you did: rather than doing a fix
point on one argument, you can assign a measure to the (tuple of)
argument(s), and it recursively computes the value of all arguments with at
most that measure.

Right now, the measure is assumed to be a nat, but I was planning to
generalize it to any well-founded order. The idea is roughly to replace my
recursion on nat with your recursion along any well-founded order.

Maybe the reason that my definition computes, while yours doesn't, is that
the recursion itself doesn't depend on any proofs. There is a "default"
return value, and the recursion just goes along computing as best as it
can. You have to prove that recursive calls are decreasing to establish the
correctness of the construction, i.e. that the desired equation holds. But
that proof stands outside the computation.

Does that make sense? These things still confuse me.

I was planning to try to generalize general_recursion.lean to arbitrary
well-founded orders, but if it would be better to wait to see what you do,
I can work on other things.

Jeremy

avigad:
Oh my goodness! Of course, the recursor should be a definition rather than
a theorem. I will fix that.

(Today I caught similar mistakes in group.lean, where I declared class
instances as theorems rather than definitions. That led to some strange
failures.)

But the induction version that comes next -- which takes values in Prop --
should be a theorem, and there I do need to give the parameter 0. (That is
o.k. with me.)

avigad:
Regarding my second-to-last comment, I see now that the reason fib gets stuck is not that the function definition depends on `or.elim` `lt.trans` ..., but the recursor for the well-founded recursion. So my method will face exactly the same problem. I need to think about this.

leodemoura:
> Regarding my second-to-last comment, I see now that the reason fib gets stuck is not that the function definition depends on or.elim lt.trans ..., but the recursor for the well-founded recursion. So my method will face exactly the same problem. I need to think about this.

There are a couple of things. Your approach is closer to the approach used by McBride's `below_rec`: you compute the value of the function for **every** value smaller than the input value. The `well_founded.fix` only computes for the ones actually used. 
Remark: `well_founded.fix` is based on the Coq well_founded recursion library file.

`well_founded.fix` gets stuck because our theorems are opaque. If we ignore the opaqueness, then everything computes. 

`well_founded` is a type class, we can build instances for pairs, lists, etc. It is quite general.

The constructions in `wf.lean` are clean and all straightforward. Last night, I was in the mode "Type theory is awesome!". Today, I'm back to "Type theory sucks".

avigad:
> Last night, I was in the mode "Type theory is awesome!". Today, I'm back
> to "Type theory sucks".

Both statements are true! Would it be helpful for me to convert my approach
to well-founded recursion? Or should we talk about it first?

leodemoura:
We should talk about it. It seems we should support multiple approaches. Each one has advantages and disadvantages. 
Here is the long email I was writing to you.

I tried the following tirck to deal with `well_founded.fix`. It is essentially a generalization of the approach we already use for equality.
The fib example gets stuck at `acc.rec F H` where `(H : acc R x)`, but `H` is a complicated term which is not `acc.intro …`.
It cannot be reduced to `acc.intro` because our theorems are opaque. If we make them transparent, then it would reduce, but it would be inefficient.
The solution I propose is a generalization of the K-axiom trick used in many systems.
Given any `(H : acc R x)`, we have that

```
    acc.intro x (fun (y : A) (H2 : R y x), acc.inv H H2)      :    acc R x
```

Since `(acc R x)` is a proposition,  `H` and the term above are definitionally equal, but the second allows the computation to continue.
So, whenever the kernel is stuck at

```
    acc.rec F H
```

it can simply replace `H` with `(acc.intro x (fun (y : A) (H2 : R y x), acc.inv H H2))`, and the computation will continue.
In general, we cannot construct these propositions. So, the plan would be to allow the user to define “canonical” elements.
For example, he would define

```
definition canonical {A : Type} {R : A → A → Prop} {x : A} (a : acc R x) : acc R x :=
acc.intro x (λ y y_lt_x, acc.inv a y_lt_x)
```

and mark it as canonical in the kernel. Then, it is automatically used whenever a term `(acc.rec F H)` is stuck.

The problem of this approach is that not all evaluation orders are guaranteed to terminate.
For example, consider the term

```
    fun (Hwf : well_founded nat.gt),
                   fix x (fun x r, r succ_x_gt_x  (succ x))
```

The type `(Hwf : well_founded nat.gt)` is uninhabited. However, if we try to normalize the nested term

```
       (fix x (fun x r, r succ_x_gt_x  (succ x))
```

we get

```
    (acc.rec (Hwf x) (fun x ac r, r succ_x_gt_x  (succ x))
```

This computation is stuck since `Hwf` is a variable. 
However, if we replace it with the canonical term above, the computation will continue.
It will continue forever.
This is the same problem we have with extensional type theory: the type checker may loop under false assumptions.
If I implement the feature above, this may happen. Of course, we can avoid it by simply adding a bound on the number of times it is applied.

A simpler approach would be: don’t add any feature. But, define a bounded version of the well-founded instance we used.
For example, the following instance allows us to perform 4 nested recursive calls.

```
definition lt.wf_k [instance] : well_founded lt :=
well_founded.intro
 (take n : nat,
    acc.intro n
      (λ n₁ H₁, acc.intro n₁
      (λ n₂ H₂, acc.intro n₂
      (λ n₃ H₃, acc.intro n₃
      (λ n₄ H₄, well_founded.apply lt.wf n₄)))))
```

We can even do better, we can define a `wf_k` transformer

```
definition wf_k {A : Type} {R : A → A → Prop} (Hwf : well_founded R) (k : nat) : well_founded R :=
well_founded.intro
(nat.rec_on k
   (λ n : A, well_founded.apply Hwf n)
   (λ (k' : nat) (f : Πa, acc R a), (λ n : A, acc.intro n (λ y H, f y))))
```

Given any `(Hwf : well_founded R)` construction, `wf_k` builds a new one that applies `acc.intro k` times before using `Hwf`.
Then, we can simply use

```
definition lt.wf_k [instance] : well_founded lt :=
wf_k lt.wf 100
```

If we want 100 nested recursive calls. 

Best,
Leo

avigad:
This sounds really interesting. It reminds me of the singleton elimination
trick used to prove (exists n, P n) -> (sigma n, P n) for decidable P. It
would be nice to be able to make use of concrete computations in the
kernel.

I will study this and think about it, and reread the McBride paper. I will
be working at home on Tuesday and free all day to talk about it -- I'll
send you an e-mail in the morning and maybe we can set up a time.

I am not wedded to the approach in general_recursion.lean, and I would be
happy to toss it in favor of something better. And I am still not clear how
the different approaches relate.

Best wishes,

Jeremy

avigad:
With the more radical approach (adding the unfolding rule to the kernel),
is there a way to limit it by making the user supply a proof that the
evaluation order terminates?

leodemoura:
>  I will be working at home on Tuesday and free all day to talk about it -- I'll
> send you an e-mail in the morning and maybe we can set up a time.

Great!

> With the more radical approach (adding the unfolding rule to the kernel),
> is there a way to limit it by making the user supply a proof that the
> evaluation order terminates?

One way to block non-termination is to apply the unfolding only for closed terms (and no axioms).
That is the kernel would perform the reduction

```
    acc.rec F H
===>
    acc.rec F (acc.intro x (fun (y : A) (H2 : R y x), acc.inv H H2))
```

only if `H` is a closed term. This is too restrictive. Usually `H` is of the form `Hwf x` where `x` is not closed, but `Hwf` is. `Hwf : well_founded R` is the proof that the relation is well founded. This will be usually a closed term synthesized using class-instance resolution.

leodemoura:
> It reminds me of the singleton elimination
> trick used to prove (exists n, P n) -> (sigma n, P n) for decidable P.

BTW, for defining `fix` using `acc`, I had to modify the inductive datatype module (the kernel). 
Otherwise, we would not be able to eliminate `acc` into `Type`. 
The problem is that `acc.intro` had type

```
acc.intro : Pi (x : A), (all y, y < x -> acc R y) -> acc R x 
```

The basic "singleton elimination" condition does not apply because `acc.intro` has an argument (`x : A`) which is not a propostion. However, `x` occurs in `acc R x`. So, we are not "revealing" any "forgotten" information by allowing the user to eliminate into `Type`. The `x` in `acc.intro` is already known, it is part of the type `acc R x`. This generalization was introduced in Coq 8.x (I don't know the precise version). They say `x` is a "non-uniform parameter". 

There are other subtle points. In Coq, the theorem corresponding to `well_founded.fix_eq` has an extra hypothesis about the extensionality of `F`. We don't need it because we have proof irrelevance in the kernel.

leodemoura:
Minor remark: maybe we should treat "theorems" such as `lt.wf` as "data". That is, we should view them as a procedure/program. It is not hard to construct `lt.wf` without using opaque theorems.
If we do that, then there is no issue. It will be like `decidable`. Eventually, we made everything transparent because our `if-then-else`'s were getting stuck.

leodemoura:
BTW, I just confirmed that well founded recursion also gets stuck in Coq if opaque theorems are used.
I also found the following remark on Chlipala's Book (http://adam.chlipala.net/cpdt/html/GeneralRec.html):

> Notice that we end these proofs with Defined, not Qed. Recall that Defined marks the theorems as transparent, so that the details of their proofs may be used during program execution. Why could such details possibly matter for computation? It turns out that Fix satisfies the primitive recursion restriction by declaring itself as recursive in the structure of Acc proofs. This is possible because Acc proofs follow a predictable inductive structure. We must do work, as in the last theorem's proof, to establish that all elements of a type belong to Acc, but the automatic unwinding of those proofs during recursion is straightforward. If we ended the proof with Qed, the proof details would be hidden from computation, in which case the unwinding process would get stuck.
> To justify our two recursive mergeSort calls, we will also need to prove that split respects the lengthOrder relation. These proofs, too, must be kept transparent, to avoid stuckness of Fix evaluation. We use the syntax @foo to reference identifier foo with its implicit argument behavior turned off. (The proof details below use Ltac features not introduced yet, and they are safe to skip for now.)

avigad:
Thanks for all the information. It makes sense to think of lt.wf as data --
it is basically a well-founded tree. Is it enough just to make it a
definition? Are there any arguments for or against making "well_founded R"
a Type instead of a Prop? (I see that acc R a has to be a Prop.)

leodemoura:
> Thanks for all the information. It makes sense to think of lt.wf as data --
> it is basically a well-founded tree. Is it enough just to make it a
> definition? Are there any arguments for or against making "well_founded R"
> a Type instead of a Prop?

If we do that, then we need functional extensionality for proving the fix point theorem.
We can avoid it because we have proof irrelevance. In `wf.lean`, I added comments where proof irrelevance is used.

leodemoura:
BTW, I have merged the modifications :-)

PR #320, opened 2014-11-14
feat(library/hott) funext_from_ua.lean almost done

javra:
What's missing is to fix the universes, but I guess we have to start fixing path.lean and trunc.lean

### DIFF
--- a/library/hott/equiv.lean
+++ b/library/hott/equiv.lean
@@ -53,7 +53,8 @@ Equiv A B
 
 namespace Equiv
 
-  definition equiv_fun [coercion] {A B : Type} (e : Equiv A B) : A → B :=
+  --Note: No coercion here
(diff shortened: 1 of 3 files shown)


leodemoura:
I get the following compilation errors when I try to compile the library

```
/home/leo/projects/lean/library/hott/trunc.lean:213:2: error: type mismatch at application
  equiv_preserves_contr f
term
  f
has type
  A ≃ B
but is expected to have type
  ?M_1 → B
/home/leo/projects/lean/library/hott/trunc.lean:229:2: error: type mismatch at application
  trunc_equiv n f
term
  f
has type
  A ≃ B
but is expected to have type
  ?M_1 → B
```

Does it work for you?

javra:
Oh, I didn't check those. I removed the coercion from Equiv to IsEquiv as Floris and I both ran into problems with it. I will adjust the other files, too.

PR #328, opened 2014-11-19
feat(library/hott) make paths universe polymorphic

javra:
complete the proof of funext_from_ua. Note that for proving funext at the moment one needs to have univalence in two consecutive universes, this should probably be changed somehow.

### DIFF
--- a/library/hott/funext_from_ua.lean
+++ b/library/hott/funext_from_ua.lean
@@ -67,91 +67,83 @@ context
   protected definition diagonal [reducible] (B : Type) : Type
     := Σ xy : B × B, pr₁ xy ≈ pr₂ xy
 
-  protected definition isequiv_src_compose {A B C : Type}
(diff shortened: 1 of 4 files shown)


avigad:
Congratulations!

Why is it a problem to have the two different univalence parameters? It
clarifies what is being used.

Jeremy

On Wed, Nov 19, 2014 at 11:20 PM, Jakob von Raumer <notifications@github.com

javra:
If it is really needed, there's no problem. But if there's proof requiring only one univalent universe, we should probably prefer that one. But the fact that Coq-HoTT also needs two univalence parameters might be a hint that there is no such proof.

One other generalization one could think of whould be to have a flexible difference in universe levels. e.g. having l+1and max(l k)+1 instead of l+1 and l+2, or even better l+1 and k+1...

PR #367, opened 2014-12-05
feat(library/hott) add categories and precategories in HoTT style

javra:
some proofs esp. tactic proofs that once worked are broken and need to be fixed


javra:
Oh wait, I think I messed up something with the merge... path.lean shouldn't change at all...

javra:
I guess everything was correct, the changes that seemed fishy to me are mostly changes in indentation in path.lean, that I cannot remember doing ;)

fpvandoorn:
Yes, I indented the stuff in the namespace of `path.lean`.  Sorry if that caused you a lot of trouble...

PR #375, opened 2014-12-13
feat(library/algebra/ordered_ring): start on ordered_ring, and minor changes elsewhere

avigad:
There is a proof that takes a long time in ordered_ring:

  https://github.com/avigad/lean/blob/master/library/algebra/ordered_ring.lean#L252-L333

I don't know what the performance bottleneck is, other than that it is somewhat long. There are no instances of calc. I wrote another version with tactics, to see if that would be faster, but it wasn't.

(By the way, my "timing" tests amount to using C-c C-l to run Lean on the file. Does it use the cache? Would it be more informative if I deleted the cache?)

The slow proof is not a problem, but I commented it in the file in case it is helpful for optimization later.

### DIFF
--- a/library/algebra/order.lean
+++ b/library/algebra/order.lean
@@ -81,7 +81,7 @@ theorem le.antisym [s : weak_order A] {a b : A} : a ≤ b → b ≤ a → a = b 
 structure linear_weak_order [class] (A : Type) extends weak_order A :=
 (le_total : ∀a b, le a b ∨ le b a)
 
-theorem le.total [s : linear_weak_order A] {a b : A} : a ≤ b ∨ b ≤ a :=
+theorem le.total [s : linear_weak_order A] (a b : A) : a ≤ b ∨ b ≤ a :=
 !linear_weak_order.le_total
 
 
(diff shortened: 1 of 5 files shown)


leodemoura:
The performance problems are due to
1- Long "coercion" chains. We have discussed this issue in the past. In the future, lean will automatically short-circuit these chains. For now, I "short-circuited" them by hand
I added 

```
  -- TODO: remove after we short-circuit class-graph
  definition linear_ordered_ring.to_mul [instance] [priority 100000] : has_mul A :=
  has_mul.mk (@linear_ordered_ring.mul A s)
  definition linear_ordered_ring.to_lt [instance] [priority 100000] : has_lt A :=
  has_lt.mk (@linear_ordered_ring.lt A s)
  definition linear_ordered_ring.to_zero [instance] [priority 100000] : has_zero A :=
  has_zero.mk (@linear_ordered_ring.zero A s)
```

2- Nested substitutions, e.g.: `!zero_mul_eq ▸ Ha⁻¹ ▸ Hab`
The previous issue aggravates this one. A term such as `a * b > 0` is not as small as we may think:

```
@has_lt.gt A
    (@order_pair.algebra.to_has_lt A (@ordered_comm_group.algebra.to_order_pair A 
       (@ordered_ring.algebra.to_ordered_comm_group A 
         (@linear_ordered_ring.algebra.to_ordered_ring A s))))
    (@has_mul.mul A
       (@mul_zero.algebra.to_has_mul A
          (@semiring.algebra.to_mul_zero A 
            (@ordered_semiring.algebra.to_semiring A (@ordered_ring.to_ordered_semiring A 
              (@linear_ordered_ring.algebra.to_ordered_ring A s)))))
       a
       b)
    (@has_zero.zero A
       (@zero_ne_one_class.algebra.to_has_zero A (@ring.algebra.to_zero_ne_one_class A 
          (@ordered_ring.algebra.to_ring A (@linear_ordered_ring.algebra.to_ordered_ring A s)))))
```

We are using higher-order unification and a lot of computation to be able to elaborate expressions such as `!zero_mul_eq ▸ Ha⁻¹ ▸ Hab`

As we discussed in the past, this kind of "rewriting" will be performed using the simplifier.

leodemoura:
>  Here is the really strange thing: compiling the file with both proofs seems faster than
>   compiling the file with just one of them.

This is not the case. Note that, lean Emacs mode is running processes in the background. 
Sometimes we make a change, and manually compile. This may produce incorrect runtime information because the emacs background processes are also running in the background processing the file.

To measure the runtime precisely, we should close emacs or disable the background processes.

leodemoura:
> Without any of the "proof ... qed" pairs, it exceeds the unifier maximum number of steps.

After the changes I made, it is faster, and no `proof ... qed` is needed.

leodemoura:
> Adding an extra "proof ... qed" around "!mul_zero_eq ▸ Hb⁻¹ ▸ Hab" fails with "value has
>   metavariables". I am not sure why

This is a known issue. 
`proof ... qed` is meant to be used with `have ..., proof ... qed` or `show ..., proof ... qed`.

leodemoura:
```
 /-
            proof -- gives invalid local context
              have H : 0 > 0, from !mul_zero_eq ▸ (eq.symm Hb) ▸ Hab,
              absurd H (lt.irrefl 0)
            qed
-/
```

This is similar to the previous issue. Right now, the ideal way to switch back to proof mode is using `show ....`. We can do it in the example above using

```
show a > 0 ∧ b > 0 ∨ a < 0 ∧ b < 0, from
   have H : 0 > 0, from !mul_zero_eq ▸ (eq.symm Hb) ▸ Hab,
   absurd H (lt.irrefl 0),
```

No `proof ... qed` is needed.

avigad:
Thanks. This information is very helpful. I will use these tricks too,
including short-circuiting the class graph. I will also use "show" to
switch back to proof mode when using tactics.

When I type C-c C-l from Emacs, how does the Lean process that is started
make use of the other processes that are running? Is it through the cache
files, or do the processes actually communicate?

Would it work if I ran Lean from a separate terminal window while Emacs is
still running?

On Sat, Dec 13, 2014 at 5:01 PM, Leonardo de Moura <notifications@github.com

leodemoura:
> When I type C-c C-l from Emacs, how does the Lean process that is started
> make use of the other processes that are running? Is it through the cache
> files, or do the processes actually communicate?

`C-c C-l` runs a completely independent process.  
By default, it does not even use the cache.

The processes in background affect the runtime because they are all memory intensive. In some sense, these processes are all "fighting" to access the memory (which is slow in comparison with the microprocessor), and the limited internal microprocessor cache.

> Would it work if I ran Lean from a separate terminal window while Emacs is
> still running?

It does, but you have to wait for the background processes to finish (and/or go into "sleep")
To be sure the background processes created by Emacs are done, I use the `top` command in the terminal. It shows the most expensive processes running in our machine. There is a column `%CPU`.

avigad:
Thanks, again this is helpful.

On Sat, Dec 13, 2014 at 6:15 PM, Leonardo de Moura <notifications@github.com

PR #379, opened 2014-12-17
refactor(library/data/int/basic): make the integers an instance of ring

avigad:
I made the integers and instance of the class comm_ring, and then "ported" facts about rings to the integers.

Some notes:

As proof of concept, it went pretty well. It is nice that basic facts about the integers are now instances of the generic algebraic theorems, with the same names.

There is still a lot more to do. I still need to do the same for the ordering on int, and also make int an integral domain with dvd and abs, and ordered_ring, etc. And I have to do the similar thing for nat.

As predicted, porting the list of theorems by hand is tedious, but workable. It will be nice to have automation for this later.

One of the theorems uncovered a possible bug. At least, I can't understand why one theorem (and only one from the list I ported!) fails to type check. Lean reports a Type mismatch, but evaluating both sides yields exactly the same term. Search on "TODO" in int/basic.lean.

Question: when porting theorems, I used "open algebra". Is there any way to "limit" the open to a section, or "close" it afterwards? It would be nice if the "open" were limited to the scope of a section. In general, I have noticed that overloading can be a real pain -- error messages, for example, become less helpful. When working with ints, it would be helpful to open nat notation very selectively.

@javra : I am afraid I renamed some theorems in group.lean:

  eq_of_neg_eq_neg -> neg.inj
  neg_zero_eq -> neg_zero
  neg_neg_eq_ -> neg_neg
  eq_of_inv_eq_inv -> inv.inj
  inv_one_eq -> inv_one
  inv_inv_eq -> inv_inv

I am afraid there will be a lot more renaming and housecleaning to come. In particular, I have to revise nat to match our naming conventions.

By the way: the naming conventions are working really well. Autocompletion is great.

### DIFF
--- a/library/data/int/basic.lean
+++ b/library/data/int/basic.lean
@@ -85,24 +85,12 @@ definition mul (a b : ℤ) : ℤ :=
         (take n, neg_of_nat (succ m * n))                -- b = of_nat n
         (take n, of_nat (succ m * succ n)))              -- b = neg_succ_of_nat n
 
-definition sub (a b : ℤ) : ℤ := add a (neg b)
-
-definition nonneg (a : ℤ) : Prop := cases_on a (take n, true) (take n, false)
-
-definition le (a b : ℤ) : Prop := nonneg (sub b a)
-
-definition lt (a b : ℤ) : Prop := le (add a 1) b
-
 /- notation -/
 
 notation `-[` n `+1]` := int.neg_succ_of_nat n    -- for pretty-printing output
 prefix - := int.neg
 infix +  := int.add
 infix *  := int.mul
-infix -  := int.sub
-infix <= := int.le
-infix ≤  := int.le
-infix <  := int.lt
 
 /- some basic functions and properties -/
 
@@ -563,7 +551,7 @@ zero_ne_one mul.comm
 Instantiate ring theorems to int
 -/
 
-section port_algebra
+context port_algebra
   open algebra    -- TODO: can we "open" algebra only for the duration of this section?
   instance comm_ring
 
@@ -608,13 +596,16 @@ section port_algebra
     @algebra.add_eq_iff_eq_neg_add _ _
   theorem add_eq_iff_eq_add_neg : ∀a b c : ℤ, a + b = c ↔ a = c + -b :=
     @algebra.add_eq_iff_eq_add_neg _ _
+
+  definition sub (a b : ℤ) : ℤ := algebra.sub a b
+  infix - := int.sub
+
   theorem sub_self : ∀a : ℤ, a - a = 0 := algebra.sub_self
   theorem sub_add_cancel : ∀a b : ℤ, a - b + b = a := algebra.sub_add_cancel
   theorem add_sub_cancel : ∀a b : ℤ, a + b - b = a := algebra.add_sub_cancel
   theorem eq_of_sub_eq_zero : ∀{a b : ℤ}, a - b = 0 → a = b := @algebra.eq_of_sub_eq_zero _ _
   theorem eq_iff_sub_eq_zero : ∀a b : ℤ, a = b ↔ a - b = 0 := algebra.eq_iff_sub_eq_zero
--- TODO: is there a bug here? see below
--- theorem zero_sub_eq : ∀a : ℤ, 0 - a = -a := algebra.zero_sub_eq
+  theorem zero_sub_eq : ∀a : ℤ, 0 - a = -a := algebra.zero_sub_eq
   theorem sub_zero_eq : ∀a : ℤ, a - 0 = a := algebra.sub_zero_eq
   theorem sub_neg_eq_add : ∀a b : ℤ, a - (-b) = a + b := algebra.sub_neg_eq_add
   theorem neg_sub_eq : ∀a b : ℤ, -(a - b) = b - a := algebra.neg_sub_eq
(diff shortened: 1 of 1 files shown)


leodemoura:
> One of the theorems uncovered a possible bug. At least, I can't understand why one theorem (and only one from the list I ported!) fails to type check. Lean reports a Type mismatch, but evaluating both sides yields exactly the same term. Search on "TODO" in int/basic.lean.

The source of the problem is that there are two `sub`s: `algebra.sub` and `int.sub`.
They are definitionally equal, however they create problems when we are solving unification constraints. The issue is not due to higher-order unification. It is just remainder that even higher-order pattern matching fragment + computation is not a trivial problem. 
I tried several work-arounds. They solve this problem, but they break other proofs in the library.
The "real fix" produces timeouts in many theorem in the library.
I have already spent 4hs on this issue. I'm giving up for now. Here are two workarounds that can be used:
1- define `int.sub` as

```
definition int.sub (a b : int) := algebra.sub a b
```

2- Control higher-unifier using `reducible` command.

```
context
  irreducible int.add
  theorem zero_sub_eq : ∀ a : int, 0 + -a = -a := algebra.zero_sub_eq
end
```

We are "helping" the unifier by saying it should not unfold `int.add`.
I'm using `context just to make the change local.

leodemoura:
> Question: when porting theorems, I used "open algebra". Is there any way to "limit" the open to a section, or "close" it afterwards?

You want to limit the scope of the notation, right?
We can use `context` for that.

```
context foo
  open nat
  check  2+3
end foo
check 2+3  -- <<<< error
```

I will fix the `section`s. It seems the move to context/section introduce a bug in the notation management.

avigad:
Oh no! I am sorry it caused so much trouble. I feel especially guilty
because my gut feeling was that defining the two separate subs was dubious,
but I rationalized that they would be definitionally equal. And the error
message convinced me that the two terms had already been elaborated, but
for some reason would not type check.

The reason I did it was that I had defined <= and < using - at the top of
the file. I could not use algebra.sub until I proved that int is an
instance of add_group.

The natural solution is to define int.sub in terms of algebra.sub later in
the file, and define <= and < after that. I did that now:

  https://github.com/avigad/lean/blob/master/library/data/int/basic.lean

Again, I am really sorry to cause so much trouble. There is a moral here;
if you want to import results from algebra, use the definitions from
algebra. I will remember to say that in the tutorial somewhere, when
writing about algebraic structures.

On Wed, Dec 17, 2014 at 6:00 PM, Leonardo de Moura <notifications@github.com

avigad:
Yes, switching to context works nicely. Is the intended behavior for
section different?

At some point I will write about sections and contexts in the tutorial. I
am trying to remember the distinctions. This is what (I think) I know:
- in context, notations are transient: they disappear when the context is
  closed. In section, they are permanent.
- parameters are only allowed in contexts. This is because in a section, a
  notation that depended on a parameter would not make sense.

Are these the only two differences?

On Wed, Dec 17, 2014 at 6:48 PM, Leonardo de Moura <notifications@github.com

leodemoura:
> Yes, switching to context works nicely. Is the intended behavior for
> section different?

No, it is not. I have to fix this issue.
The bug was introduced when we modified the `section` semantics.
The old `section` were renamed into `context`s, and `section` is just a scoping mechanism.

leodemoura:
> And the error message convinced me that the two terms had already been elaborated, but
> for some reason would not type check.

The error message is bad because of an optimization.
We apply substitutions lazily to improve performance, but we apply them fully before reporting errors. Internally we are trying to solve the following constraint when the error is produced.

```
int.rec .... t[?m4] =?= int.rec ... (int.neg a)
```

The converter tries to solve the constraints by unifying the `int.rec` arguments.
It "incorrectly" assumes both `int.rec`'s are stuck. This is not really the case if we substitute `?m4`. If we substitute `?m4`, then the first one reduces to another `int.rec ... t2`. Then, we can solve the new constraint by unifying the arguments of these two `int.rec`s.  

The `?m4` above is the term that converts `int` into an additive group.

I'm trying to improve the unifier. It is tricky, we fix this issue, and many other theorems start breaking.

avigad:
I don't think this particular case is a serious problem for the unifier.
There was an easy workaround, the new definitions are more natural, and now
I know what to avoid.

The biggest problem is that other users may come across similar problems,
and won't know what to do.

Later in the tutorial we should write a chapter on the elaborator to help
users understand what is going on. We should think about all the things we
have learned, so we can convey good advice.

On Wed, Dec 17, 2014 at 9:31 PM, Leonardo de Moura <notifications@github.com

leodemoura:
> The biggest problem is that other users may come across similar problems,
> and won't know what to do.

I'm also concerned about that. Actually, I'm very concerned.
I agree users would be completely lost in this kind of situation. 
Last night, I tried to add workarounds, and failed again. 
It feels like the higher-order unifier is an "local optimum", and any improvement would require massive re-engineering effort :-(
I will revisit this issue in the future. 

BTW, the problem you experienced would not happen if we had a command for automatically porting the theorems. It would not happen because we would bypass the elaborator/unifier.

> Later in the tutorial we should write a chapter on the elaborator to help
> users understand what is going on. We should think about all the things we
> have learned, so we can convey good advice.

I agree.

Here are some suggestions for "desperate situations".
1- provide implicit information explicitly
2- break big definitions/theorems into small declarations
3- set pretty printer options to reveal hidden information (btw, we should have `pp.all`, it is really boring to set `pp.implicit`, `pp.notation`, `pp.universes`, `pp.coercions`, ...)
4- use `print` command to inspect available class instances, coercions, overloaded notation, etc.
5- use `# namespace` (e.g., `#int`, `#nat`) to bypass overloading. 

I think we also need a way to "trace" the unifier. Like we do with the class-instance resolution.
It is quite tricky since the information would be much harder to digest.
However, it would give hints to users on how to use commands such as `irreducible` to workaround elaboration problems.
For example, when we see the execution trace for the `zero_sub_eq` problem, it gives us a hint that we can workaround the problem by using `irreducible int.add`.

avigad:
On Thu, Dec 18, 2014 at 12:26 PM, Leonardo de Moura <
notifications@github.com> wrote:

> The biggest problem is that other users may come across similar problems,
> and won't know what to do.
> 
> I'm also concerned about that. Actually, I'm very concerned.
> I agree users would be completely lost in this kind of situation.
> Last night, I tried to add workarounds, and failed again.
> It feels like the higher-order unifier is an "local optimum", and any
> improvement would require massive re-engineering effort :-(
> I will revisit this issue in the future.
> 
> I agree. In fact, I think it is probably close to the global optimum, and
> the best strategy is to build infrastructure so that diminishes the need
> for crazy unification problems.
> 
>  BTW, the problem you experienced would not happen if we had a command
> for automatically porting the theorems. It would not happen because we
> would bypass the elaborator/unifier.
> 
> Later in the tutorial we should write a chapter on the elaborator to help
> users understand what is going on. We should think about all the things we
> have learned, so we can convey good advice.
> 
> I agree.
> 
> Here are some suggestions for "desperate situations".
> 1- provide implicit information explicitly

In fact, I just confirmed that giving the class arguments explicitly would
have solved the problem:

```
theorem zero_sub_eq : ∀a : ℤ, 0 - a = -a := @algebra.zero_sub_eq int
```

comm_ring

I wonder why I did not think to try that.

2- break big definitions/theorems into small declarations

> 3- set pretty printer options to reveal hidden information (btw, we should
> have pp.all, it is really boring to set pp.implicit, pp.notation,
> pp.universes, pp.coercions, ...)
> 4- use print command to inspect available class instances, coercions,
> overloaded notation, etc.

I forgot about these. I will add them to the tutorial. I know about

  print definition
  print notation
  print classes
  print coercions

Are there others?

5- use # namespace (e.g., #int, #nat) to bypass overloading.

> I think we also need a way to "trace" the unifier. Like we do with the
> class-instance resolution.
> It is quite tricky since the information would be much harder to digest.
> However, it would give hints to users on how to use commands such as
> irreducible to workaround elaboration problems.
> For example, when we see the execution trace for the zero_sub_eq problem,
> it gives us a hint that we can workaround the problem by using irreducible
> int.add.

Sounds good. But really, don't worry about it too much. Things are working
well.

leodemoura:
> Are there others?

Yes, here is a list of them.

```
print fields algebra.group  -- display all "fields" of a structure

-- print instances <class-name>
print instances inhabited  -- display all instances of the given class
-- print notation <tokens>*
print notation + -   -- display notation using + or - tokens
print notation       -- display everything

print axioms -- display assumed axioms

print options -- display options set by user and/or emacs mode

print prefix nat.le -- display all declarations in the namespace nat.le

print coercions     -- display all coercions
print coercions num -- display only the coercions from num

print classes       -- display all classes
```

avigad:
There are really helpful. I will add the list to chapter 4.

On Thu, Dec 18, 2014 at 8:58 PM, Leonardo de Moura <notifications@github.com

PR #381, opened 2014-12-24
more housecleaning

avigad:
... and nat is now an instance of semiring. (But I still have to make nat and int instances of ordered_semiring and ordered_ring, respectively.)


avigad:
I realized that with my first pull request, we need to do "git clean -fd", to remove the folders I deleted.

Things are coming along nicely. I rewrote data.int.order entirely, make int an instance of ordered_ring, and transferred all the theorems. I also did some renaming and housecleaning in algebra.

There is one test broken, interactive/num2.input. The file num2.lean imports a non-existent module.

I have two questions. First, there are a number of auxiliary constructions in int, that are irrelevant after the integers are defined. It would be nice not to clutter up the namespace and to prevent the user from seeing them, even with "open int". The best way I can think to do this is to do this for each group of auxiliary definitions and theorems:

  namespace aux
    private definition ...
    private theorem ...
  end aux
  open aux

Does that sound like the right thing to do?

Second, I tried to make int an instance of algebra.integral domain:

  https://github.com/avigad/lean/blob/master/library/data/int/basic.lean#L605-L614

Two things were strange to me. First, after "namespace algebra", I still had to "open algebra", or the class integral domain was not visible. Second, after I declared the instance, closed algebra, and then wrote "open algebra" a second time, it didn't work: the instance I defined wasn't available, even though I wrote "persistent". Is this the intended behavior?

My current "to do" list for the number structures is here:

data.nat.order : make instance of ordered semiring
data.nat.sub : rename theorems
algebra.ordered_group : add absolute value
algebra.ordered_ring : add sign, theorems about sign and abs 
data.nat.div : rename theorems, expand
data.int.div : create
data.rat : create, define rationals, make instance of an ordered ordered_ring
algebra.field : create, and make rationals an instance
algebra.ordered_field : create, and make rationals an instance
data.real
data.complex

leodemoura:
> namespace aux
> private definition ...
> private theorem ...
> end aux
> open aux

I don't see a motivation for the `aux` namespace. The `private` keyword is enough for making them "invisible" when the module is imported.
When we use `private`, lean creates a name of the form: `private.<hashcode>.<name>`. 
The hashcode is computed using position information (line/col), file name, and the given name. 
It also create an alias from the given name `<name>` to `private.<hashcode>.<name>`.

I will investigate the other issues.

leodemoura:
> Second, I tried to make int an instance of algebra.integral domain:
> 
> https://github.com/avigad/lean/blob/master/library/data/int/basic.lean#L605-L614
> 
> Two things were strange to me. First, after "namespace algebra", I still had to "open algebra", or the
> class integral domain was not visible.

I can explain this one.
The `namespace algebra` is inside of `namespace int`. So, you are defining the "nested" namespace `int.algebra`, and it is not related with the namespace `algebra`. This is why you had to use `open algebra`. 
If we replace the lines 605-614

```
namespace algebra
--  open algebra    -- TODO: why do we have to do this?
  print classes
  instance [persistent] int.integral_domain
end algebra
```

with

```
end int
namespace algebra
--  open algebra    -- TODO: why do we have to do this?
  print classes
  instance [persistent] int.integral_domain
end algebra
namespace int
```

then everything works the way you expected.

Similarly, you had to use `instance int.integral_domain` at

```
section port_algebra
  open algebra
  instance int.integral_domain  -- TODO: why didn't the setting above persist?
```

because the instance is part of the namespace `int.algebra` and not `algebra`.

BTW, the `open` command has many modifiers. For example, `open [classes] algebra` will add only the classes/instances in `algebra` to the current scope. This is useful because it allow us to get the classes/instances without getting the notations and aliases contained in this namespace.
It also allow us to define `int.integral_domain` in the following way:

```
open [classes] algebra

protected definition integral_domain [instance] : algebra.integral_domain int :=
algebra.integral_domain.mk add add.assoc zero zero_add add_zero neg add.left_inv
add.comm mul mul.assoc (of_num 1) one_mul mul_one mul.left_distrib mul.right_distrib
zero_ne_one mul.comm @eq_zero_or_eq_zero_of_mul_eq_zero
```

avigad:
Very good. I though that "private" was private to the namespace, so that
you would only see it when you open the namespace. I didn't want the user
to see it when the opened int. Now I understand that "private" means
private to the module.

On Sun, Dec 28, 2014 at 5:21 PM, Leonardo de Moura <notifications@github.com

avigad:
Sorry, that was dumb on my part. Somehow, I forgot that I was already in
another namespace. This makes perfect sense now.

On Sun, Dec 28, 2014 at 5:58 PM, Leonardo de Moura <notifications@github.com

PR #397, opened 2015-01-20
Python 2/3 compatibility. Closes #393.

### LINKED ISSUE #393: Compatibility with Python 2/3
Lean does not compile if `python` is `python3`. I can think of two ways to increase the compatibility: (1) replace every `python` by `python2`, or (2) convert the code so that it can run under `python3` and also later versions of `python2`. (I can also submit a pull request once a consensus is reached.)

favonia:
Python 2/3 compatibility. Closes #393. Needs python-six library.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -189,6 +189,7 @@ install:
   - if [[ $REPO == BLESSED && ($UPLOAD || $BUILD_DOXYGEN == TRUE) ]]; then
        sudo apt-get -qq install python python-pip;
        sudo pip install dropbox;
+       sudo pip install six;
     fi
   - if [[ $TESTCOV == ON ]]; then
        wget http://downloads.sourceforge.net/ltp/lcov-1.10.tar.gz;
--- a/bin/leantags
+++ b/bin/leantags
@@ -5,6 +5,9 @@
 #
 # Author: Soonho Kong
 
+# Python 2/3 compatibility
+from __future__ import print_function
+
 import argparse
 import fnmatch
 import os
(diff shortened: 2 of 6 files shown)


soonhokong:
As a note, on a Ubuntu-12.04 machine, I had a problem:

First, I install `python-six` package via `apt-get`. Note that this package has the version 1.1.0-2, which is outdated. So, I had the following error:

```
Traceback (most recent call last):
  File "/usr0/home/soonhok/work/lean/src/../bin/linja", line 28, in <module>
    from six.moves import urllib
ImportError: cannot import name urllib
```

I install `six` via `pip`, which gives me the latest version 1.9.0 and it solves the problem.

soonhokong:
I'm on my way to Pittsburgh from Korea. I'll do more tests and close this one..

leodemoura:
@soonhokong It seems `six` is just a single file. I found the following at https://pypi.python.org/pypi/six

> Six supports every Python version since 2.5. It is contained in only one Python file, so it can be easily copied into your project. (The copyright and license notice must be retained.)

So, should we copy `six` into our code base? We have done that with `cpplint.py`. 
What do you think?

soonhokong:
I really like the idea! I'll add it today.

Soonho

On Tuesday, January 20, 2015, Leonardo de Moura notifications@github.com
wrote:

> @soonhokong https://github.com/soonhokong It seems six is just a single
> file. I found the following at https://pypi.python.org/pypi/six
> 
> Six supports every Python version since 2.5. It is contained in only one
> Python file, so it can be easily copied into your project. (The copyright
> and license notice must be retained.)
> 
> So, should we copy six into our code base? We have done that with
> cpplint.py.
> What do you think?
> 
> —
> Reply to this email directly or view it on GitHub
> https://github.com/leanprover/lean/pull/397#issuecomment-70709594.

## 

Soonho Kong
soonhok@cs.cmu.edu

leodemoura:
It has been integrated: https://github.com/leanprover/lean/commit/073c1d1c31f8262c8f1ae04006796215d2fdab1a

PR #403, opened 2015-01-21
add abs and sign

avigad:
I have three small points to discuss.

First, an observation about error messages: one point of difficulty is the use of numerals. For example, in the algebra library, if a theorem with "0" has an error anywhere, it is likely to be blamed on the zero ("type holder in placeholder assigned to ... none of the overloads is applicable").

This is just an instance of the usual problem with overloads and choices. One workaround it to use algebra.zero or something like that. But I thought I would mention it in case you (Leo) can think of another clever workaround.

Second, something funny is happening with numerals and substitution. E.g. here:

  https://github.com/avigad/lean/blob/master/library/algebra/ring.lean#L202

we cannot use "(one_mul a)^⁻¹" or even "{(one_mul a)^⁻¹}", and here:

  https://github.com/avigad/lean/blob/master/library/algebra/ordered_ring.lean#L366

we cannot get rid of the {}. The workarounds are easy, and I realize that it is a complicated example: it combines higher-order unification, type class inference, and insertion of a coercion.

Finally, the heuristic for "a = b : t" where we try t, t⁻¹, {t}, {t⁻¹} only seems to work when t is a single identifier, not a compound term. Is this intended? Is there any reason not to do it for compound terms as well? For example, it would let us write

  sign_of_pos H1

instead of 

  {(sign_of_pos H1)⁻¹}

here:

  https://github.com/avigad/lean/blob/master/library/algebra/ordered_ring.lean#L362

There are lots of examples of this in that file.

### DIFF
--- a/library/algebra/order.lean
+++ b/library/algebra/order.lean
@@ -290,14 +290,71 @@ section
     (assume H', absurd (H' ▸ !le.refl) H)
     (assume H', H')
 
-  theorem lt_or_ge (a b : A) : a < b ∨ a ≥ b :=
+  theorem lt_or_ge : a < b ∨ a ≥ b :=
   lt.by_cases
     (assume H1 : a < b, or.inl H1)
     (assume H1 : a = b, or.inr (H1 ▸ le.refl a))
     (assume H1 : a > b, or.inr (le_of_lt H1))
 
-  theorem le_or_gt (a b : A) : a ≤ b ∨ a > b :=
+  theorem le_or_gt : a ≤ b ∨ a > b :=
   !or.swap (lt_or_ge b a)
+
+  theorem lt_or_gt_of_ne {a b : A} (H : a ≠ b) : a < b ∨ a > b :=
+  lt.by_cases (assume H1, or.inl H1) (assume H1, absurd H1 H) (assume H1, or.inr H1)
+end
+
+structure decidable_linear_order [class] (A : Type) extends linear_strong_order_pair A :=
+(decidable_lt : decidable_rel lt)
+
+section
+  variable [s : decidable_linear_order A]
+  variables {a b c d : A}
+  include s
+  open decidable
+
+  definition decidable_lt [instance] : decidable (a < b) :=
+    @decidable_linear_order.decidable_lt _ _ _ _
+
+  definition decidable_le [instance] : decidable (a ≤ b) :=
+  by_cases
+    (assume H : a < b, inl (le_of_lt H))
+    (assume H : ¬ a < b,
+      have H1 : b ≤ a, from le_of_not_lt H,
+      by_cases
+        (assume H2 : b < a, inr (not_le_of_lt H2))
+        (assume H2 : ¬ b < a, inl (le_of_not_lt H2)))
+
+  definition decidable_eq [instance] : decidable (a = b) :=
+  by_cases
+    (assume H : a ≤ b,
+      by_cases
+        (assume H1 : b ≤ a, inl (le.antisymm H H1))
+        (assume H1 : ¬ b ≤ a, inr (assume H2 : a = b, H1 (H2 ▸ le.refl a))))
+    (assume H : ¬ a ≤ b,
+      (inr (assume H1 : a = b, H (H1 ▸ !le.refl))))
+
+  definition lt.by_cases' {a b : A} {P : Type}
+    (H1 : a < b → P) (H2 : a = b → P) (H3 : b < a → P) : P :=
+  if H4 : a < b then H1 H4 else
+    (if H5 : a = b then H2 H5 else
+      H3 (lt_of_le_of_ne (le_of_not_lt H4) (ne.symm H5)))
+
+  definition lt.by_cases'_of_lt {a b : A} {P : Type}
+      (H1 : a < b → P) (H2 : a = b → P) (H3 : b < a → P) (H4 : a < b) :
+    lt.by_cases' H1 H2 H3 = H1 H4 := !dif_pos
+
+  theorem lt.by_cases'_of_eq {a b : A} {P : Type}
+      (H1 : a < b → P) (H2 : a = b → P) (H3 : b < a → P) (H4 : a = b) :
+    lt.by_cases' H1 H2 H3 = H2 H4 :=
+  have H5 [visible] : ¬ a < b, from assume H6 : a < b, ne_of_lt H6 H4,
+  dif_pos H4 ▸ dif_neg H5
+
+  theorem lt.by_cases'_of_gt {a b : A} {P : Type}
+      (H1 : a < b → P) (H2 : a = b → P) (H3 : b < a → P) (H4 : a > b) :
+    lt.by_cases' H1 H2 H3 = H3 H4 :=
+  have H5 [visible] : ¬ a < b, from lt.asymm H4,
+  have H6 [visible] : a ≠ b, from (assume H7: a = b, lt.irrefl b (H7 ▸ H4)),
+  dif_neg H6 ▸ dif_neg H5
 end
 
 end algebra
--- a/library/algebra/order.lean
+++ b/library/algebra/order.lean
@@ -333,6 +333,22 @@ section
     (assume H : ¬ a ≤ b,
       (inr (assume H1 : a = b, H (H1 ▸ !le.refl))))
 
+  -- testing equality first may result in more definitional equalities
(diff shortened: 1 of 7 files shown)


leodemoura:
> First, an observation about error messages: one point of difficulty is the use of numerals. For example, in the algebra library, if a theorem with "0" has an error anywhere, it is likely to be blamed on the zero ("type holder in placeholder assigned to ... none of the overloads is applicable").

I see the problem. The point is that the numerals have a default interpretation (type `num`).
When we define a notation for a specific numeral, we are overloading it. 
We could add an option to disable the default interpretation for numerals.
Then, in the algebra files, `0` would have only one interpretation.
What do you think?

avigad:
Yes, I think that would help a lot, and it seems like a reasonable thing to
do.

In general (not just for numerals), is there a way to disable a notation?

On Wed, Jan 21, 2015 at 4:57 PM, Leonardo de Moura <notifications@github.com

leodemoura:
> we cannot get rid of the {}

I changed how the "calc assistant" works.
When, we write `calc lhs = rhs : pr`. The "calc assistant" first decides whether a `!` needs to be used or not (this is cheap). Then, it tries in sequence the following interpretations: `pr`, `pr⁻¹`, `{pr}`, `{pr⁻¹}`.
The problem is that in some cases, the right choice is `{pr⁻¹}`, but it takes a really long time to realize that there is no way to elaborate `{pr}`. The problem is due to the interaction of computation + higher-order unification (and missing information). 
To minimize the problem, the unifier has a "conservative" mode, where only constants marked as reducible are unfolded, other expensive cases are also omitted. Now, the calc assistant uses the conservative mode when processing `{pr}` and `{pr⁻¹}`. The idea is that the "calc assistant" covers the simple cases. 
When the user writes `{pr}`, the calc assistant is not being used, and the option "conservative" is not used unless the user have executed

```
set_option unifier.conservative true
```

You may ask why this affects an example such as `neg_zero`. The issue in this example is that `neg_zero` has an implicit argument (the structure). The argument is filled automatically using class-instance resolution. The automatically generated instances are marked as `reducible`, but the ones, you defined, are not. Thus, the unification problem cannot be solved because we cannot establish that the `0` in the `neg_zero` theorem is definitionally equal to the `0` in the theorem you are trying to prove.

leodemoura:
> In general (not just for numerals), is there a way to disable a notation?

Not really. We don't have names for notation. 
So, one option is to provide a command that disables notation containing a user provided token.
Example

```
notation [disable] +
```

would disable any notation declaration containing the token `+`.

avigad:
Thanks for the explanation of the changes to the calc assistant. They seem perfectly reasonable. The workarounds are easy, and it helps to have a sense of when they are needed. 

Just to make sure I understand: in this example,

  https://github.com/avigad/lean/blob/master/library/algebra/ring.lean#L202

I just confirmed that we _cannot_ use {(one_mul a)⁻¹}. Is that expected? I thought you said that when you give the substitution explicitly, it is not in conservative mode.

avigad:
notation [disable] + is a partial solution. Usually, we want to enable one notation when it conflicts with another. A workaround is to disable all of them and then re-declare (either manually or with "open") the one you want.

I think better management for notation should be low priority, something to think about on a rainy day.

leodemoura:
> I just confirmed that we cannot use {(one_mul a)⁻¹}. Is that expected? I thought you said that when you give the substitution explicitly, it is not in conservative mode.

The conservative mode is not use. Note that `{(one_mul a)⁻¹}` operationally equivalent to `one_mul a ▸ rfl`. The latter is executed by the calc assistant. Moreover, the calc assistant is only activated after the equation has been elaborated.

leodemoura:
> notation [disable] + is a partial solution. Usually, we want to enable one notation when it conflicts with another. A workaround is to disable all of them and then re-declare (either manually or with "open") the one you want.
> I think better management for notation should be low priority, something to think about on a rainy day.

Perhaps, we should allow the user to provide names to notations.

Another possibility, we specify a constant symbol. Example

```
notation [disable] algebra.mul
```

would disable any notation whose interpretation uses the constant `algebra.mul`.

avigad:
Hmm... I tend to favor the second option, I think. It seems easier for you to implement. For the user, it is more work to name notations, and then to remember the name of a notation. It is usually easy to guess the name of a constant.

The only downside I can think of is when we have multiple notations that mention the same symbol, like 

  notation a `**` b := a \* b \* a

Then notation [disable] algebra.mul would turn that off, too. If it is a problem we can fix it by introducing a definition for a \* b \* a.

Alternatives are to have variants of the command that disable notation whose head symbol matches "foo", or a command disable notation that exactly matches a term "t", etc.

It probably makes the most sense to start with the simple solution you described, and wait and see if anyone ever wants more than that.

I understand the issue with calculations now.

leodemoura:
I added the missing `[reducible]`s, and the problems were solved.
https://github.com/leanprover/lean/commit/ae7b5a9bc916ea31397e5e7cac1d31815baa5aff

BTW, it would be nice to have a shortcut for 

```
[instance] [coercion] [reducible]
```

avigad:
Wonderful! This is a valuable lesson -- instance declarations should be
reducible.

By the way, should we use the same tags for concrete instance? For example,
in data.int.basic we have

  protected definition integral_domain [instance] : algebra.integral_domain
int := ...

Should that have [coercion] and [reducible] too?

How about having an instance command?

  instance foo : bar := ...

could be shorthand for

  definition foo [instance] [coercion] [reducible] : bar := ...

On Wed, Jan 21, 2015 at 7:27 PM, Leonardo de Moura <notifications@github.com

leodemoura:
> Should that have [coercion] and [reducible] too?

I think so. For example, the `[reducible]` will be needed whenever we try a step such as

```
          calc    lhs = rhs :  {pr}
```

where `pr` is built using a theorem from algebra, and `lhs` and `rhs` are terms of the concrete structure. 
The unifier will need to establish that constants (`*`, `0`) from the concrete structure are definitionally equal to terms such as `has_mul.mul ...`.

leodemoura:
> How about having an instance command?
> 
>  instance foo : bar := ...

I thought about that too. The problem is that we already have a command `instance`.

```
instance foo
instance [persistent] foo
```

avigad:
Good point. Maybe "structure instance" for the three of them?

  definition foo [structure instance] : bar :=

and

  structure instance foo

I don't love it, though.

Another thought is that maybe it would be clearer to have a uniform way of
declaring attributes, like

  declare foo [instance]
  declare foo [coercion]
  declare foo [coercion] [instance]
  declare foo [reducible]

rather than "instance foo" and "coercion foo". We could use

  persistent declaration foo [instance]

for the persistent version. But this would be more work for you, and we
would have to change things all over the library.

leodemoura:
I think `[structure instance]` looks good.

> Another thought is that maybe it would be clearer to have a uniform way of
> declaring attributes, like

I think this is a really good idea. 

However, the name `declaration` may give the impression we are declaring something.
What about `set_attribute` or `attribute`?
`persistent set_attribute` looks weird. I think `persistent attribute` is nice.
What do you think?

avigad:
I like "attribute" and "persistent attribute". We can get in the habit of
calling the things in brackets "attributes".

On Thu, Jan 22, 2015 at 9:21 PM, Leonardo de Moura <notifications@github.com

leodemoura:
Great!
I will make the changes...

avigad:
Hmm... I wonder, do we need something like

  set attribute foo [coercion]
  clear attribute foo [coercion]

?

On Thu, Jan 22, 2015 at 9:34 PM, Leonardo de Moura <notifications@github.com

leodemoura:
It would be useful to have a `clear_attribute` command.
However, there are some issues:
- What would be the meaning of `clear persistent attribute`? I think this would create problems.
- Some attributes are not easy to "clear". Coercion is an example. We maintain the transitive closure of the coercion graph.
- Some `clear` commands would affect other attributes. For example: `clear attribute inhabited [class]`, must also "clear" all instance attributes. That is, if a declaration `D` has an `[instance]` attribute, and `D` is an instance of inhabited, then this `[instance]` attribute must also be cleared.

I think there are more problems, I just list the first problems that came to my mind.

avigad:
Very good point. It is not worth the effort. These issues are best handled
using namespaces and sections.

avigad:
I realized that it makes no sense to declare

```
protected definition integral_domain [instance] :
```

algebra.integral_domain int := ...

a [coercion], because it is not a function. But I will make it reducible,
and similarly for the other instances in nat and int.

PR #405, opened 2015-01-22
Python3 compat fix

favonia:
I put an extra (and wrong) colon in the end.


leodemoura:
@favonia @soonhokong 
We have another problem: style check does not work if the system does not have `six`. 
The problem is that `cpplint.py` is not in `bin` but `src/cmake/Modules/cpplist.py`.

soonhokong:
@favonia, what would be a good solution for this? What I know is to append "../../../bin/python_lib" to python path using `sys.path.append`. What do you think?

soonhokong:
@favonia, could you check the two commits - e7099ee and 421846e? If you confirm this, I'll push them to the master branch.

favonia:
@soonhokong `unicode` is not valid in Python 3. I thought `cpplint.py` was used in the configuration/compilation so I didn't notice these bugs (including the colon). I'd like to fix these scripts. Could you tell me the usage and the commands you ran to produce these errors?

soonhokong:
@favonia, I see.

> Could you tell me the usage and the commands you ran to produce these errors?

We have a build target `style` which runs `cpplint.py` over `.cpp` and `.h` files. In case you're using `ninja`, you can invoke it via `ninja style`. If you're using GNU make, you can do `make style`.

To simplify the test, you can pass any `cpp`/`.h` file to `cpplint.py` as an argument.

favonia:
This pull request should fix many known problems.  Moreover, we no longer need ``six.py``.

favonia:
This pull request should fix many known problems.  Moreover, we no longer need `six.py`.  However some builds are still failing and I'm still working on them.

leodemoura:
Ok, I will wait.

favonia:
I think it is working.  Please check.

leodemoura:
It works for me. I will merge your changes.

Many thanks,
Leo

PR #420, opened 2015-02-01
housecleaning

avigad:
Two run tests break with this, because a slightly modified dependency introduces a line "warning: imported file uses sorry". Leo, I think these are easy to fix, but if you tell me how I will do it in the future. (Is it just a matter of editing the file with the expected output?)

I renamed logic.axioms.classical to logic.axioms.prop_complete. I also renamed logic.axioms.default to logic.axioms.classical. That way, either

  import classical

or 

  import logic.axioms.classical

imports the classical axioms. I did this because a user might want to import the classical axioms without importing the whole library. 

When we import the classical axioms, the user now has "by_cases" and "by_contradiction" available at top level, like "em", rather than in namespace "decidable".

I made some partial moves to put more notation for prod and sigma at top level, but some changes need discussion. Since prod and sigma are in init, I did not want to do anything dramatic.

Ideally, I think the opening lines of logic.init.prod and logic.init.sigma should be parallel. We should have the following available at the top level:

  prod: \times, *, pair, pr1, pr2, (a, b)
  sigma: \Sigma, dpair, dpr1, dpr2, <a, b>

We should also have .1 and .2 in prod.ops and sigma.ops.

One issue is that the constructors for prod and sigma are now called prod.mk and sigma.mk. We have defined pair := @prod.mk and dpair := @sigma.mk. I am sure the internal code uses the names prod.mk and sigma.mk extensively, so I think it is o.k. to have pair and dpair as abbreviations.

Another issue is that the projections for both prod and sigma are already called pr1 and pr2. So if we define pr1 and pr2 at top level, then we will have a conflict if we open prod or sigma. Is there a way to protect the projections? If not, an alternative is to rename the internal projections proj1, proj2 (and maybe dproj1, dproj2 for sigma) -- but that probably involves changing a lot of code.

Anyhow, I made some of the changes, but here is what is left to be done:

(1) move the definition of sigma from init.sigma to init.datatypes, in parallel with int.prod, unless there is a reason that it is bad to introduce the dependency on int.datatypes.

(2) move the \times and \* notation in init.prod outside the prod namespace. (I could have done that, but I was afraid to break something.)

(3) Q: is there some way to define pr1 and pr2 at top level for prod, without introducing conflicts with "open prod" and "open sigma"?

When all these changes are merged, I will have to update the tutorial. In particular, I will remove "open prod" and "open sigma" where possible, and give examples of by_contradiction and by_cases in the section on classical logic.


avigad:
Correction: the notation for (a, b, c) and <a, b, c> are currently in the prod and sigma namespaces, respectively. Should they be moved outside?

avigad:
I just realized that, at worst, if we just define

  pr1 := @prod.pr1
  pr2 := @prod.pr2

at the top level, then at worst the symbols will be overloaded when we open sigma and prod. Maybe that is not a big problem, especially if users don't have to open sigma and prod very often.

leodemoura:
> Correction: the notation for (a, b, c) and are currently in the prod and sigma namespaces, respectively. Should they be moved outside?

Yes, I think so.

leodemoura:
> I just realized that, at worst, if we just define
> 
> pr1 := @prod.pr1
> pr2 := @prod.pr2
> 
> at the top level, then at worst the symbols will be overloaded when we open sigma and prod. Maybe
> that is not a big problem, especially if users don't have to open sigma and prod very often.

I think this will generate confusion.
Why do we need pr1 and pr2 in the top-level?
I think pattern-matching minimizes the need for these projection operators.

avigad:
Occasionally, people will want to write pr1 p. It is natural. We already do
it in the tutorial.

In that case, we can "open prod", or write prod.pr1, or "open prod.ops" and
write p.1.

Is it confusing that I added dpr1 and dpr2 at the top level for sigmas? I
can remove that.

It it a problem that if we "open prod sigma" then pr1 and pr2 are
overloaded?

leodemoura:
> In that case, we can "open prod", or write prod.pr1, or "open prod.ops" and
> write p.1.

Exactly.

> Is it confusing that I added dpr1 and dpr2 at the top level for sigmas? I
> can remove that.

Yes, I think it is confusing.

> It it a problem that if we "open prod sigma" then pr1 and pr2 are
> overloaded?

It might be, but if this is the case the user may still write `sigma.pr1` and `prod.pr1`.
The user may also use `#sigma` and `#prod` to avoid overloading.

avigad:
O.k., so here is the proposal (on top of the changes I have already made):

(1) move the definition of sigma from init.sigma to init.datatypes (this is just cosmetic)
(2) delete the definitions of dpr1, dpr2
(3) put \times and \* for prod in the top level
(4) put the notations (a, b) and <a, b> at top level

All these changes are to in init. If they seem o.k. to you, let me know, and I will make the changes and test them.

Note that we also have "pair a b" and "dpair a b" at top level.

leodemoura:
The new proposal looks good. My only concern is `*` for prod in the top level. 
Then, when we open `nat`, `algebra` or `int`, `*` will be overloaded.

avigad:
Good point. I will leave \* in the namespace.

avigad:
I made the changes. Two more tests broke, because now output is printing with notation (\times, I think, in both cases, instead of prod).

leodemoura:
I will integrate.

leodemoura:
Another detail.

```
(a, false, b, true, c) : num × (Prop × (num × (Prop × num)))
```

The notation `(...)` is right associative, but `\times` is left associative. 
I think we should change `\times` to right associative.
BTW, `\and` is right associative. So, it seems natural to have the notation for products to be right too.
On the other hand, `*` is left associative... So, if people want to use \times for vectors or even arithmetic, it will be weird ahhhhhh!!! 

Anyway, I think the following is weird and inconvenient 

```
(a, false, b, true, c) : num × (Prop × (num × (Prop × num)))
```

For example, the following produces a type error

```
definition foo : num × Prop × num := (1, true, 0)
```

leodemoura:
Ignore last message... there is something else going on...

leodemoura:
I have integrated the changes. 
I added instructions for fixing tests https://github.com/leanprover/lean/blob/master/doc/fixing_tests.md
They might be useful in the future.

avigad:
Thanks. How did you resolve the problem with products? It seems that the
same issue arises with sums.

I just checked Coq. Their product _and_ tupling operation associates to the
left:

Require Import Notations.
Require Import Logic.

Section Test.

Variables A B C : Type.
Variable p : A \* B \* C.

Check fst p. (\* has type A \* B *)

Variables (a : A) (b : B) (c : C).

Check (a, b, c).

Check fst (a, b, c). (\* has type A \* B *)

End Test.

leodemoura:
> How did you resolve the problem with products?

I was confused. The test redefined the notation `(...)`.

PR #423, opened 2015-02-03
new theorems in library/data/nat/div, and minor corrections to /src/emacs/README.md

avigad:
The emacs README file begins with a section "Requirements" that tells the user that certain packages are needed. I would like to add the following sentence to the end of that section:

"Both the optional and required packages will be installed for you automatically the first time you use `lean-mode`, if you follow the installation instructions below."

Is that o.k.? It will save users from thinking that they have to install them manually.

Also: I have been adopting the practice of writing "Lean" instead of "lean", except for the syntax of commands.

### DIFF
--- a/library/data/nat/div.lean
+++ b/library/data/nat/div.lean
@@ -5,7 +5,8 @@ Released under Apache 2.0 license as described in the file LICENSE.
 Module: data.nat.div
 Authors: Jeremy Avigad, Leonardo de Moura
 
-Definitions of div, mod, and gcd on the natural numbers.
+Definitions and properties of div, mod, gcd, lcm, coprime. Much of the development follows
+Isabelle's library.
 -/
 
 import data.nat.sub tools.fake_simplifier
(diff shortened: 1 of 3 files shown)


leodemoura:
> Is that o.k.? It will save users from thinking that they have to install them manually.

I think this is a good idea. Tom Ball made exactly this mistake.
He thought he had to install the packages manually, and spent time trying to figure out how to do it.

avigad:
I just made this change. I added "Closes #423" to the commit  message -- we'll see if that works (I have never tried it before).

soonhokong:
I agree. Furthermore, we can simplify the instructions by asking users to add only one line:

``` elisp
(load "/path/to/lean-mode/lean-setup.el")
```

while `lean-setup.el` contains what we currently have in the `README.md` file. Inside of elisp, we can detect the platform that a user is using and perform platform-specific instructions. Unless a user checked out lean source code and built it by herself, we can detect `lean-rootdir` by searching known locations (i.e.`/usr/local`, or `/usr`), so we don't need to ask them to specify it either.

leodemoura:
> I agree. Furthermore, we can simplify the instructions by asking users to add only one line: `(load "/path/to/lean-mode/lean-setup.el")`

This is a good idea. 

BTW, where is `lean-setup.el` located in the code base?

soonhokong:
> BTW, where is lean-setup.el located in the code base?

It's just my plan for now. I'll add it soon.

leodemoura:
> I just made this change. I added "Closes #423" to the commit message -- we'll see if that works (I have never tried it before).

It works :-)

Issue #422, opened 2015-02-03
tab completion in namespaces


avigad:
If you open a new file and write

```
import data.nat
open nat
check mul.as
```

and hit tab, you are offered `mul.assoc` as a completion, which is good.

But suppose you are using a namespace instead:

```
import data.nat
namespace nat
check mul.as
```

The only option is to insert `nat.mul.assoc` and delete the `nat` prefix manually.

Could tab completion strip the prefix in namespaces just as with open?

leodemoura:
@avigad Here is a related problem.
Consider the following scenario

```
import data.nat

check add.asso   -- press TAB after 'asso'
```

The auto-completion contains the `add.assoc` projection for many structures in the algebraic hierarchy.
Now, try `check add -- press TAB after 'add'`, then it is a "tsunami" of irrelevant information since there are many projections that contain the substring `add`. 
It is really hard to find anything. I essentially stopped using the auto-completion for any file that imports directly or indirectly the algebraic hierarchy.

One quick fix is to suppress "inherited projections". For example, the field `add_assoc` is defined at `add_semigroup`. So, the auto completion would suppress/hide `algebra.S.add_assoc` for any structure `S` that inherits the field `add_assoc` from `add_semigroup`. 
What do you think?

avigad:
Good point. 

It is rare that anyone should have to access structure fields directly, because we have generic theorems and notations for that. For example, in algebra.group we define `algebra.add.assoc`. The user can always apply `add.assoc`, and should never have to write `s.add_assoc`. If they want to specify the structure directly, they can type `@algebra.add.assoc _ s`, and `s` will be coerced to an additive semigroup.

I guess what I am saying is that we can suppress _all_ structure projections, unless perhaps the user actually writes the structure in the prefix.

Does that make sense?

leodemoura:
It makes sense. 
I was concerned about projections such as `sigma.pr1`, but I think you are right.

leodemoura:
@avigad I can't reproduce the issue you have described. I got `mul.assoc` (not `nat.` prefix) in both cases. Could you please confirm the issue?

BTW, in the following

```
import data.nat
open nat algebra
check mul.as
```

The auto-completion will show `nat.mul.assoc` and `algebra.mul.assoc` since `mul.assoc` is overloaded. This is the indented behavior.

avigad:
[...]

PR #425, opened 2015-02-04
feat(bin): add linja.in and LEAN_BIN_DEP cmake option

soonhokong:
see the discussion in issue #422


leodemoura:
worked!

Thanks,
Leo

PR #432, opened 2015-02-13
make lemmas about isomorphisms accept more iso parameters. new lemmas ab...

javra:
...out groupoids


soonhokong:
@javra, it seems that there is a problem in `hott/algebra/groupoid.lean` file. Could you check?

https://travis-ci.org/leanprover/lean/jobs/50582522

javra:
Ah thanks, I accidentally added some unfinished and unused theorems...

leodemoura:
I squashed the commits and pushed to blessed.

PR #542, opened 2015-04-10
small changes

fpvandoorn:
I made `is_typeof` an abbreviation instead of a definition in both libraries. Then we can insert them anywhere without affecting anything else (the change in `sigma.hlean` would otherwise break when `is_equiv_sigma_rec` was used as instance)
I also added `core.hlean`, which imports the core part of the library.


avigad:
I like the first change. It used to bother me that writing (t : T) really changed the term.

PR #549, opened 2015-04-18
the rationals

avigad:
It was really natural and pleasant to define the rationals. I love quotients. I also love "migrate".

There is only one "sorry" in rat/basic.lean, two long expressions that are equivalent up to AC rewriting. I will wait for the simplifier for that.

@rlewis1988 @soonhokong: There are lots of apparent changes to one of Rob's files, but mostly it is the fact that my emacs deleted whitespace at the end of the line. Rob, I think you can get this by choosing "customize lean-mode" from the Lean menu and toggle the value on. Soonho, can you confirm this?


soonhokong:
> Rob, I think you can get this by choosing "customize lean-mode" from the Lean menu and toggle the value on. Soonho, can you confirm this?

@avigad and @rlewis1988: Yes, I just checked that it works. @rlewis1988, find `Lean Delete Trailing Whitespace` in "customize lean-mode". You can also set this option by having the following line in your `.emacs` file (or `.emacs.d/init.el`).

``` elisp
(setq lean-delete-trailing-whitespace t)
```

robertylewis:
Ah, good to know. Thanks!

PR #556, opened 2015-04-22
another series of small changes

### LINKED ISSUE #87: Issues in existential quantification exercises
I noticed a few issues while going through the existential quantification exercises at the end of Chapter 4:

- `(∀x, p x → r) ↔ (∃x, p x) → r`

This one is actually unsolvable as written. Merely outlining the proof already results in an error:
```lean
example : (∀x, p x → r) ↔ (∃x, p x) → r :=
  iff.intro
    (assume H: (∀x, p x → r), sorry)
    (assume H: (∃x, p x) → r, sorry)
```  
Adding parentheses around the right side of the biconditional makes everything right again. I'm not sure whether this is a bug or just a typo.

- `(∃x, p x → r) ↔ (∀x, p x → r)`

This isn't valid. I think it's supposed to be `(∃x, p x → r) ↔ ((∀x, p x) → r)`?

- `(∃x, p x) ↔ (∀x, ¬p x)`  
  `(∀x, p x) ↔ (∃x, ¬p x)`

Neither of these are valid&mdash;each is missing an initial `¬` on the right side.

fpvandoorn:
The most notable change is to lower the binding power of `<->` and `↔`, so that 
`a → b ↔ c` and `a ↔ b → c` are interpreted as `(a → b) ↔ c` and `a ↔ (b → c)`, respectively. This is a [common convention](http://en.wikipedia.org/wiki/Logical_connective#Order_of_precedence). It fixes leanprover/tutorial#87. (I had to add parenthesis in library at 5 places for this, which is not much)


fpvandoorn:
EDIT: this only fixes the first bullet point of that issue, not the other two.

PR #562, opened 2015-04-27
the rationals as an ordered field

avigad:
The good news: we have the rationals and a discrete linear ordered field.

The bad news: the migrate command at the end of the file rat/order.lean fails on my computer, so I commented it out. The error messages is: "excessive memory consumption detected at 'whnf' (potential solution: increase memory consumption thresold)". I don't know how to increase the memory threshold.

(BTW: there is a typo, "thresold" -> "threshold".)

As far as the rat formalization, there is one more thing that needs to be done: write `reduce (a : prerat)`, which reduces a fraction to lowest terms. There are two uses for this:

(1) For more efficient computation, we can reduce fractions to lowest terms after computing them.
(2) We can define `numerator (a : rat) : int` and `denominator (a : rat) : int` such that 
`a = numerator a / denominator a`.

For this, I need to do some more work on int.div.

### DIFF
--- a/library/algebra/order.lean
+++ b/library/algebra/order.lean
@@ -116,7 +116,7 @@ wf.rec_on x H
 /- structures with a weak and a strict order -/
 
 structure order_pair [class] (A : Type) extends weak_order A, has_lt A :=
-(lt_iff_le_ne : ∀a b, lt a b ↔ (le a b ∧ a ≠ b))
+(lt_iff_le_and_ne : ∀a b, lt a b ↔ (le a b ∧ a ≠ b))
 
 section
   variable [s : order_pair A]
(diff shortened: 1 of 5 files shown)


leodemoura:
> The bad news: the migrate command at the end of the file rat/order.lean fails on my computer, so I commented it out. The error messages is: "excessive memory consumption detected at 'whnf' (potential solution: increase memory consumption thresold)". I don't know how to increase the memory threshold.

The problem is that `migrate` tries to unfold definitions such as `algebra.abs` and `algebra.sign` in the migrated theorems. The unfolded versions are huge.
I improved the `migrate.trace` option. Now, it also covers "out-of-memory" situations.

``` lean
set_option migrate.trace true
```

I fixed the issue by using:

``` lean
  definition abs (n : rat) : rat := algebra.abs n
  definition sign (n : rat) : rat := algebra.sign n
  migrate from algebra with rat replacing abs → abs, sign → sign
```

It is possible to perform this kind of replacement automatically.
To be able to do that, we need to migrate `algebra.abs` before we migrate any theorem that uses `algebra.abs`. The declarations are being migrated in an arbitrary order. Moreover, the command should perform the replacement only if the user did not specify one.

Issue #297, opened 2014-11-05
universe polymorphic inductive structures


fpvandoorn:
When we write 

```
inductive nat : Type := zero, succ : nat -> nat
```

we get a `nat` which lives in `Type₁`. However, when we write

```
inductive nat2 : Type₊ := zero, succ : nat2 -> nat2
```

We get `nat2.{l_1} : Type.{succ l_1}`. This second definition looks more desirable: we don't have to lift our nat in Type₁ to higher types. In the file [trunc.lean](https://github.com/fpvandoorn/lean/blob/master/library/hott/trunc.lean#L42) the definition of homotopy-n-types only accepts `A : Type₁` which sounds like a restriction which could lead to problems later on. I _think_ this is caused by the fact that `path` lives in `Type₁`.

So my question: is it possible that whenever we write `inductive foo : Type := ...` Lean will try to create a universe polymorphic `foo` for all positive universe levels (which - I think - is always the desired behavior). If this is impossible / hard / low priority, is it an idea to replace all `inductive foo : Type := ...` by `inductive foo : Type₊ := ...` in the library, so that we still get the universe polymorphic inductive types?

leodemoura:
> So my question: is it possible that whenever we write inductive foo : Type := ... Lean will 
> try to create a  universe polymorphic foo for all positive universe levels (which - I think - 
> is always the desired behavior). 

It is possible. However, I think it would overwhelm the system with universe level parameters.
For example, suppose we define a record with 3 nats. If the user doesn't say anything, he would get three universe level parameters. I see the following workarounds:

1) Define a simpler version based on the polymorphic one: `nat1 := nat.{1}`.
Perhaps, we should call them `nat := poly_nat.{1}`. 

2) The user provides explicit universes when declaring the record.

```
record point3.{l} : Type.{l} :=
(x y z : nat.{l})
```

This one is really bad for novice users that probably don't even want to know that universes exist.

3) Have two `nat` types: polymorphic and non-polymorphic. I didn't check carefully, but the file `trunc.lean` doesn't seem to need a lot of constructions from the nat library.

fpvandoorn:
[...]

PR #589, opened 2015-05-07
fundamental group of the circle

fpvandoorn:
I proved that [`(base = base) ≃ ℤ`](https://github.com/fpvandoorn/lean/blob/master/hott/hit/circle.hlean#L223-224), i.e. almost the statement that the fundamental group of the circle is the integers (I'm missing that they are equal as groups). This also required porting some files about algebra, nat and int from the standard library, and proving some additional theorems about logic, which were used in the standard library, but didn't exist in the HoTT library yet. These files should probably be ported again after they are finalized in the standard library (after we have the simplifier), but I can do that, it's not much work.

I have extended the script `script/rename.pl` and made a modified version specifically for porting: `script/port.pl`. See documentation in those files for usage.

I also made some changes in the HoTT library to eliminate some difference with the standard library, such as renaming namespace `path_algebra` to `algebra` and using `▸` for transport, instead of `▹`.

Lastly, I changed the definition of `empty` to live in `Type.{0}` instead of having it universe polymorphic. This didn't require much changes, and even simplified one proof: https://github.com/leanprover/lean/commit/ab9056c5816115cca93950850eb34de5799723ce#diff-edf78a3ab02963f43d8d1a729e2dba77. I think we should do the same for `unit`, but I couldn't do that myself, because of some built-in function of numerals (I get an error at the definition of `pos_num`).


fpvandoorn:
I've made one more commit which fixes most of the tests. The tests which still fail are

```
        682 - leanhotttest_contra1.hlean (Failed)
        699 - leanhotttest_exfalso1.hlean (Failed)
```

The issue is that the `contradiction` and `exfalso` tactics use `empty` with a universe parameter.

To clarify a bit on the reason to make `empty : Type.{0}`: I had universe errors at various places after porting files. At many places I had to use `empty.{0}` instead of `empty` anyway, because otherwise definitions had an extra (nonsensical) universe parameters. It is also in line with what we said here: https://github.com/leanprover/lean/issues/297#issuecomment-94332053.

PS: Is there a way to just re-check the tests which failed before (to double check that you solved all errors in the failed tests)?

leodemoura:
> To clarify a bit on the reason to make empty : Type.{0}

I think this is a great change. I don't see any reason to make it universe polymorphic.

PR #602, opened 2015-05-14
feat(library/tactic/inversion_tactic): consistent orientation of generated equalities

Kha:
Generated equalities in proof irrelevant environments were inverted
compared with the documentation and the proof relevant case, which
resulted in newly generated local vars replacing equivalent old ones
instead of the other way around.

### DIFF
--- a/library/init/nat.lean
+++ b/library/init/nat.lean
@@ -138,9 +138,9 @@ namespace nat
 
   theorem le.rec_on {a : nat} {P : nat → Prop} {b : nat} (H : a ≤ b) (H₁ : P a) (H₂ : ∀ b, a < b → P b) : P b :=
   begin
-    cases H with b' hlt,
+    cases H with b hlt,
       apply H₁,
-      apply H₂ b' hlt
+      apply H₂ b hlt
   end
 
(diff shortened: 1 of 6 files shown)


leodemoura:
Today, I'm going to cry of happiness :-)
Thanks a lot for going into the source code and making the necessary modifications! 
They are very welcome!

library/init/nat.lean
```diff
@@ -138,9 +138,9 @@ namespace nat
 
   theorem le.rec_on {a : nat} {P : nat → Prop} {b : nat} (H : a ≤ b) (H₁ : P a) (H₂ : ∀ b, a < b → P b) : P b :=
   begin
-    cases H with b' hlt,
+    cases H with b hlt,
```
Kha, line 141: Note that the first `with` parameter is actually irrelevant now, I've renamed it merely for consistency (as in the corresponding `hlean` file). I'll look into generally eliminating irrelevant identifiers next.

Kha:
I am glad if I can be of any help :) . Having been purely a user of theorem provers up to this point, actually working on an implementation is a very interesting endeavor for me. Oh, and thank you for the new debug print functions, they were very helpful for inspecting intermediate goals.

PR #607, opened 2015-05-18
fundamental groups, comma category and some fixes

fpvandoorn:
I proved that the fundamental group of the circle is equal the integers, as groups:

``` lean
definition fundamental_group_of_circle : π₁(S¹) = group_integers
```

Removed some `sorry`'s and added `[unfold-c]` and `[constructor]` attributes throughout the hott library (but not nearly everywhere). 

I had to fix a couple of tests because of this. Is there a way to only execute the tests which failed the previous time, to see whether you fixed them correctly? All tests pass now.


leodemoura:
> Is there a way to only execute the tests which failed the previous time, to see whether you fixed them correctly? All tests pass now.

There is an option called `--rerun-failed`, but I never used it.

```
ctest --rerun-failed
```

I know that the failures are usually the first to be executed the next time.

fpvandoorn:
Ok, I'll try that next time.

> I know that the failures are usually the first to be executed the next time.

That is also useful to know. I assumed there would be some fixed order.

PR #635, opened 2015-05-29
Reals form an ordered ring

robertylewis:
The algebraic hierarchy before became nonconstructive very early: the reals did not even form an "order_pair" under the old definitions. I've modified things so that the reals form an ordered ring (so far), and updated the proofs that nat, int, and rat form the appropriate structures.

### DIFF
--- a/library/algebra/order.lean
+++ b/library/algebra/order.lean
@@ -116,63 +116,31 @@ wf.rec_on x H
 /- structures with a weak and a strict order -/
 
 structure order_pair [class] (A : Type) extends weak_order A, has_lt A :=
-(lt_iff_le_and_ne : ∀a b, lt a b ↔ (le a b ∧ a ≠ b))
+(le_of_lt : ∀ a b, lt a b → le a b)
+(lt_of_lt_of_le : ∀ a b c, lt a b → le b c → lt a c)
+(lt_of_le_of_lt : ∀ a b c, le a b → lt b c → lt a c)
+(lt_irrefl : ∀ a, ¬ lt a a)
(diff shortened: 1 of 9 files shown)


fpvandoorn:
Very nice! It's nice to stay constructive for this long! Do you know if you can do more things constructively? For the inverse, you probably need to decide whether `x = 0` or `x ≠ 0`, right?

Also, I see that you don't use `proof_irrel` in your real files so far. That is really nice, it might mean that we can actually port the files you have so far to the HoTT library. Or do you see problems with that?

robertylewis:
As we define the algebraic hierarchy right now, I think this is the breaking point. It would be possible to refactor things again and go a bit further, by using separation instead of equality in the axioms and theorems for ordered field, etc. But this would be a significantly bigger job than what I've done here, and I'm not sure that it's worth the effort.

For porting to the HoTT library: I don't see any problems, but I'm also not completely familiar with the issues that might come up. How much of the algebraic hierarchy has been ported?

fpvandoorn:
It sounds reasonable to go classical from this point. Not using equality is probably a huge pain.

Most of the algebraic hierarchy has been ported (up to fields and up to ordered rings), a bit less has been ported relating to the number systems: only parts of nat and int have been ported.

The problematic issues with porting are the usage of proof irrelevance or classical axioms, but it looks like they aren't used so far. (other issues which potentially could come up are the lack of an impredicative universe or the fact that the `cases` tactic is slightly more powerful in the standard library). I will read up in the HoTT book how your construction relates with the construction of the Cauchy Reals in the HoTT book.

PR #653, opened 2015-06-04
add theorems from Haitao and more

avigad:
I started working in Haitao's "extras". There is a lot there! 

@hthz, I will keep going over the next couple of change. I will send you notes about what I changed (mostly renaming, and stating theorems with iff instead of =). I have also cloned your repository, and I will make my version compatible with the changes I am making.

@leodemoura, note, I made algebra/function less reliant on function extensionality. For example, it is a lot more useful to say `∀x, f (g x) = x` instead of `f ∘ g = id`. Similarly, in lists, 

```
theorem map_id' {f : A → A} (H : ∀x, f x = x) : ∀ l : list A, map f l = l
```

is generally more useful than

```
theorem map_id : ∀ l : list A, map id l = l
```

if we are trying to avoid function extensionality.

### DIFF
--- a/library/algebra/function.lean
+++ b/library/algebra/function.lean
@@ -1,7 +1,7 @@
 /-
 Copyright (c) 2014 Microsoft Corporation. All rights reserved.
 Released under Apache 2.0 license as described in the file LICENSE.
-Author: Leonardo de Moura
+Author: Leonardo de Moura, Jeremy Avigad, Haitao Zhang
 
 General operations on functions.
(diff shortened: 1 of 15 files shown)


leodemoura:
> @leodemoura, note, I made algebra/function less reliant on function extensionality. For example, it is a lot more useful to say ∀x, f (g x) = x instead of f ∘ g = id. Similarly, in lists,

I agree. This is a good change.

PR #659, opened 2015-06-08
congruences for logic, and additional theorems for div and gcd

avigad:
In the file int/gcd.lean, there is a rewrite commented out. (Search on "TODO"). Ordinarily, the file compiles in a second or two, but with the commented out rewrite, it takes 40 seconds. 

It was not hard to work around it, but I have no idea why the rewrite is problematic. I left it in the comment, in case it is a bug.

### DIFF
--- a/library/algebra/ordered_ring.lean
+++ b/library/algebra/ordered_ring.lean
@@ -154,6 +154,36 @@ section
     (assume H2 : a ≤ 0,
       have H3 : a * b ≤ 0, from mul_nonpos_of_nonpos_of_nonneg H2 H1,
       not_lt_of_ge H3 H)
+
+  theorem nonneg_of_mul_nonneg_left (H : 0 ≤ a * b) (H1 : 0 < a) : 0 ≤ b :=
+  le_of_not_gt
+    (assume H2 : b < 0,
+      not_le_of_gt (mul_neg_of_pos_of_neg H1 H2) H)
+
+  theorem nonneg_of_mul_nonneg_right (H : 0 ≤ a * b) (H1 : 0 < b) : 0 ≤ a :=
+  le_of_not_gt
+    (assume H2 : a < 0,
+      not_le_of_gt (mul_neg_of_neg_of_pos H2 H1) H)
+
+  theorem neg_of_mul_neg_left (H : a * b < 0) (H1 : 0 ≤ a) : b < 0 :=
+  lt_of_not_ge
+    (assume H2 : b ≥ 0,
+      not_lt_of_ge (mul_nonneg H1 H2) H)
+
+  theorem neg_of_mul_neg_right (H : a * b < 0) (H1 : 0 ≤ b) : a < 0 :=
+  lt_of_not_ge
+    (assume H2 : a ≥ 0,
+      not_lt_of_ge (mul_nonneg H2 H1) H)
+
+  theorem nonpos_of_mul_nonpos_left (H : a * b ≤ 0) (H1 : 0 < a) : b ≤ 0 :=
+  le_of_not_gt
+    (assume H2 : b > 0,
+      not_le_of_gt (mul_pos H1 H2) H)
+
+  theorem nonpos_of_mul_nonpos_right (H : a * b ≤ 0) (H1 : 0 < b) : a ≤ 0 :=
+  le_of_not_gt
+    (assume H2 : a > 0,
+      not_le_of_gt (mul_pos H2 H1) H)
 end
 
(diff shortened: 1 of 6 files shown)


avigad:
P.S. I tried "set_option rewrite.trace true" but that had no effect.

leodemoura:
I found the performance problem. The `rewrite` tactic always try to apply reflexivity.
The idea is to try to close trivial goals. After we apply

``` lean
rewrite [-gcd_abs_abs a₂ b₂],
```

The conclusion is:

``` lean
divide (abs a₂) (gcd (abs a₂) (abs b₂)) = divide (abs a₁) (gcd (abs a₁) (abs b₁))
```

When Lean tries to apply reflexivity, it tries to check whether the left and right hand sides are definitionally equal. It turns out this definitionally equality test is super expensive, and fails in the end.
The problem is a feature of the type checker.
Whenever it is checking `f t =?= f s`, it first tries to check whether `t =?= s`, if it fails it unfolds `f` and tries again. This feature is essential for performance reasons, but it "backfires" in this case.
It generates more than 2 million steps of this kind. 
Recall that Georges mentioned this problem in March. He said Coq uses a similar trick, and it created many problems for him in the past. I even reported it in the discussion list (item 4 in the following message):
https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!msg/lean-discuss/oJXwW5wT2Ds/d1Dgr8B-pE0J
It is funny because I made the following comment in the end of this message:

>  In practice, it is also not clear whether (instances of) Georges’ example occurs in our library.

Well, now we know it does :-)
The good news is that there is an easy solution. We just need to cache failed attempts of applying this feature.

leodemoura:
BTW, the performance issue was also affecting the following definition
https://github.com/leanprover/lean/blob/e59400b58c1f962c35078c70a9a5506c720ebaa0/hott/algebra/category/functor.hlean#L168
The elaboration times for this definition:
before 496189fe:  2.16 secs
after    496189fe:  0.08 secs

avigad:
I just realized why this performance issue came up in connection with `gcd`. The reason is that we currently define `gcd with

``` lean
private definition pair_nat.lt.wf : well_founded pair_nat.lt :=
intro_k (measure.wf pr₂) 20  -- we use intro_k to be able to execute gcd efficiently in the kernel
```

So `gcd` will unfold 20 steps with any arguments at all (even variables). I think we used `intro_k` mainly to support computation with closed terms, but it seems undesirable to have this computation happen in general (despite the fix). Removing the `intro_k` doesn't break anything, so I will remove it for now.

leodemoura:
> So gcd will unfold 20 steps with any arguments at all (even variables).

Are you sure it is unfolding 20 steps during the `is_def_eq` test? Recall that during `is_def_eq` we perform lazy delta reduction (i.e., we expand on demand). If Lean is really expanding it 20 times, I think it is quite impressive we can answer the query in 0.08 secs since the expanded term is quite big.

avigad:
Maybe not. The situation is this: the rewriter is trying to determine whether two terms `gcd a b` and `gcd c d` are definitionally equal, but `a` `b` `c` and `d` are distinct. We have

``` lean
definition gcd.F (p₁ : nat × nat) : (Π p₂ : nat × nat, p₂ ≺ p₁ → nat) → nat :=
prod.cases_on p₁ (λx y, nat.cases_on y
  (λ f, x)
  (λ y₁ (f : Πp₂, p₂ ≺ (x, succ y₁) → nat), f (succ y₁, x mod succ y₁) !gcd.lt.dec))
```

Now it looks to me that after the first expansion, the first clause of the `prod.cases_on` is enough to determine that the terms are different, so there shouldn't be more than one layer of unfolding.

Sorry about the false alarm. I'll leave `intro_k` alone.

leodemoura:
I was investigating #705, and it is related to this issue.
When Lean is trying to assign `?M l_1 ... l_n := t`, it checks whether `t` contains `?M`, and/or a free variable not included in `l_1 ... l_n`. If it contains, then the assignment is not valid. 
Then, it tries to "recover" by trying to normalize term nested in `t`. This helps when `t` is of the form `pr_1 (mk a ?M)` after we reduce `?M` disappears from `t`. 
https://github.com/leanprover/lean/blob/master/src/library/unifier.cpp#L773
Well, this call to `normalize` is not `lazy`. It will try really hard to get rid of the offending term, and Lean hangs.  
There is an easy fix, but I wanted to tell you that your comment helped me to quickly identify the problem at #705

avigad:
That was good luck -- I am glad it helped.

PR #663, opened 2015-06-10
feat(library/rat/basic.lean): add reduce for rat, and num and denom

avigad:
Note that many of the changes in the file comparisons are "noise": when I made minor changes to Rob's files, it deleted whitespace at the end of his lines (and actually resulted in merge conflicts when I tried to rebase).

@rlewis1988, could you enable the "delete trailing whitespace"?


robertylewis:
Sorry- I thought I had done that before. Looks like the setting didn't save or something. I think I've fixed it now.

By the way, you've readded abs_add_three and add_le_add_three to real/basic. I moved those to algebra yesterday.

avigad:
O.k., I will fix those. (Leo, hold off on the merge.)

avigad:
O.k., it should be fixed now.

PR #681, opened 2015-06-17
feat(library/logic/eq.lean): add C-rule for eq

avigad:
I'll use / discuss these in a new tutorial chapter on equality and dependent types.

### DIFF
--- a/library/logic/eq.lean
+++ b/library/logic/eq.lean
@@ -11,6 +11,17 @@ open eq.ops
 namespace eq
   variables {A B : Type} {a a' a₁ a₂ a₃ a₄ : A}
 
+  -- the "C rule" for eq, which also contracts the equality proof to refl
+  protected theorem rec' {C : Π (a' : A), a = a' → Type}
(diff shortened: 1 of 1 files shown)


leodemoura:
We already have `eq.drec` and `eq.drec_on`:
https://github.com/leanprover/lean/blob/master/library/init/logic.lean#L44-L51
The `subst` tactic uses `eq.drec`.
https://github.com/leanprover/lean/blob/master/src/library/tactic/subst_tactic.cpp#L113

BTW, I regret the decision of generating non-dependent eliminators for inductive predicates.
This has proved to be really annoying. In several parts of the code I have to "branch" on that.
For the Boolean connectives, it is nice use the non-dependent eliminator (e.g., `or.elim`), but we could have defined them on top of the dependent one. 
I'm not sure how big is the change, but we should make it in the future. 
The source code would be more uniform.

avigad:
Ah, somehow I missed that. drec and drec_on are better names anyhow. Thanks.

PR #688, opened 2015-06-23
Maintenance in reals

robertylewis:
There are various things left to do in the reals, including embedding the rationals. This was done haphazardly before the proof of completeness; I've moved some of the relevant definitions and theorems in order to do it earlier.

This created some minor issues (rewrite was getting confused between +, \* defined on R and +, \* inferred from algebra), so I figured it was time to migrate theorems from algebra to R. This is not done yet, and I'll be traveling for the next couple weeks. This commit is so Jeremy can make changes to min/max in algebra without interfering with my changes. I'll clean things up when I get back to Pittsburgh!

### DIFF
--- a/library/algebra/ordered_field.lean
+++ b/library/algebra/ordered_field.lean
@@ -71,8 +71,7 @@ section linear_ordered_field
     iff.intro
     (assume H : 1 < a / b,
       calc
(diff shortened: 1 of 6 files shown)


leodemoura:
subsumed by #697

PR #697, opened 2015-06-24
migrate real theorems from algebra

avigad:
I hope I did this right: I pulled from @rlewis1988, and added a commit. So this pull request includes his commits.

The migrate went smoothly. There are two things that were finicky:
- I had to remember to declare the previous instances with each successive migrate.
- I had to remember to make sure to have local definitions of each definition in algebra. 

For example, I had trouble because gt and ge were not defined in real. The migrate silently tried to use rat.gt and rat.ge and didn't tell me there was a problem. @leodemoura plans to improve migrate so that the second point won't be an issue. (I am not sure there is a good solution to the first, but it is not so bad.)

The good news is that modulo the finickiness, the migrate worked beautifully. The files all compile very quickly now on my laptop: basic, order, division, and complete require 2, 4, 6, and 1 seconds, respectively.

Rob, I did not define the coercions from numerals (so I used "zero" and "one" for real 0 and 1).

Also, a small thing: I have been following the convention that any arguments that appear in a hypothesis to a theorem are made implicit, for example, `le_of_lt {x y : real} (H : x < y) : x <= y`. I changed this in `le_of_lt` and two surrounding theorems, so that they agree with the instances in algebra, nat, int, etc. But I did not check through your files to see if there are other arguments that should also be made implicit.

### DIFF
--- a/library/algebra/ordered_field.lean
+++ b/library/algebra/ordered_field.lean
@@ -71,8 +71,7 @@ section linear_ordered_field
     iff.intro
     (assume H : 1 < a / b,
       calc
-        b   = b           : refl
-        ... < b * (a / b) : lt_mul_of_gt_one_right Hb H
+        b   < b * (a / b) : lt_mul_of_gt_one_right Hb H
         ... = a           : mul_div_cancel' Hb')
     (assume H : b < a,
       have Hbinv : 1 / b > 0, from  div_pos_of_pos Hb, calc
--- a/library/data/data.md
+++ b/library/data/data.md
@@ -15,6 +15,7 @@ Basic types:
 * [int](int/int.md) : the integers
 * [rat](rat/rat.md) : the rationals
 * [pnat](pnat.lean) : the positive natural numbers
+* [real](real.lean) : the real numbers 
 
 Constructors:
 
--- a/library/data/real/basic.lean
+++ b/library/data/real/basic.lean
@@ -103,6 +103,7 @@ theorem factor_lemma_2 (a b c d : ℚ) : (a + b) + (c + d) = (a + c) + (d + b) :
 
 --------------------------------------
 -- define cauchy sequences and equivalence. show equivalence actually is one
+namespace s
 
 notation `seq` := ℕ+ → ℚ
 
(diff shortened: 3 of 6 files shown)


robertylewis:
Cool! Thanks for doing the dirty work for me...

> The good news is that modulo the finickiness, the migrate worked beautifully. The files all compile very quickly now on my laptop: basic, order, division, and complete require 2, 4, 6, and 1 seconds, respectively.

You didn't happen do do a before/after comparison, did you? This isn't significantly different (perhaps even longer) than the times I remember seeing when I compiled the files directly.

> Rob, I did not define the coercions from numerals (so I used "zero" and "one" for real 0 and 1).

This is at the top of my list once I'm back -- I had just started to rearrange things to simplify defining the coercions.

> Also, a small thing: I have been following the convention that any arguments that appear in a hypothesis to a theorem are made implicit, for example, le_of_lt {x y : real} (H : x < y) : x <= y. I changed this in le_of_lt and two surrounding theorems, so that they agree with the instances in algebra, nat, int, etc. But I did not check through your files to see if there are other arguments that should also be made implicit.

I've been following that convention as well (apparently, not perfectly). Since I've been using mostly tactic proofs, I didn't notice those were different. There shouldn't be many other cases of this, but I'll look later.

(Hello from Christchurch, by the way!)

avigad:
Sounds good. I just compared your last commit with mine using the linux "time" command with nothing else running. Here is what I got:

(old / new)
basic: 2.0 / 2.0
order: 2.9 / 2.4
division: 1.8 / 2.5
complete: 2.0 / 0.7

I don't know why order got faster. I guess division takes longer because I now do a migrate there. The significant comparison is complete, which used to use the theorems from algebra, and now uses the migrated theorems in real. Remember, you commented out the last few theorems in complete, so the improvement is even more than that shown.

I am glad to hear you made it to New Zealand safely. I'm looking forward to hearing about your travels there.

robertylewis:
Oh, great! Those numbers sound much better than the 2/4/6 sec from your first post. Definitely an improvement.

PR #706, opened 2015-06-26
Suspensions and hits with 2-constructors

fpvandoorn:
Two major additions:
- A theorem which states that the suspension is left adjoint to the loop space in the 'category' of pointed types.
- The construction of hits with 2-dimensional path constructors.


fpvandoorn:
Oops. I still have to fix some tests

leodemoura:
I also got compilation errors for hott/types/nat/hott.hlean

fpvandoorn:
I fixed the tests.
I think I forgot to build the whole library again after I rebased my commits on your latest commits. In `hott/types/nat/hott.hlean` there is a problem with `induction (lt.trichotomy a b) with H' H'`, while `generalize (lt.trichotomy a b), intro X, induction X with H' H'` seems to work. I think this is caused by one of the changes of today, and I'm worried it's because of the change I made to `unit` (although `induction_tactic.cpp` doesn't contain the string `unit`). Sorry if I caused it.

I added a commit which works around the above issue, but probably it's better to fix the underlying issue.

Issue #347, opened 2014-11-26
automatically generate `direct_subterm` relation


leodemoura:
and well founded theorem for it

PR #716, opened 2015-06-29
implement min and max

avigad:
I implemented a `lattice` class in algebra, and other subclasses, to define `min` and `max`. There are theorems relating them to an additive group structure, but I still have to prove theorems for ordered rings, e.g. `max (a * b) (a * c) = a * max b c` when `a >= 0`.

The types `nat`, `int`, `rat`, and `real` are instances, so we have `min` and `max` for all of them uniformly. Handling `nat` was a little tricky, because `min` and `max` are defined in `init`, so we cannot use the generic definitions from algebra.

I also got rid of an unused class in `order.lean`.

@htzh These should include at least some of theorems you wanted for min and max. Please let me know if there are others you need.

@fpvandoorn Unfortunately, this makes the standard library diverge from the hott library. It would be really tedious to track all the changes and make them again in hott. Perhaps when we are both back in Pittsburgh and the concrete number types in the standard library have settled, we can do another port? I think the concrete number types are close to settling down. (A note for the port: I renamed the theorems `max_eq_right` and `max_eq_left` in `init/nat` to `max_eq_right'` and `max_eq_left'`, because those theorems mean something slightly different (and more natural) in the algebra.)

@rlewis1988 I defined `abs a b := max a (-a)` which is also valid constructively, and I moved a number of theorems to a class that can be instantiated constructively. One thing I could not prove was `abs a b >= 0`, even though that is true of the constructive reals. I also don't know how to prove the triangle inequality constructively in a general algebraic setting. We should talk about this, and find a good constructive axiomatization, to which we can instantiate the reals.

### DIFF
--- a/library/algebra/group_power.lean
+++ b/library/algebra/group_power.lean
@@ -96,38 +96,38 @@ open nat int algebra
 
 definition ipow (a : A) : ℤ → A
 | (of_nat n) := a^n
-| -[n +1]    := (a^(nat.succ n))⁻¹
+| -[1+n]     := (a^(nat.succ n))⁻¹
 
 private lemma ipow_add_aux (a : A) (m n : nat) :
-  ipow a ((of_nat m) + -[n +1]) = ipow a (of_nat m) * ipow a (-[n +1]) :=
+  ipow a ((of_nat m) + -[1+n]) = ipow a (of_nat m) * ipow a (-[1+n]) :=
 or.elim (nat.lt_or_ge m (nat.succ n))
   (assume H : (#nat m < nat.succ n),
     assert H1 : (#nat nat.succ n - m > nat.zero), from nat.sub_pos_of_lt H,
     calc
-      ipow a ((of_nat m) + -[n +1]) = ipow a (sub_nat_nat m (nat.succ n)) : rfl
-        ... = ipow a (-[nat.pred (nat.sub (nat.succ n) m) +1])            : {sub_nat_nat_of_lt H}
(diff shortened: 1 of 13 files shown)


fpvandoorn:
@avigad: It is fine that the libraries diverge. I was already planning to remove the ported HoTT files, and port everything again after the new simplifier exists and we don't rely on the fake simplifier anymore. I tried to make as little additions to the HoTT files as possible, and instead created a new file with HoTT-specific theorems (e.g. [`hott/types/int/hott.hlean`](https://github.com/avigad/lean/blob/master/hott/types/int/hott.hlean)).

@leodemoura: do we need `max` and `min` for `nat` in `init`? Currently everything up to `data.nat.order` compiles if I remove those definitions. Will later features use them? If not, I think it would be nicer to move `max`/`min` to `data.nat.order` and use the algebra library as much as possible to prove things about them.

leodemoura:
> @leodemoura: do we need max and min for nat in init? Currently everything up to data.nat.order compiles if I remove those definitions. Will later features use them? If not, I think it would be nicer to move max/min to data.nat.order and use the algebra library as much as possible to prove things about them.

The original plan was to use `max` when automatically generating the `direct_subterm` relation (#347). However, this approach would not cover reflexive datatypes such as 
https://github.com/leanprover/lean/blob/master/tests/lean/run/ftree_brec.lean#L3-L5

So, I don't have any reason for keeping `max` and `min` at `init`.

avigad:
I will remove max and min from init when I am back in Pittsburgh. Floris, I will also remove it from the HoTT library, to facilitate the port later on, unless you tell me not to.

fpvandoorn:
That's good!

PR #721, opened 2015-07-04
Fin choice

fgdorais:
I've found `fin.choice` useful for a project I'm working on and I thought I'd share it.

This is my first pull request to this project, so I'm not sure my proofs are up to standards. Please suggest corrections where necessary; I need the practice writing good proofs!

I just realized that I changed `fin.elim0` in a way that may be incompatible with the original version, though it doesn't seem to break any proofs from the current library.

### DIFF
--- a/library/data/fin.lean
+++ b/library/data/fin.lean
@@ -258,4 +258,40 @@ lemma val_succ : ∀ (i : fin n), val (succ i) = nat.succ (val i)
 definition elim0 {C : Type} : fin 0 → C
 | (mk v h) := absurd h !not_lt_zero
 
+definition zero_rec {C : fin 0 → Type} : Π i : fin 0, C i
+| (mk v h) := absurd h !not_lt_zero
+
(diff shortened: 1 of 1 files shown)


leodemoura:
Thanks for making the pull request. Contributions are very welcome. I have integrated your changes.
I took the liberty to change the commit messages, and use our commit conventions
https://github.com/leanprover/lean/blob/master/doc/commit_convention.md

BTW, in Lean, we have two libraries: standard and hott. In the standard library, Prop is proof irrelevant and impredicative. In "HoTT mode", there is no Prop. By default, lean assumes that `.lean` files use standard, and `.hlean` files use hott.
I wanted to mention it because it seems you are interested in HoTT.

fgdorais:
Thank you for incorporating my contribution! I will make sure to use your commit conventions in future contributions.

Yes, I am interested in HoTT. However, the project I am currently working on is based on the proof-irrelevant version. That said, my next project is likely to be in HoTT mode.

PR #725, opened 2015-07-08
prime factorization

avigad:
The main thing here is a definition of `mult p n`, the multiplicity of `p` in `n`, i.e. the largest power of `p` that divides `n` (assuming `n > 0` and `p > 1`). Here is one of the main theorems:

``` lean
theorem dvd_of_forall_prime_mult_le {m n : ℕ} (mpos : m > 0) 
    (H : ∀ {p}, prime p → mult p m ≤ mult p n) : m ∣ n
```

I still need to prove

``` lean
n > 0 → n = ∏ p : prime_factors n, p^(mult p n)
```

but I don't think it will be hard with the facts already there. We may need more properties of sums and products. I also started developing some properties of powers in algebra, but there is still a lot to do.

@htzh I renamed a few theorems in primes.lean in this commit. I am sorry if this breaks some of your proofs, but it should be easy to fix.

@leodemoura There might be a bug here:

https://github.com/avigad/lean/blob/master/library/theories/number_theory/prime_factorization.lean#L106

It is strange, the proof breaks when I put brackets around that line, but `begin ... end` works o.k. It might be hard to produce a small counterexample; I did this sort of thing many times, without problems. But let me know if you want me to try.

There are a lot of proofs by induction and cases in prime_factorization, like this one:

https://github.com/avigad/lean/blob/master/library/theories/number_theory/prime_factorization.lean#L68-L89

If we eventually have good syntax for induction and cases in proof terms, this would be a good place to try it out.

### DIFF
--- a/library/algebra/algebra.md
+++ b/library/algebra/algebra.md
@@ -13,7 +13,9 @@ Algebraic structures.
 * [ring](ring.lean)
 * [ordered_group](ordered_group.lean)
 * [ordered_ring](ordered_ring.lean)
+* [ring_power](ring_power.lean) : power in ring structures
 * [field](field.lean)
 * [ordered_field](ordered_field.lean)
+
 * [category](category/category.md) : category theory
 
--- a/library/algebra/ordered_ring.lean
+++ b/library/algebra/ordered_ring.lean
@@ -19,7 +19,7 @@ definition absurd_a_lt_a {B : Type} {a : A} [s : strict_order A] (H : a < a) : B
 absurd H (lt.irrefl a)
 
 structure ordered_semiring [class] (A : Type)
-  extends semiring A, ordered_cancel_comm_monoid A, zero_ne_one_class A :=
+  extends semiring A, ordered_cancel_comm_monoid A :=
 (mul_le_mul_of_nonneg_left: ∀a b c, le a b → le zero c → le (mul c a) (mul c b))
 (mul_le_mul_of_nonneg_right: ∀a b c, le a b → le zero c → le (mul a c) (mul b c))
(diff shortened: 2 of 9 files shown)


leodemoura:
> It is strange, the proof breaks when I put brackets around that line, but begin ... end works o.k. It might be hard to produce a small counterexample; I did this sort of thing many times, without problems. But let me know if you want me to try.

I think you have defined tokens such as `]}`. 
So, the string

``` lean
{rewrite [pow_zero, one_mul, zero_add]}
```

is tokenized as

``` lean
'{'   'rewrite'   '['   'pow_zero'   ','    'one_mul'   ','  'zero_add'  ']}'   
```

avigad:
Doh! Burned by my own notation yet again. It is from the notation `{[a]}` for a singleton in finset (and also in set).

I will change it. If it caused problems for me, it will cause problems for other users. Maybe I will try `'{a}`, unless you foresee problems with that too. I defined notation `f '[s]` for the image of `s` under `f`, and that hasn't caused problems, at least not yet.

leodemoura:
`'{a}` is fine. Usually, a token is problematic when it is the concatenation of two existing tokens.

fpvandoorn:
This seems to be a recurring problem. Would it help if the notation for finset was declared as 

``` lean
notation `{` `[` a `]` `}` := ...
```

Or does that have other problems?

leodemoura:
> Or does that have other problems?

Yes, it conflicts with notation decls such as 

``` lean
notation `{` binders `|` r:(scoped:1 P, subtype P) `}` := r
```

We cannot have both of them simultaneously. The problem is that after consuming `{`, the first declaration is telling the parser to consume `[`, and the one above it is saying to consume a binder. Moreover, a binder may start with `[`.

PR #726, opened 2015-07-11
choose, and varia

avigad:
The file `choose.lean` gives a recursive description of `choose n k` and proves it is equivalent to the usual definition with factorials. (There is a lot more to do in that file.) The name currently conflicts with `nat.choose` in `/library/data/nat/choose.lean`, so we cannot load both at once.

Questions:
- Should we change the other `choose` to `find`?
- Should we introduce infix notation for `choose`?

@htzh: this pull request includes the two theorems you asked for, with two small changes:
- I reoriented the equation in `image_compose` (nso it agrees with the corresponding theorem for sets)
- The theorem `Union_const` was false as you stated it (this is why we have formal proof!). I added the hypothesis `s ≠ ∅`.
  By the way, I also changed the theorem `non_empty_of_card_succ` to `ne_empty_of_card_eq_succ`. I think you use that theorem?

I changed the notation `{[a, b, c]}` to `'{a, b, c}`, to avoid notational conflicts. Anyhow I like the new notation better.

By the way, `choose.lean` contains some more examples of proofs with induction and cases. It would be interesting to know if they can be made more readable with better support for induction and cases in proof term mode.

### DIFF
--- a/library/data/finset/basic.lean
+++ b/library/data/finset/basic.lean
@@ -213,7 +213,7 @@ decidable.by_cases
   (assume H : a ∈ s, by rewrite [card_insert_of_mem H]; apply le_succ)
   (assume H : a ∉ s, by rewrite [card_insert_of_not_mem H])
 
-lemma non_empty_of_card_succ {s : finset A} {n : nat} : card s = succ n → s ≠ ∅ :=
+lemma ne_empty_of_card_eq_succ {s : finset A} {n : nat} : card s = succ n → s ≠ ∅ :=
 by intros; substvars; contradiction
 
 protected theorem induction [recursor 6] {P : finset A → Prop}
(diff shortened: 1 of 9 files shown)


leodemoura:
> Should we change the other choose to find?

No problem.`find` is also a good name for the other `choose`.

> Should we introduce infix notation for choose?

This will create problems. If we do that, then `choose` will be a "keyword", and we will not be able to use it as an identifier anymore.

htzh:
@avigad Thanks for getting to this so fast. The non empty requirement is not a problem. However I think you have an extraneous argument `a : A` in the lemma signature: https://github.com/avigad/lean/commit/913d72b82f6829c88b7106a18d591ce906111cb1#diff-4ae31cb8c9d30d393fae380ac6d6e5eaR122

avigad:
@htzh, thanks for catching that. It is fixed now.

@leodemoura, I will rename the other choose later, but I will not use infix.

leodemoura:
> @leodemoura, I will rename the other choose later, but I will not use infix.

BTW, there is a `choose` in encodable.lean too.

PR #731, opened 2015-07-16
More lemmas in advance of group theory migration

htzh:
After this one I will be moving the main files into theories/group_theory.

### DIFF
--- a/library/algebra/group.lean
+++ b/library/algebra/group.lean
@@ -281,6 +281,54 @@ section group
   theorem mul_eq_one_iff_mul_eq_one (a b : A) : a * b = 1 ↔ b * a = 1 :=
   iff.intro !mul_eq_one_of_mul_eq_one !mul_eq_one_of_mul_eq_one
 
+  definition conj_by (g a : A) := g * a * g⁻¹
+  definition is_conjugate (a b : A) := ∃ x, conj_by x b = a
+
+  local infixl `~` := is_conjugate
+  local infixr `∘c`:55 := conj_by
+
+  lemma conj_compose (f g a : A) : f ∘c g ∘c a = f*g ∘c a :=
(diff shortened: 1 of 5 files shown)


avigad:
I would like to clean up the treatment of "disjoint". We have three ways of saying that two sets are disjoint. There is one here:

  https://github.com/leanprover/lean/blob/master/library/data/finset/basic.lean#L486-L521

Now there is the new `disjoint_sets`, and there is simply `s₁ ∩ s₂ = ∅`. It is confusing to have so many. 

Both `disjoint_sets` and `s₁ ∩ s₂ = ∅` are useful. (The second is clearly decidable, so we can do proofs by cases, and it is also good for calculation.)

I propose we:
- get rid of the first
- rename `disjoint_sets` to `disjoint`.
- for uniformity, generally prefer `s₁ ∩ s₂ = ∅` in the statement of theorems (it is short, decidable, and clear)
- use lemmas `disjoint_of_inter_eq_empty` and `inter_eq_empty_of_disjoint` to mediate between the two formulations.

We should use the same convention for `set`.

@leodemoura @htzh What do you think?

htzh:
@avigad disjoint_sets describes a set of sets that are pairwise disjoint. I don't think it is an interchangeable concept with disjoint. It does use the empty intersection definition of disjoint, as the comment in finset/basic makes clear that the quot definition of disjoint is for internal use.

htzh:
The definition of `disjoint_sets` and the lemma `disjoint_sets_filter_of_disjoint_sets` should be moved out of partition namespace and into the finset namespace as they are more general concepts than partition.

avigad:
My apologies -- I scanned the files too quickly. Yes, I agree, it should be moved to the finset namespace. I'll be happy to do it when I get back, but, as always, feel free to do it if you prefer.

PR #738, opened 2015-07-24
feat(library): clean up "sorry"s in library

digama0:
Breaking changes: pnat was redefined to use subtype instead of a custom inductive type, which affects the notation for pnat 2 and 3


leodemoura:
Wow! I'm impressed.

Thanks,
Leo

avigad:
I am too -- thanks also for cleaning up so many definitions and proofs. They look great.

BTW, the congruence classes in logic/instances are "legacy", from a previous attempt at handling rewriting with congruences. @leodemoura, should this be deleted now?

PR #754, opened 2015-07-29
Minor changes related to unfolding things

fpvandoorn:
I added the `unfold` attribute to `add`, `sub`, `mul` and `of_num` in `nat`. This might break some proofs using natural numbers + unfolding tactics (like `esimp`). I fixed all occurrences in the library.


fpvandoorn:
I added two more commits with some new theorems.

leodemoura:
Dear Floris,

I'm getting compilation errors when I try to compile your commits on top of the latest master branch.
Could you please take a look?

Thanks,
Leo

fpvandoorn:
Oops, I thought I double checked it. I will fix it tomorrow.

fpvandoorn:
There were indeed 2 more errors after rebasing my files on `blessed/master`. I'm not sure why they appeared. I fixed them now.

leodemoura:
I'm still getting errors at `susp.hlean` and `pnat.lean`.

``` lean
/home/leo/projects/lean/hott/hit/susp.hlean:159:33: error:invalid 'rewrite' tactic, rewrite step failed
proof state:
X : Type,
x : X,
Y : Type,
f : X → Y,
p : x = x
⊢ ap f p = pmap.rec
    (λ
     (map :
       nat.rec (λ (A : Pointed), A) (λ (n : nat) (IH : Pointed → Pointed) (A : Pointed), IH ΩA) (nat.of_num 0)
         Ω(pointed.Mk x) →
       nat.rec (λ (A : Pointed), A) (λ (n : nat) (IH : Pointed → Pointed) (A : Pointed), IH ΩA) (nat.of_num 0)
         Ω(pointed.Mk (f x)))
     (resp_pt :
       map
         (Pointed.Point
            (nat.rec (λ (A : Pointed), A) (λ (n : nat) (IH : Pointed → Pointed) (A : Pointed), IH ΩA)
               (nat.of_num 0)
               Ω(pointed.Mk x))) = Pointed.Point
         (nat.rec (λ (A : Pointed), A) (λ (n : nat) (IH : Pointed → Pointed) (A : Pointed), IH ΩA) (nat.of_num 0)
            Ω(pointed.Mk (f x)))),
       map)
    (nat.rec (λ (A : Pointed) (B : Pointed) (f : A →* B), f)
       (λ (n : nat) (IH : Π (A : Pointed) (B : Pointed), A →* B → Ω[n ] A →* Ω[n ] B) (A : Pointed)
        (B : Pointed) (f : A →* B),
          IH ΩA ΩB
            (pmap.mk
               (λ (q : Pointed.Point A = Pointed.Point A), (respect_pt f)⁻¹ ⬝ ap (pmap.map f) q ⬝ respect_pt f)
               (con.left_inv (respect_pt f))))
       (nat.of_num 0)
       Ω(pointed.Mk x)
       Ω(pointed.Mk (f x))
       (pmap.mk (λ (q : x = x), refl (f x) ⬝ ap f q) idp))
    p
```

and

``` lean
/home/leo/projects/lean/library/data/pnat.lean:193:2: error: 1 unsolved subgoal
p : ℕ+,
H : rat_of_pnat (tag 2 dec_trivial * p) = rat_of_pnat p * (1 + 1)
⊢ 1 / rat_of_pnat (tag 2 dec_trivial * p) + 1 / rat_of_pnat (tag 2 dec_trivial * p) = 1 / (rat_of_pnat
     p * (1 + 1)) + 1 / (rat_of_pnat p * (1 + 1))
```

fpvandoorn:
Sorry that I messed up again. Apparently compiled files where cached, even though a file they imported was changed. 

I see your point. But I think the datatype `nat` is just not suited for working with large numbers. Even this already gives a large output:

``` lean
open nat
set_option pp.numerals false
eval (100 : nat)
```

And adding unfold attributes does has its merits. E.g. if I write `2 + 1` then it's nice if `esimp` reduces this to just `succ (succ (succ zero))`, because then it probably matches with more expressions. Or if I define

``` lean
definition f [unfold 1] (n : ℕ) : ℕ :=
by induction n with n fn; exact zero; exact succ (succ fn)
```

and I write `f (n + 2)` then calling `esimp` does nothing, because the head of `n + 2` is not a constructor. Moreover, rewrite cannot match the pattern `f (succ ?M)` with it (though `xrewrite` can).

Examples:

``` lean
open nat
definition f [unfold 1] (n : ℕ) : ℕ :=
by induction n with n fn; exact zero; exact succ (succ fn)
theorem foo (n : ℕ) : f (succ n) = succ (succ (f n)) := rfl

-- disable unfold attributes
attribute add nat.of_num [unfold 3]

-- need to explicitly unfold f
example : f 3 = 0 :=
begin esimp, state,
      unfold f, state, exact sorry end

example : f 3 = 0 :=
begin try rewrite foo, state,
      xrewrite foo, state, exact sorry end

set_option pp.numerals false
example : 2 = succ 3 :=
begin esimp, state, exact sorry end
```

I will push my files again. If you want to can leave out the last commit where I add the unfold attributes if you don't like that change.

fpvandoorn:
It seems that destructively updating my commits remove the message you posted on one of my commits from this pull request. I was responding to [this message](https://github.com/fpvandoorn/lean/commit/f62af59f36136c54e57763eced2435b6a65ab907#commitcomment-12455450):

> @fpvandoorn I think this is a dangerous annotation.
> If the user has a term such as `a + 100000` and runs `esimp`, then Lean will run out of memory. I think the user will not understand what is going on.
> Similarly, a term such `10000 * 10000` will generate problems.

leodemoura:
> I write f (n + 2) then calling esimp does nothing, because the head of n + 2 is not a constructor.

Hopefully, we will be able to address this kind of issue with better automation.
`esimp` is a "dumb" black box. It does not have support for deciding whether it is a good idea to reduce `n+k` or not.  

What about the following alternative?
1- We put `attribute add sub ... [unfold 2]` in a separate namespace (e.g., `nat.esimp`)
2- Then, if users want this feature they just open this namespace.

fpvandoorn:
Ok, that is reasonable. I put those attributes in the namespace `nat_esimp` (`nat.esimp` is not accepted since `esimp` is already a tactic).

leodemoura:
Thanks.

PR #757, opened 2015-07-30
Correct various mistakes in the basic tutorial

Zimmi48:
These are mistakes I was able to catch while reading the tutorial. I tried to respect the guide for commit messages. Sorry in advance if they are not exactly as you asked.


leodemoura:
Thanks! I have integrated your changes.

PR #759, opened 2015-07-31
Integer floor/ceiling and supremum property for reals

robertylewis:
This proves that every bounded, inhabited set of reals has a supremum. The proof does not use Cauchy completeness, so in effect this is a second completeness proof. (The proof from Cauchy completeness uses most of the same machinery as the direct proof.)

To prove this we use various embedding integers in the reals: we need integer floor and ceiling, and an Archimedean property.

### DIFF
--- a/library/algebra/field.lean
+++ b/library/algebra/field.lean
@@ -297,8 +297,6 @@ section field
   theorem div_div_div_div (Hb : b ≠ 0) (Hc : c ≠ 0) (Hd : d ≠ 0) : (a / b) / (c / d) = (a * d) / (b * c) :=
     by rewrite [(div_div_eq_mul_div Hc Hd), (div_mul_eq_mul_div), (div_div_eq_div_mul Hb Hc)]
 
-  -- remaining to transfer from Isabelle fields: ordered fields
-
 end field
 
 structure discrete_field [class] (A : Type) extends field A :=
--- a/library/algebra/field.lean
+++ b/library/algebra/field.lean
@@ -150,11 +150,11 @@ section division_ring
             ... = -(b * (1 / a))  : neg_mul_eq_mul_neg
             ... = - (b * a⁻¹)     : inv_eq_one_div
 
-  theorem neg_div (Ha : a ≠ 0) : (-b) / a = - (b / a) :=
+  theorem neg_div : (-b) / a = - (b / a) :=
     by rewrite [neg_eq_neg_one_mul, mul_div_assoc, -neg_eq_neg_one_mul]
 
   theorem neg_div_neg_eq_div (Hb : b ≠ 0) : (-a) / (-b) = a / b :=
-    by rewrite [(div_neg_eq_neg_div Hb), (neg_div Hb), neg_neg]
+    by rewrite [(div_neg_eq_neg_div Hb), neg_div, neg_neg]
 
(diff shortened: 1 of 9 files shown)


leodemoura:
@rlewis1988 in the list above, we have the commit `temporary commit: ...`.
I squashed it before integrating your changes.
I also checked if it was still crashing Lean. It does not. I think the bug fixes performed this week fixed the problem.

Best,
Leo

robertylewis:
Oops, sorry! I thought I had removed that commit from master when I made a new branch.

robertylewis:
A question about this commit: in complete.lean line 753, I define

``` lean
theorem rat.pow_add (a : ℚ) (m : ℕ) : ∀ n, rat.pow a (m + n) = rat.pow a m * rat.pow a n :=
  rat.pow_add a m
```

Without this "redefinition," the rewrite using rat.pow_add in the following proof breaks. It remains broken if you try substituting rat.pow_add instead of rewriting. There are no implicit arguments that are being filled in by the redefinition.

I guess, since rat.pow_add is defined by migrate, the definition is "complicated" and Lean doesn't find the useful version without a hint. This is problematic though. Leo, do you know what's going on here? Is there a good way to avoid this?

avigad:
If you print `rat.pow_add` just before redefining it, you will see that it is a mess. I think this is the result of a lack of robustness in the migrate command -- things that are defined in algebra have to be redefine manually before the migrate, or else they get expanded too much. I will try to fix this now.

avigad:
Yup, that was it. It was easy to fix (in pull request #761). Because `field` imports `group_power`, power is was defined for `rat` in `rat.basic`, and we have to define the local version before doing the migrate.

This is an annoying, non-user friendly aspect of `migrate`. I believe Leo is planning to fix this, by making migrate declare local versions of all definitions in the source. But it is tricky. `migrate` should possibly also define notations and transfer attributes, which it doesn't do now. But then if it does, there will probably be situations where the user wants to override that behavior.

leodemoura:
@avigad the `migrate` command uses a simple heuristic to decide what is going to be unfolded: it does not unfold definitions from the namespace we are migrating to. The problem is that `pow_add` is mixing `rat` and `nat`. So, the definitions from `nat` are being expanded when we migrate from `algebra` to `rat`. This is why we see `nat.rec`s when we print `rat.pow_add`.

I'm not happy with this command, we need a better solution. 
Another nasty thing with this command is that we have to enable `migrate.trace` to understand what is going on.

avigad:
Actually, the problem is that we needed to define `rat.pow` explicitly in `rat.basic` before the migrate. I did that now, so we're good. And we're on the same page about `migrate`. 

Getting all the things we want -- handling definitions, handling metadata,  handling parameters, ... will be a real headache for you. But let's not let that obscure the fact that even in its current state, `migrate` is really great.

leodemoura:
> But let's not let that obscure the fact that even in its current state, migrate is really great.

Whenever I use it I get upset :-(
I basically have to turn on `migrate.trace` and then manually inspect the trace messages and check if it did what I wanted it to do.

avigad:
I usually do `print prefix` afterwards, open up the messages window in a new frame, and scroll through to make sure everything looks o.k.

Remember, I used to do migrates /by hand/. Maybe the command isn't robust enough yet for everyday users (and we still haven't described it in the tutorial), but I am very happy we have it.

PR #762, opened 2015-08-01
feat(library/data/fin): add foldl and foldr

fgdorais:
Two small convenient additions to finite ordinal types.


leodemoura:
Thanks!
It has been merged.

Issue #760, opened 2015-07-31
iff or two implies?


robertylewis:
Throughout the algebra library, we have theorems like the following:

``` lean
add_lt_add_iff_lt_sub_right : a + b < c ↔ a < c - b
```

Sometimes these theorems have one-directional versions alongside, other times they don't. (I tried to include the one-directional versions in field and ordered_field, although I'm not sure how consistent I was with it.) There are a few reasons they are nice to have around. 

One, the names can be confusing. When I'm looking for a theorem using our standard naming convention, the iffs are half the time named in the opposite order, which makes tab-complete less effective.

Two, when these theorems appear within quantifiers, the syntax to use them gets slightly annoying. I've found myself having to do the following quite a lot:

``` lean
iff.mpr !int.lt_add_iff_sub_lt_right Hz
```

It's very easy to forget to include the !, especially in tactic mode (`apply iff.mpr !int.lt_add_iff_sub_lt_right` doesn't feel natural).

These are small complaints, but I think the algebra library would be more pleasant to use with the one-directional versions added. Thoughts?

avigad:
I agree that we should also have the one-directional versions. The `iff` are good for rewrite proofs (and `calc` proofs, if we use propext), but it is annoying to use `iff.mp` and `iff.mpr`, especially for common calculation rules.

Most things in the library have at least the one-way versions, and possibly all three. But building the library was tedious, and I got lazy at times. Now, I try to make it a habit of adding one-way versions whenever I want them and they are not there. If others get in the habit of doing the same, that would be really helpful.

robertylewis:
When I have the chance, I'll go through and add as many missing ones as I have the patience to do. A bit of frustration now to save a lot in the future, maybe?

avigad:
Yes, that's library development, in a nutshell. It is hard to have the discipline to do everything right the first time, but it is more work to repair things later.

PR #763, opened 2015-08-03
Cleanup and small additions to reals

robertylewis:
This is just a cleanup/minor refactoring of the reals. I've moved a bunch of the lemmas to the appropriate places in algebra.


robertylewis:
I've added one more commit. Jeremy noticed that the following failed:

``` lean
import data.real
open real

check 0 + 0
check 3 + 3
```

This is because the rat namespace was opened without coercions in real/basic where of_rat is defined. The last commit fixes this.

robertylewis:
And, one more commit, unrelated to the others. This begins to address #760.

PR #764, opened 2015-08-03
feat(library/algebra/group_power): add times (additive variant of power)

fgdorais:
Following common usage, I made `times` act on the left rather than acting on the right, like `pow`. I adjusted the order of arguments to all theorems to match.

### DIFF
--- a/library/algebra/group.lean
+++ b/library/algebra/group.lean
@@ -348,6 +348,11 @@ structure comm_group [class] (A : Type) extends group A, comm_monoid A
 structure add_group [class] (A : Type) extends add_monoid A, has_neg A :=
 (add_left_inv : ∀a, add (neg a) a = zero)
 
+definition add_group.to_group {A : Type} [s : add_group A] : group A :=
+⦃ group, add_monoid.to_monoid,
+  mul_left_inv := add_group.add_left_inv ⦄
+
+
 section add_group
 
   variables [s : add_group A]
@@ -561,7 +566,7 @@ section add_comm_group
 
   theorem sub_sub_self (a b : A) : a - (a - b) = b :=
   by rewrite [sub_eq_add_neg, neg_sub, add.comm, sub_add_cancel]
- 
+
 end add_comm_group
 
 definition group_of_add_group (A : Type) [G : add_group A] : group A :=
--- a/library/algebra/group_power.lean
+++ b/library/algebra/group_power.lean
@@ -167,4 +167,71 @@ theorem pow_two_add (n : ℕ) : pow 2 n + pow 2 n = pow 2 (succ n) :=
 
 end ordered_ring
 
+/- additive monoid -/
+
+section add_monoid
+variable [s : add_monoid A]
+include s
+local attribute add_monoid.to_monoid [trans-instance]
+open nat
+
+definition times : ℕ → A → A := λ n a, pow a n
+
+infix [priority algebra.prio] `*` := times
+
+theorem zero_times (a : A) : (0:ℕ) * a = 0 := pow_zero a
+theorem succ_times (n : ℕ) (a : A) : succ n * a = times n a + a := pow_succ a n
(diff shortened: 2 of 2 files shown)


avigad:
@fgdorais This is a nice addition. Thanks!

As I was driving in two work this morning, I thought about it, and two things occurred to me. The first is minor, the second is more important.

First, I think I would prefer to use `smul` for scalar multiplication instead of `times`. It seems awkward to use both `times` and `mul` to describe different kinds of multiplication, and also one would expect `times` to have two arguments of the same type. The phrase "scalar multiplication" is familiar and more descriptive.

Second, I am worried about further overloading `*` whenever this file is imported and algebra is open. At times, it will even be ambiguous: another way to interpret `3 * x` in a ring is to coerce `3` to an element of the ring and carry out the usual multiplication there. One solution is to put the notation in a namespace so that the user has to open it explicitly. I think a better solution is to use another symbol, like `⬝`. (This is currently entered as \cdot, but we can give it other names as well.)

In short, I am proposing to
- rename `times` to `smul` (and revise the theorem names accordingly)
- replace `*` by `⬝`
  Please let me know if you see any problems with this.

By the way, I am also planning to rename the file `monoid_power`, and move the group power operation to theory/finite_groups. Using negative exponents in that way is more specialized to group theory. We should also define `field_power`, so that we can use negative exponents with the rationals and reals. I will keep that in the algebra folder.

fgdorais:
@avigad Happy to help!

I'm happy with `smul` and your rationale makes a lot of sense, though I'm less happy with `ismul` which suggests a proposition `is_mul`. Another option is to make the left type explicit, for example: `ntimes`/`itimes`, `nat_times`/`int_times`, `nmul`/`imul`.

I was worried about overloading `*` too and I was surprised that there were no issues. I'm happy with using `⬝` instead and I think that will also improve overall readability.

I'm less sure about your renaming plan. It seems to me that algebra is a fine place for this since it has broad applicability. It seems odd to have to import theory/finite_groups to get `ipow`.

avigad:
Ah, I forgot about the "i" variant. I'd like to keep the names "pow" and "smul" for the nat versions, because they are the most commonly used.

We will have a similar problem when we define integer power for fields, which will be a different notion, since we will want to define a value for 0. How about `gpow` and `gsmul` for integer power and scalar multiplication in a group, and `fpow` for integer power in a field? (Fortunately or unfortunately, gsmul will work on the additive part of a field, so users will have to get used to that. Anyhow, in classes like the reals etc. I think it will be more common to write n \* a and have n coerced to a real.)

You have convinced me that it is worthwhile keeping integer powers in algebra, so I'll leave them there, whatever we end up calling them.

fgdorais:
@avigad I think `gpow` and `fpow` make sense as a way of distinguishing the group and field cases. What do Coq and Isabelle do? Perhaps it's worth thinking about what people who will write algebra/ring_module or theories/modules would prefer? If that were me, I would want to use `smul` for scalar multiplication on modules and I might find it annoying that `smul` is already defined.

Let me make a quick visual argument in favor of `ntimes`:
t

At this point, I've thought about this a little too much to think any option is much better than the others. Your global vision of the lean library is much better than mine and I'm happy to go with what you think is best.

avigad:
I see, people will want to use smul for scalar multiplication with the associated ring, not just scalar multiplication by natural numbers. Good point. Life is so complicated!

I am now inclined to go with the following scheme:

  nmul : scalar multiplication by a natural number
  imul  : scalar multiplication by an integer
  smul : general scalar multiplication

as well as

  pow   : natural number power in any monoid
  gpow : integer power in a group
  fpow  : integer power in a field

It's not perfect but it does group the analogous concepts together and the mnemonics make sense. I think it will work.

fgdorais:
@avigad I don't see any immediate issues with this scheme. I'm on board!

PR #765, opened 2015-08-04
refactoring max and min, etc.

avigad:
This pull request gets rid of nat.diff (which duplicated nat.dist), separates lattices from order (we will use min and max for linear orders, inf and sup for lattices), and makes nat an instance of discrete_linear_order, so it shares properties of min and max with the other number classes. It also renames some theorems involving max and min.

### DIFF
--- a/library/algebra/algebra.md
+++ b/library/algebra/algebra.md
@@ -3,10 +3,12 @@ algebra
 
 Algebraic structures.
 
+* [prio](prio.lean) : priority for algebraic operations
 * [function](function.lean)
 * [relation](relation.lean)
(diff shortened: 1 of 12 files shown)


avigad:
I almost forgot. When I ran ctest, two tests failed:

```
The following tests FAILED:
         23 - show_goal_bag (Failed)
        848 - leanittest_num2.input (Failed)
Errors while running CTest
```

The second one always fails for me, so I ignore it. But I have never seen the first one before. I found a script `show_goal_bag.sh` in tests/extras, but I could not figure out what to do with it. If there is something I should do, please let me know.

PR #781, opened 2015-08-06
mapping from HoTT book to HoTT library

fpvandoorn:
I added a file which indicates what parts of the HoTT book have been formalized: https://github.com/fpvandoorn/lean/blob/596a9987551c8481b9ca405d881349147268ef05/hott/book.md

We have most material of the first 7 chapters (part I), but have very little of the last 4 (part II).

I also added some files to make sure our library has at least everything of chapter 2. And I did some refactoring.


avigad:
This list is really nice. Should we add a link to the Lean page? Under "documentation", we can expand the line

> You can also browse the homotopy type theory library directly or through markdown files.

to

> You can also browse the homotopy type theory library directly or through markdown files. Substantial parts of the [HoTT book] have been formalized; see the list [here].

Thoughts?

PR #785, opened 2015-08-10
filters on sets form a complete lattice

avigad:
@leodemoura, I instantiated complete_lattices based on a previous commit. So my last commit fixes a conflict that arises two commits earlier, and my second-to-last conflict fixes three tests that were broken two commits earlier. I don't know how to fix the intermediate commits, short of squashing the last four down to one. Please let me know if I should have done that, or if there was something better to do.

It is nice that you got complete_lattice down to a minimal set of primitives and axioms. But I think it might be more natural and convenient to go the other way, and have inf, sup, Inf, and Sup as part of the structure, with the relationships between them proved as identities.

Here is why. In the example I just did, it is easy to define Sup on filters, but hard to define Inf. In fact, I defined Inf in terms of Sup, using the dual equivalence to the one you used. On other structures, there may be natural definitions of the various operations. For example, on sets, inf, sup, Inf, and Sup are binary intersection, binary union, set intersection, and set union. On the reals, they are min, max, lub, and glb. With more elements in the structure, users can choose the definitions they want to work with, and then the alternative definitions are general theorems.

If you think this makes sense, I will be happy to make these changes when you are done with complete lattices.


leodemoura:
I refactored `complete_lattice.lean`. Now, `complete_lattice` is the expanded version with all fields.
Then, I added two minimal versions: one based on `Inf`, and another on `Sup`. Then, I added definitions to map between them.

avigad:
Perfect. Thanks!

PR #795, opened 2015-08-14
feat(library/coercion): improve error messages for coercions

### LINKED ISSUE #794: Error message for level-param mismatch in coercion
The following:

``` Lean
constants (A B : Type)
constants (f : A → B) 
attribute f [coercion]
```

yields:

```
 error: invalid coercion, type of 'f' does not match any of the allowed expected types for coercions
  Pi (x_1 : A_1) ... (x_n : A_n) (y: C x_1 ... x_n), D t_1 ... t_m
  Pi (x_1 : A_1) ... (x_n : A_n) (y: C x_1 ... x_n), Type
  Pi (x_1 : A_1) ... (x_n : A_n) (y: C x_1 ... x_n), A -> B
```

The problem is `f`'s extra level corresponding to `B`. The coercion is pruned at

https://github.com/leanprover/lean/blob/master/src/library/coercion.cpp#L327

and it works if we fix the universe of `B`.

The issue is that the error message does not display enough information to diagnose the problem.

dselsam:
Include level information in primary coercion error message if pp_options are set to display levels.

### DIFF
--- a/src/library/coercion.cpp
+++ b/src/library/coercion.cpp
@@ -341,8 +341,9 @@ static environment add_coercion(environment const & env, name const & f, name co
             check_levels(const_levels(C_fn), d.get_univ_params())) {
             optional<name> cls = type_to_coercion_class(binding_body(t));
             if (!cls)
-                throw exception(sstream() << "invalid coercion, '" << f << "' cannot be used as a coercion from '"
-                                << C << "'");
+                throw exception(sstream() << "invalid coercion, '" << f << "' cannot be used as a coercion from source '"
+                                << C << "', since target does not match any of the acceptable forms\n"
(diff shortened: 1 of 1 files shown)


dselsam:
There are still paths in the coercion-handler that yield unhelpful error messages, but I think this addresses the issue at hand. The approach is to avoid the complexity of tracking all the checks that failed, and instead simply display generic level information in the message if the pp_options are set to display levels. I assume there is a rule never to mention the existence of levels with the default options, or else I might suggest always displaying the levels in the message. I also assume that when a user is confused about a message she will know to try `set_option pp.all true`. I am also open to alternate solutions if this one is unsatisfying.

PR #798, opened 2015-08-14
sqrt 2 irrational

avigad:
I will prove a better theorem soon, namely, that any integer that is not a perfect square has an irrational square root. But I wanted to push this in the meanwhile, since I made small changes all over the place.

The good news is that one statement of the fact that the square root of two is irrational has a very short, nice proof:

https://github.com/avigad/lean/blob/master/library/theories/number_theory/square_root_irrational.lean#L13-L24

We can hope that soon every line can be proved "by auto". 

The less good news is that stating it requires a lot more work to translate this into a form that mentions the rationals:

https://github.com/avigad/lean/blob/master/library/theories/number_theory/square_root_irrational.lean#L32-L51

The problem is that we have to translate the statement about the rationals to a statement about the integers, and then translate a statement about the integers to a statement about the natural numbers. It is very fiddly and delicate.

I mention this only to bookmark it as an example of the difficulties in mixing different number domains, and a good test case for future automation that tries to address the problem.

### DIFF
--- a/library/algebra/group_bigops.lean
+++ b/library/algebra/group_bigops.lean
@@ -84,7 +84,7 @@ section monoid
   | (a::l) := take b, assume Pconst,
     assert Pconstl : ∀ a', a' ∈ l → f a' = b,
       from take a' Pa'in, Pconst a' (mem_cons_of_mem a Pa'in),
-    by rewrite [Prodl_cons f, Pconst a !mem_cons, Prodl_eq_pow_of_const b Pconstl, length_cons, add_one, pow_succ' b]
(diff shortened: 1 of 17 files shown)


leodemoura:
In Isabelle, they have:

```
       prime p -> forall (q : rat), not (q^2 = p)
```

To be more precise they have

```
theorem sqrt_prime_irrational: "prime p ⟹ sqrt (real p) ∉ ℚ"
```

Do you think it is interesting to have it in Lean?

avigad:
Yes, I'll prove a better theorem, `q^2 = a -> denom q = 1`, i.e. `q^2 = a -> a = (num q)^2`. This last one is an equation in the integers; it is easy to prove that if p is prime, then we can't have `p = x^2`.

PR #811, opened 2015-08-27
various housecleaning

avigad:
There are three things worth mentioning.

First, I modified the definition of the reals so that the numerals 0 and 1 are used by migrate; otherwise, e.g., `add_zero` is `x + zero = x`, and fails to rewrite `t + 0`. Getting migrate to work was tricky and I still don't fully understand what the issues are, though things seem to be o.k. now. See the commit message to 0385ba6.

Second, I renamed lots of theorems in fields and ordered_fields to be more consistent with conventions followed elsewhere in the library, and set implicit /explicit arguments. (@rlewis tended to leave everything implicit.) There are a lot of theorems that are proved for division rings and fields, and then proved in a stronger form for discrete fields, where we assume x / 0 = 0. I made an executive decision as to how to handle that (but can be talked out of it). See the commit message to cc67d59.

Finally, I defined "-s" to be set complement in set and finset. In `init`, we have

``` lean
reserve prefix `¬`:40
reserve prefix `~`:40

reserve prefix `-`:100
```

I think we set the binding power of `~` low because we intended to use it as an ASCII equivalent to ¬. But we never defined that notation. So we now have some options:
- use `~` for not
- set the prefix binding power of `~` to 100, and use it instead of `-` for set complement.
- keep `-` for set complement, and leave `~` unused for now.
  If nobody has any opinions about this, I'll stick with the third.

### DIFF
--- a/library/algebra/field.lean
+++ b/library/algebra/field.lean
@@ -6,7 +6,6 @@ Authors: Robert Lewis
 Structures with multiplicative and additive components, including division rings and fields.
 The development is modeled after Isabelle's library.
 -/
-----------------------------------------------------------------------------------------------------
 import logic.eq logic.connectives data.unit data.sigma data.prod
 import algebra.binary algebra.group algebra.ring
 open eq eq.ops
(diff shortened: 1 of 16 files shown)


leodemoura:
> keep - for set complement, and leave ~ unused for now. If nobody has any opinions about this, I'll stick with the third.

I'm happy with the third option too.

Issue #812, opened 2015-08-30
lean compile issues


EgbertRijke:
I had some issues with compiling the hott library. My goal was to compile myself an up to date version of lean (with the help of Floris), so I cloned the code from github. However, when I run make, the hott library doesn't compile. A typical error message that appears in the output is:

`/home/egbert/Documents/Lean/hott/init/num.hlean:22:30: error: unknown identifier 'succ'`

Floris and I just found out that apparently, there is exists an old installation on my Ubuntu system, in `/usr/local/lib`, because in the file `succ.d` we find reference to these. When I go to this folder to inspect the lean files there, I find that these are created in January, maybe when I installed lean from the repository.

So I'd like to remove this installation. However, I cannot remove this installation with `sudo apt-get remove lean`, because it gives the error that it is unable to locate lean. We both don't really know how to proceed.

soonhokong:
@EgbertRijke, I'll see how to manually remove the previously installed lean files.

Apart from that, I think there is another issue. You (1) checkout the source code of Lean, (2) build a Lean binary, and 3) compile standard/hott libraries. In step (3), it should use the lean binary generated in step (2). Maybe that's not the case for now, and it uses the one in the system installed via apt-get. I'll double-check this one as well.

fpvandoorn:
I think the problem is with the generation of the dependency files. For example `num.d`, located in 

```
/home/egbert/Documents/Lean/hott/init/num.d
```

referenced

```
/usr/lib/lean/hott/init/logic.olean
```

(or something similar), instead of

```
/home/egbert/Documents/Lean/hott/init/logic.olean
```

(so `/home/egbert/Documents/Lean` is the location of the repository he just cloned and `/usr/lib/lean/` is the location of the old Lean which wasn't changed since January).

Also: the standard library compiled just fine.

soonhokong:
@EgbertRijke, @fpvandoorn, could you test the following?

``` bash
cd /home/egbert/Documents/Lean/hott
git clean -fxd
../bin/linja
```

EgbertRijke:
I get some warnings:

```
[95/98] "/home/egbert/Documents/Lean/b.../hott/algebra/category/adjoint.ilean" 
/home/egbert/Documents/Lean/hott/algebra/category/adjoint.hlean:138:12: warning: using 'sorry'
/home/egbert/Documents/Lean/hott/algebra/category/adjoint.hlean:164:46: warning: using 'sorry'
[...]

PR #813, opened 2015-08-31
changes in HoTT

fpvandoorn:
I've been working on many different things in the HoTT library again: some refactoring, some more theorems from chapter 3 of the book, some more category theory.

I have one concern. [This proof](https://github.com/fpvandoorn/lean/blob/ad9cceddde69fad7d5dd1f61007db4f51791308d/hott/algebra/category/adjoint.hlean#L78-L128) in the category theory library takes ~10 seconds to compile because of multiple times using `krewrite`. Because of that I've commented it out for now, but I will need to use it in the future. I don't really know what to do with it, I'm afraid it will be annoying to compile it every time if it is uncommented.

### DIFF
--- a/hott/algebra/category/adjoint.hlean
+++ b/hott/algebra/category/adjoint.hlean
@@ -4,12 +4,12 @@ Released under Apache 2.0 license as described in the file LICENSE.
 Authors: Floris van Doorn
 -/
 
-import algebra.category.constructions .constructions function arity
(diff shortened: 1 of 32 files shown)


fpvandoorn:
Oops, I still have some uncommented sorry's (which are referenced in https://github.com/leanprover/lean/issues/812#issuecomment-136509719). I'll fix it tomorrow.

PR #814, opened 2015-09-01
add basic syntax support for vim

c-cube:
just a basic syntax coloring for vim, highlighting the keywords, matching comments and delimiters, and  coloring modifiers.

Btw, any chance lean will have a vim plugin for filling holes and displaying types?

### DIFF
--- a/src/vim/README.md
+++ b/src/vim/README.md
@@ -1,0 +1,9 @@
+# Vim Syntax
+
+The following files are provided:
+
+- `syntax/lean.vim` (syntax highlighting)
+- `ftdtect/lean.vim` (detect that `.lean` are lean files)
+
+Just copy (or symlink) them to the corresponding `~/.vim/syntax/` and
+`~/.vim/ftdetect/` directories.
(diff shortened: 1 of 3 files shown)


soonhokong:
@c-cube, thank you!

Could you add a `src/vim/README.md` file explaining how to use the files?

> Btw, any chance lean will have a vim plugin for filling holes and displaying types?

There is a documentation, https://github.com/leanprover/lean/blob/master/doc/server.org , explaining the communication protocol between lean-server and an IDE. For now, emacs is the only IDE that implements the interface so far. I think it's possible to implement it on Vim while I have a little knowledge on Vim.

leodemoura:
@c-cube Thanks! 

> Btw, any chance lean will have a vim plugin for filling holes and displaying types?

We have been also implementing a simpler mechanism for extracting/reporting information from .lean files.
The following command lines can be used to extract information (we will add more options in the future).
- Retrieving the proof state at a given position in tactic mode

```
lean --line=<linenum> --col=<colnum> --goal file.lean
```

One can also use the cache file to speedup the execution.

```
lean --line=<linenum> --col=<colnum> --goal -c file.clean file.lean
```

In the lean Emacs mode, we generate the `.clean` when we compile the .lean file using flycheck.
- Retrieving the expected type for a "hole" `_`

```
lean --line=<linenum> --col=<colnum> --hole file.lean
```
- Retrieving information about an identifier and/or token.

```
lean --line=<linenum> --col=<colnum> --info file.lean
```

PR #815, opened 2015-09-01
changes in HoTT

fpvandoorn:
New version of pull request #813, without any `sorry`.


leodemoura:
@fpvandoorn I have analyzed `is_hprop_is_left_adjoint`. 
You are right, the performance bottleneck was `krewrite`. 
I rewrote the proof to workaround the performance problems. 
It now compiles under 1 sec on my machine.
I avoided some `krewrite` instances by using `esimp` before trying to rewrite the goal.
I kept `krewrite hom_of_eq_compose_right` and `krewrite inv_of_eq_compose_left`. However, I reduced the goal by abstracting irrelevant parts (I named them `lem₂` and `lem₃`). 
Note that these terms were quite big if we print them in full detail.

Unfortunately, the `abstract ... end` construct doesn't help here since we only create the abstraction after the current declaration has been fully elaborated.

fpvandoorn:
Great! I'm surprised how much `krewrites` you could replace by `rewrite` using `esimp` before that. I did try that a couple of times, but `rewrite` still failed after that.

I thought of one possible performance improvement for `krewrite`. Currently, if we rewrite with the equality `f x y = z`, then we try to match `f x y` with every subterm of the goal which has as head `f`. So we try to match it with `f a b`, but also with `f a`, `f` and `f a b c`. I think we should change it that it only tries to match `f x y` with a subterm of the goal where `f` is applied to exactly two arguments.

leodemoura:
> So we try to match it with f a b, but also with f a, f and f a b c. I think we should change it that it only tries to match f x y with a subterm of the goal where f is applied to exactly two arguments.

This is a good idea. I will do it.

PR #818, opened 2015-09-03
fix(library/data/rat/order): declare decidable_le an instance

robertylewis:
Lean does not infer that rat has decidable weak inequality. There is a theorem rat.decidable_le that is present, but seems to not be tagged as an instance.

Before this commit:

``` lean
import data.rat
open rat

example (a b : ℚ) : decidable (a < b) := _   -- this works
example (a b : ℚ) : decidable (a ≤ b) := _   -- fails to infer the placeholder
example (a b : ℚ) : decidable (a ≤ b) := decidable_le   -- this works
```

Does migrate not carry over instances? I wouldn't be surprised if there were other places in the library we have similar issues.


leodemoura:
> Does migrate not carry over instances?

It currently does not carry over any attribute.
We still have to implement this feature. 
We really want to carry attributes such as `[instance]` and `[simp]`.

PR #833, opened 2015-09-25
Cantor's theorem, and Schroeder-Bernstein

avigad:
Some notes:

(1) I added iff_subst as a [subst] rule in init/quot, so now we can do calculations that substitute propositions.

(2) In data/set/equinumerosity.lean, rather than declare a whole bunch of theorems and definitions "private", I put them in a separate namespace. It would be nice to have a command "private section ... end" that makes everything in a section private.

(3) I think there is a small bug in calc mode: if I replace the justifications in lines 172-174 of equinumerosity.lean by the commented-out versions, there is a problem later in the file. 

  https://github.com/avigad/lean/blob/master/library/data/set/equinumerosity.lean#L172-L174

It seems that the change introduces two extra universe metavariables that are not instantiated later on. Let me know if I should try to produce a smaller example of the problem.

### DIFF
--- a/library/data/set/basic.lean
+++ b/library/data/set/basic.lean
@@ -301,6 +301,9 @@ ext (take x, iff.intro
 theorem mem_sep_iff {s : set X} {P : X → Prop} {x : X} : x ∈ {x ∈ s | P x} ↔ x ∈ s ∧ P x :=
 !iff.refl
 
+theorem sep_subset (s : set X) (P : X → Prop) : {x ∈ s | P x} ⊆ s :=
+take x, assume H, and.left H
+
 /- complement -/
 
 definition complement (s : set X) : set X := {x | x ∉ s}
@@ -310,7 +313,15 @@ theorem mem_comp {s : set X} {x : X} (H : x ∉ s) : x ∈ -s := H
 
 theorem not_mem_of_mem_comp {s : set X} {x : X} (H : x ∈ -s) : x ∉ s := H
 
-theorem mem_comp_iff {s : set X} {x : X} : x ∈ -s ↔ x ∉ s := !iff.refl
+theorem mem_comp_iff (s : set X) (x : X) : x ∈ -s ↔ x ∉ s := !iff.refl
+
+theorem inter_comp_self (s : set X) : s ∩ -s = ∅ :=
+ext (take x, !and_not_self_iff)
+
+theorem comp_inter_self (s : set X) : -s ∩ s = ∅ :=
+ext (take x, !not_and_self_iff)
+
+/- some classical identities -/
 
 section
   open classical
@@ -320,6 +331,12 @@ section
 
   theorem inter_eq_comp_comp_union_comp (s t : set X) : s ∩ t = -(-s ∪ -t) :=
   ext (take x, !and_iff_not_or_not)
+
+  theorem union_comp_self (s : set X) : s ∪ -s = univ :=
+  ext (take x, !or_not_self_iff)
+
+  theorem comp_union_self (s : set X) : -s ∪ s = univ :=
+  ext (take x, !not_or_self_iff)
 end
 
 /- set difference -/
@@ -350,6 +367,8 @@ ext (take x, iff.intro
       (suppose x ∈ s, or.inl this)
       (suppose x ∉ s, or.inr (and.intro H1 this))))
 
+theorem diff_subset (s t : set X) : s \ t ⊆ s := inter_subset_left s _
+
 /- powerset -/
 
 definition powerset (s : set X) : set (set X) := {x : set X | x ⊆ s}
@@ -377,6 +396,12 @@ section
   definition sUnion : set X := {x : X | ∃₀ c ∈ C, x ∈ c}
 
   -- TODO: need notation for these
+
+  theorem Union_subset {b : I → set X} {c : set X} (H : ∀ i, b i ⊆ c) : Union b ⊆ c :=
+  take x,
+  suppose x ∈ Union b,
+  obtain i (Hi : x ∈ b i), from this,
+  show x ∈ c, from H i Hi
 end
 
 end set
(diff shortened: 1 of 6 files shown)


leodemoura:
> (3) I think there is a small bug in calc mode: if I replace the justifications in lines 172-174 of equinumerosity.lean by the commented-out versions, there is a problem later in the file.

I'm not sure the bug is really in `calc`. There is a bug somewhere. 
Yes, an extra universe level is being included.
The example is quite big. I don't have energy do debug it :-(

avigad:
Don't worry about it. I'll try to take things out, bit by bit, to see if I can get a smaller example.

avigad:
I managed to shrink the example considerably. It looks like the problem is in the handling of backticks in calc mode.

The example below works, but it breaks if you replace `Ha₁` by backtick `a₁ ∈ A` after the show. It is the same metavariable problem.

``` lean
import data.set.classical_inverse
open eq.ops classical set

section
  parameters {X Y : Type}
  parameter  {A : set X}
  parameter  {B : set Y}
  parameter  {g : Y → X}
  parameter  {dflt : Y}

  noncomputable definition ginv : X → Y := inv_fun g B dflt
  lemma g_ginv_eq {a : X} (aA : a ∈ A) : g (ginv a) = a := sorry

  lemma h_inj : inj_on ginv A :=
  take a₁ a₂,
  assume Ha₁ : a₁ ∈ A,
  assume Ha₂ : a₂ ∈ A,
  assume heq : ginv a₁ = ginv a₂,
  have eq₁ : g (ginv a₁) = a₁, from g_ginv_eq `a₁ ∈ A`,
  show a₁ = a₂, from calc
    a₁    = g (ginv a₁) : g_ginv_eq Ha₁ -- g_ginv_eq `a₁ ∈ A` doesn't work, eq₁ does
      ... = a₂          : sorry

  theorem foo (H : ∃ b, b ∉ B) : ∃ h : X → Y, inj_on h A :=
  obtain b bB, from H,
  exists.intro ginv h_inj
end
```

I have noticed other problems using backticks in calc mode, e.g. it sometimes finds the wrong assumption. I work around it by using an explicit label instead. If the problem is not easy to fix, perhaps we should warn the user about backticks in calc. (It took me a long time to find the problem initially.)

avigad:
Another funny thing: notice that the line `obtain b bB, from H,` in `foo` does absolutely nothing. If we delete that line, the problem goes away. We need both that line and the backticks in `calc` to make the problem appear.

leodemoura:
Minor comment: `obtain`-expressions are implemented using `proof ... qed` behind the scenes.
The idea is to force `H` to be elaborated before we process the rest (i.e., `b nB`, and the "body" of the `obtain`). We do that because we need to know the type of `H` to be able to compile the rest (e.g., we need to know how many arguments the constructor takes). So, having an `obtain`-expression (even if it is a useless one) changes the dynamics of how the declaration is elaborated. Since we are using `proof ... qed` behind the scenes, the `obtain` is injecting a "cut" in the elaboration search space.

avigad:
That makes sense. 

In case it helps, here is some more information. With the version that works, I added

``` lean
set_option pp.all true
reveal h_inj
print h_inj
```

after the proof of `h_inj`, and I recorded the result. Then I replaced `Ha₁` with the backtick `a₁ ∈ A`, which causes the error, and again recorded the result. The pretty-printed terms are /exactly/ the same:

```
print result:
theorem h_inj : ∀ {X : Type.{l_1}} {Y : Type.{l_2}} {A : set.{l_1} X} {B : set.{l_2} Y} {g : Y → X}
{dflt : Y},
  @set.inj_on.{l_1 l_2} X Y ginv A :=
λ (X : Type.{l_1}) (Y : Type.{l_2}) (A : set.{l_1} X) (B : set.{l_2} Y) (g : Y → X)
(dflt : Y) (a₁ a₂ : X) (Ha₁ : @set.mem.{l_1} X a₁ A) (Ha₂ : @set.mem.{l_1} X a₂ A)
(heq : @eq.{l_2} Y (ginv a₁) (ginv a₂)),
  have eq₁ : @eq.{l_1} X (g (ginv a₁)) a₁, from @g_ginv_eq a₁ Ha₁,
  show @eq.{l_1} X a₁ a₂, from
    @eq.trans.{l_1} X a₁ (g (ginv a₁)) a₂
      (@eq.symm.{l_1} X (g (ginv a₁)) a₁ (@g_ginv_eq a₁ Ha₁))
      (@sorry.{0} (@eq.{l_1} X (g (ginv a₁)) a₂))
```

I used diff to confirm that my eyes weren't tricking me. I even repeated the experiment outside the section, to print all the parameters, and again they were the same.

```
print result:
theorem h_inj : ∀ {X : Type.{l_1}} {Y : Type.{l_2}} {A : set.{l_1} X} {B : set.{l_2} Y} {g : Y → X}
{dflt : Y},
  @set.inj_on.{l_1 l_2} X Y (@ginv.{l_1 l_2} X Y B g dflt) A :=
λ (X : Type.{l_1}) (Y : Type.{l_2}) (A : set.{l_1} X) (B : set.{l_2} Y) (g : Y → X)
(dflt : Y) (a₁ a₂ : X) (Ha₁ : @set.mem.{l_1} X a₁ A) (Ha₂ : @set.mem.{l_1} X a₂ A)
(heq : @eq.{l_2} Y (@ginv.{l_1 l_2} X Y B g dflt a₁) (@ginv.{l_1 l_2} X Y B g dflt a₂)),
  have eq₁ : @eq.{l_1} X (g (@ginv.{l_1 l_2} X Y B g dflt a₁)) a₁, from
    @g_ginv_eq.{l_1 l_2} X Y A B g dflt a₁ Ha₁,
  show @eq.{l_1} X a₁ a₂, from
    @eq.trans.{l_1} X a₁ (g (@ginv.{l_1 l_2} X Y B g dflt a₁)) a₂
      (@eq.symm.{l_1} X (g (@ginv.{l_1 l_2} X Y B g dflt a₁)) a₁
         (@g_ginv_eq.{l_1 l_2} X Y A B g dflt a₁ Ha₁))
      (@sorry.{0} (@eq.{l_1} X (g (@ginv.{l_1 l_2} X Y B g dflt a₁)) a₂))
```

So it seems that there is some subtle difference in the internal representation that is making the elaborator treat them differently.

I don't want to send you on a wild good chase with this. I will make it a point to use backticks in calc and see if I can get the error to occur in other contexts.

PR #839, opened 2015-09-30
Improve pretty printer spacing

Kha:
The recent mailing list discussion about the need to improve the pretty printer's space insertion heuristic got me interested since it sounded like a relatively isolated problem. Well, it quickly became apparent that the problem of inserting a minimal number of spaces to correctly separate tokens can't be solved exactly by using a fixed-size sliding window, so I reverted to implementing a conservative approximation: The pull request separates to adjacent tokens `s` and `t` if
- the last char of `s` and the first one of `t` are both `id_rest` chars, ie `s + t` may be an identifier, or
- the longest prefix match of `s + t` in the token table is not `s`, or
- `s + t` is a prefix of a token (an approximation because of the limited lookahead)
  Looking at the test outputs, the only failure of the heuristic I could find was `[ n]` (and any other lists starting with an identifier that is a prefix of an attribute name), where `[` and `n` are separated because they are a prefix of `[none]`.

Implementing this as a transformation on the pretty-printed format sexpr looked like the most simple way.

Since the pretty printer output with minimal spacing wasn't quite appealing, I've also added some quick proof-of-concept code that allows defining spacing via quoted symbols, eg `infix ` + `` or `notation `[` l:(foldr `, ` (h t, cons h t) nil `]`)`. I've also removed `--ignore-all-space` from the test script for better comparison. Most of the resulting test failures are because of missing hints in the hott library or test itself and the aforementioned `[ n]` issue.

### DIFF
--- a/library/algebra/complete_lattice.lean
+++ b/library/algebra/complete_lattice.lean
@@ -125,8 +125,8 @@ variable [C : complete_lattice A]
 include C
 prefix `⨅`:70 := Inf
 prefix `⨆`:65 := Sup
-infix `⊓` := inf
-infix `⊔` := sup
+infix ` ⊓ ` := inf
+infix ` ⊔ ` := sup
 
 variable {f : A → A}
 premise  (mono : ∀ x y : A, x ≤ y → f x ≤ f y)
--- a/library/algebra/field.lean
+++ b/library/algebra/field.lean
@@ -25,7 +25,7 @@ section division_ring
   include s
 
   definition divide (a b : A) : A := a * b⁻¹
-  infix [priority algebra.prio] `/` := divide
+  infix [priority algebra.prio] / := divide
 
   -- only in this file
   local attribute divide [reducible]
--- a/library/algebra/group.lean
+++ b/library/algebra/group.lean
@@ -36,8 +36,8 @@ structure has_inv [class] (A : Type) :=
 structure has_neg [class] (A : Type) :=
 (neg : A → A)
 
-infixl [priority algebra.prio] `*`   := has_mul.mul
-infixl [priority algebra.prio] `+`   := has_add.add
+infixl [priority algebra.prio] ` * `   := has_mul.mul
+infixl [priority algebra.prio] ` + `   := has_add.add
 postfix [priority algebra.prio] `⁻¹` := has_inv.inv
 prefix [priority algebra.prio] `-`   := has_neg.neg
 notation 1  := !has_one.one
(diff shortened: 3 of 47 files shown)


leodemoura:
Thanks! 
The output produced for many expressions looks much nicer.

However, I found some problems.
1- The `[ n]` issue is quite annoying, and I suspect users will keep complaining this is a bug. Sometimes, the output looks quite weird. Here is an example:

``` lean
quot.sound : ∀ {A : Type} [ s : setoid A] {a b : A}, setoid.r a b → quot.mk a = quot.mk b
```

The extra space between `[` and `s`.
One dirty hack is to replace the `return true` in the end of `needs_space_sep` with `return s1 != "["`. 
I know this is big hack, but I'm inclined to use it.

2- There is an extra space in `let`-expressions.
Example:

``` lean
open nat
check let a := 10 in a + a
```

 produces

``` lean
let a := 10  in a + a : ℕ
```

There is an extra space between `10` and `in`.

I think I fixed this issue by changing the case `format_kind::LINE:` at `format::separate_tokens`.

3- For some reason, I'm getting

``` lean
∃ (x : A),p x : bool
```

instead of 

``` lean
∃ (x : A), p x : bool
```

The notation declaration at logic.lean contains the space hint.

``` lean
notation `∃` binders `, ` r:(scoped P, Exists P) := r
```

I'm still trying to understand what is happening here.

leodemoura:
Please, ignore item `3` above. The test was redefining the `exists` notation without the space hint.

Kha:
I just realized that we can easily simulate a two-token lookahead by storing the intermediate token_table between invocations. We could do infinite lookahead with a queue of them, but I'd very much suspect that a lookahead of 2 should cover all cases. I've added a commit.

> There is an extra space between `10` and `in`.
> I think I fixed this issue by changing the case `format_kind::LINE:` at `format::separate_tokens`.

Thanks, `LINE` should indeed reset `last`. Updated.

leodemoura:
Thanks!

> I just realized that we can easily simulate a two-token lookahead by storing the intermediate token_table between invocations. We could do infinite lookahead with a queue of them, but I'd very much suspect that a lookahead of 2 should cover all cases. I've added a commit.

I agree, 2 should be enough!

leodemoura:
I have integrated your changes in the main branch.
Thanks again!

@fpvandoorn I think the improved pretty printer is quite useful for the HoTT library. 
I've added some spacing hints, but feel free to change them. 
For example, with the hints, we now get

``` lean
q =[p] r
f⁻¹ 
```

instead of

``` lean
q =[ p ] r
f ⁻¹ 
```

fpvandoorn:
Thanks! That is indeed very nice! I'll add more spacing hints myself.

Kha:
Oh, I wasn't expecting it to be merged quite so soon :) . I first wanted to make sure you agree with the general direction, so I hadn't yet touched a few small issues:
- I wasn't sure whether `pp.extra_spaces` was still relevant with the improved behavior. If it is, it could be added to `needs_space_sep`, but that would also separate tokens outside of notations.
- I've added the `utf8_trim` call to the parser for `sep` tokens of `Exprs` actions, but not yet for `term`
- that single todo in `scanner.cpp`

If you want, I can fix these in a second PR.

leodemoura:
> I first wanted to make sure you agree with the general direction

I think the changes are really nice. I saw a substantial improvement in the output produced by the pretty printer. I also liked the "spacing hints", it is simple and easy to use.

> I wasn't sure whether pp.extra_spaces was still relevant with the improved behavior. If it is, it could be added to needs_space_sep, but that would also separate tokens outside of notations.

I agree. We don't need this option anymore. 

> If you want, I can fix these in a second PR.

That would be great!
Many thanks!

PR #845, opened 2015-10-10
Tactic to normalize added numerals

robertylewis:


leodemoura:
We decided to move this code to the typeclass branch.
Rob is currently updating the code to work with the new approach for encoding numerals.
He will make a new pull request.

PR #847, opened 2015-10-13
feat(hott/homotopy) add join operator, and prove two basic lemmas

javra:
I might polish the proof in future, and add the proof for the associativity of join.


javra:
... I guess this is because the folder doesn't exist at all anymore?

PR #848, opened 2015-10-16
lots of category theory

fpvandoorn:
I can rebase this on `typeclass` if necessary (I don't expect much merge conflicts).


leodemoura:
I'm working on the new type class resolution procedure: https://github.com/leanprover/lean/tree/typeclass_new
It would be great if you could rebase on top of that after I finish it.
In the meantime, rebasing on top of typeclass (and fixing problems) would be already super useful.

Thanks!

leodemoura:
subsumed by #860

PR #849, opened 2015-10-19
feat(lean-syntax): fix and extend syntax highlighting

fpvandoorn:
**Now highlighted:** (examples)

```
5  302  2.51 -- doesn't highlight 3.4.5
Σ Π &&
!
◾ ⁻¹ᵉ ≃ ≡
```

**Fixed highlighting:**

```
[parsing-only]
[unfold-full]
begin+
exists.intro
foo.exists
state --tactic at start of a line
 state.foo
foo.by+ 3
```

**Not highlighted anymore:**

```
[priorityfoobar]
[priority ]
```

**Known bugs:**
(1) When you write two keywords separated by a single space, the second won't be highlighted. For example, `if` is not highlighted here:

```
check if x > 0 then a else b
```

The problem is that the keywords `check` and `if` both look whether the next or previous character is not `.`, so both `check` and `if` have the single space between them as part of the expression, which apparently is not allowed. (Emacs doesn't support Lookahead/Lookbehind for regexes.) For the same reason, something like `check 3+5` won't highlight the `3` or `5`.

(2) Parentheses are highlighted when next to a highlighted keyword/numeral, because they are part of the expression.

(3) We don't highlight a numeral ending with a period, even though it is accepted by Lean: `7.`
**Not (yet) highlighted:**
`true` or `false` (after `set_option`)
`_root_` (It doesn't belong to the other classes since it's part of identifiers)

### DIFF
--- a/src/emacs/lean-syntax.el
+++ b/src/emacs/lean-syntax.el
@@ -27,16 +27,28 @@
 (defconst lean-keywords1-regexp
   (eval `(rx word-start (or ,@lean-keywords1) word-end)))
 (defconst lean-keywords2
-  '("by+")
+  '("by+" "begin+")
   "lean keywords ending with 'symbol'")
 (defconst lean-keywords2-regexp
   (eval `(rx word-start (or ,@lean-keywords2) symbol-end)))
 (defconst lean-constants
-  '("#" "@" "->" "∼" "↔" "/" "==" "=" ":=" "<->" "/\\" "\\/" "∧" "∨"
+  '("#" "@" "!" "->" "∼" "↔" "/" "==" "=" ":=" "<->" "/\\" "\\/" "∧" "∨"
     "≠" "<" ">" "≤" "≥" "¬" "<=" ">=" "⁻¹" "⬝" "▸" "+" "*" "-" "/" "λ"
-    "→" "∃" "∀")
+    "→" "∃" "∀" "∘" "×" "Σ" "Π" "~" "||" "&&" "≃" "≡" "≅"
+    "ℕ" "ℤ" "ℚ" "ℝ" "ℂ" "𝔸"
+    ;; HoTT notation
+    "Ω" "∥" "map₊" "₊" "π₁" "S¹" "⇒" "⟹" "⟶"
+    "⁻¹ᵉ" "⁻¹ᶠ" "⁻¹ᵍ" "⁻¹ʰ" "⁻¹ⁱ" "⁻¹ᵐ" "⁻¹ᵒ" "⁻¹ᵖ" "⁻¹ʳ" "⁻¹ᵛ" "⁻¹ˢ" "⁻²" "⁻²ᵒ"
+    "⬝e" "⬝i" "⬝o" "⬝op" "⬝po" "⬝h" "⬝v" "⬝hp" "⬝vp" "⬝ph" "⬝pv" "⬝r" "◾" "◾o"
+    "∘n" "∘f" "∘fi" "∘nf" "∘fn" "∘n1f" "∘1nf" "∘f1n" "∘fn1"
+    "^c" "≃c" "≅c")
   "lean constants")
 (defconst lean-constants-regexp (regexp-opt lean-constants))
+(defconst lean-numerals-regexp
+  (eval `(rx word-start
+	     (one-or-more digit) (optional (and "." (zero-or-more digit)))
+	     word-end)))
+
 (defconst lean-modifiers
   (--map (s-concat "[" it "]")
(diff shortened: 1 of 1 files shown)


leodemoura:
Thanks!

BTW, I've been wondering if it is possible to highlight tactic names only when the occur inside `begin ... end` blocks. That is, the following case would only be active inside `begin ... end` blocks. I searched the web, but did not find any simple solution.

```
     ;; tactics
     ("cases[ \t\n]+[^ \t\n]+[ \t\n]+\\(with\\)" (1 'font-lock-constant-face))
     (,(rx (not (any "\.")) word-start
           (group
            (or "\\b.*_tac" "Cond" "or_else" "then" "try" "when" "assumption" "eassumption" "rapply"
                "apply" "fapply" "eapply" "rename" "intro" "intros" "all_goals" "fold" "focus" "focus_at"
                "generalize" "generalizes" "clear" "clears" "revert" "reverts" "back" "beta" "done" "exact" "rexact"
                "refine" "repeat" "whnf" "rotate" "rotate_left" "rotate_right" "inversion" "cases" "rewrite"
                "xrewrite" "krewrite" "blast" "simp" "esimp" "unfold" "change" "check_expr" "contradiction"
                "exfalso" "split" "existsi" "constructor" "fconstructor" "left" "right" "injection" "congruence" "reflexivity"
                "symmetry" "transitivity" "state" "induction" "induction_using" "fail" "append"
                "substvars" "now" "with_options"))
           word-end)
      (1 'font-lock-constant-face))
```

fpvandoorn:
I also searched around and couldn't find a solution for that... There's also `by` which doesn't specify very well where the tactic ends:

```
open nat
definition constructor : ℕ := 3
check (by constructor + constructor : ℕ)
```

fpvandoorn:
Hmm... Maybe this pull request shouldn't be rebased yet. The first bug is quite annoying. Things like `private definition` or `local attribute` are not highlighted correctly anymore.

I could change it so that we only check whether there is a period _after_ a keyword to decide whether we should highlight it? That will at least stop highlighting `exists.intro` (the only definition in the library which seems to have a keyword in it). 

But maybe there is a better way to do it? So the problem is that we want to highlight keywords such as `exists` but not if there is a `.` before or after it. In my pull request this is done, but now the character before and after `exists` are part of the matched expression, which means that the patterns in `private definition` overlap, which means that they are not both highlighted. Any ideas @leodemoura or @soonhokong?

fpvandoorn:
Ah, I see that the regular expression used for tactics do exactly what I want. I will incorporate this for the other classes.

fpvandoorn:
Update: I was mistaken; the regular expression for tactics does _not_ solve the issue. It worked, because it didn't check for a period after a tactic (so `state` would be highlighted in `state.foo`).

However, it did solve bug (2), which I incorporated in the other regexes in 8ebc839.

fpvandoorn:
@soonhokong, do you know if we can modify what `word-start` and `word-end` means in `lean-mode`, so that periods don't start/end words? I tried adding the following to the syntax table, but that doesn't seem to work:

```
(modify-syntax-entry ?. "w" st)
```

soonhokong:
@fpvandoorn: I think that should work. What's the latest version that you're working on? I'll test something from there.

fpvandoorn:
My Emacs version is

```
GNU Emacs 24.3.1 (x86_64-pc-linux-gnu, GTK+ Version 3.10.7)
```

fpvandoorn:
Or if you mean the version of Lean I'm working on, it's  8ebc839

fpvandoorn:
Or a commit where I added the period to the syntax table: 3225d9cc437c3d4da226bc1893255b1e16660fa6
If you can ensure that none of this is highlighted here, then `.` is counted as a word-character:

```
begin. .begin .begin.
```

soonhokong:
@fpvandoorn: thanks, I'll test something on your commit.

soonhokong:
@fpvandoorn: it's because of https://github.com/leanprover/lean/blob/master/src/emacs/lean-syntax.el#L117-L119. Note that there is `.` in line 119. If you remove it (`.`), it should work as you expected.

fpvandoorn:
Thanks! That was indeed the problem. This makes the changes much less a hack. I'm testing it now, I'll make a final pull request later.

soonhokong:
:+1:

fpvandoorn:
I updated the commit, and I think it's now good to go. I haven't found any highlighting bugs in this commit (although there is undoubtedly some notation which are not yet highlighted).

leodemoura:
Subsumed by #860 
Remark: the commit has been integrated into the `typeclass` branch.

PR #850, opened 2015-10-20
Update numeral normalizer for typeclass branch

robertylewis:
The norm_num tactic should now normalize numeral addition and multiplication using the new encodings.


leodemoura:
Merged in the typeclass branch

PR #854, opened 2015-10-23
Typeclass

avigad:
I protected many theorem names. The general philosophy is that since "algebra" will often be open, theorems in algebra should not become overloaded when other namespaces are opened. So, any theorem in a namespace that has the same name as a theorem in algebra, or even close, should be protected.

For example, many theorems in pnat had to be protected. Many theorems in nat involving "sub" had to be protected, like "sub_zero" and "mul_sub_left_distrib". Similarly for theorems with the name "div", like "div_zero" and "mul_div_cancel".

The one exception is theorems with "dvd". Some of the nat theorems have names that conflict wit the algebra ones. I left them they way they are because for now we will mostly use "dvd" with nat and int.

Changing all this was a lot more work than I thought it would be. I think I got almost everything. 

I think most of the tactics proofs in real can now be simplified by eliminating "algebra." or "rat." from the theorem names. I did some of them. But I did not have the energy to do them all. 

I still want to go back and make the "div" in nat and int and instance of the div class, so we will write "m / n" on the integers. I will do that soon; I don't have the energy for it now.

I will also update the tutorial soon.


avigad:
I cleaned things up so that now we have

  has_div, div, /
  has_dvd, dvd, |
  has_mod, mod, %

No more divide, divides, division, ... nonsense. The expressions look really sharp. 

I could fix all the tests but these four:

```
     18 - lean_print_notation (Failed)
    147 - leantest_793a.lean (Failed)
    245 - leantest_793b.lean (Failed)
    882 - leanittest_num2.input (Failed)
```

I cannot find the first. The second and third parse floating point constants; the problem is that division is now named "div". The fourth always fails on my computer (I don't know why).

PR #861, opened 2015-10-31
Unelaborate

dselsam:


fpvandoorn:
What is a partial explicit expression? How would `@f` work with the new notation?

dselsam:
On this branch, `@@` means what `@` used to mean, and `@` means only require the implicit arguments that are Pi types. Here is an example:

``` Lean
constants (f : Π {T : Type} {a : T} {P : T → Prop},
                     P a → Π {b : T} {Q : T → Prop}, Q b → Prop)
constants (T : Type) (a : T) (P : T → Prop) (pa : P a)
constants (b : T) (Q : T → Prop) (qb : Q b)

check @@f T a P pa b Q qb -- Prop
check @f P pa Q qb -- Prop
check f pa qb -- Prop
```

How often is the 'partial_explicit' all you need, and how often do you think you need the original full explicit? The notation for the latter is obviously more cumbersome now.

leodemoura:
@fpvandoorn The main motivation for the change is the "unelaborator".
This command will allow us to pretty print fully elaborated terms in a compact and reliable way.
By reliable, I mean the elaborator does not need full higher order unification to (re-)elaborate the result.
This is useful when repairing libraries.

dselsam:
And also as a way around the following pattern: `@f _ _ _ _ ho_implicit explicit`

leodemoura:
Example:

``` lean
variables a b c : nat
variable   H1 : a = b
variable   H2 : a + b + a + b = 0
check eq.rec_on H1 H2    -- b + b + b + b = 0
check @eq.rec_on (fun x, x + b + a + b = 0) H1 H2  -- b + b + a + b = 0
-- using the old approach
check @@eq.rec_on _ _ (fun x, x + b + a + b = 0) _ H1 H2  -- b + b + a + b = 0 
```

dselsam:
as opposed to

``` Lean
check @@eq.rec_on _ _ (fun x, x + b + a + b = 0) _ H1 H2
```

avigad:
I get it -- it seems reasonable. Though I think I rarely used @ to give a predicate. More often, it was giving a type or a parameter to help the elaborator. I'll have to get used to the change.

dselsam:
@avigad If you think of a less ugly alternative to `@@`, please let us know.

leodemoura:
@avigad The idea for the unelaborator is the following.
Suppose a term `t` is broken after we made a change. Then, we recompile the file using the old version, but annotate `t`. For example, we add `[| t |]`. This annotation will instruct Lean that it should pretty print `t` after elaboration, and 

1- use `@` for any sub-term that contains a higher-order implicit argument.
2- disable overloaded notation _and_ notation that contains implicit arguments (e.g., `H1 ▸ H2`)
3- use full-name whenever the short one is overloaded
4- show coercions

The resulting term is much smaller that the one we get with `pp.all true`, and it is easy to (re-) elaborate it. 

The `@` is a must have when pretty-printing any proof that makes heavy use of `rewrite` or uses the `induction` tactic.

avigad:
Even setting that application aside, it seems to be useful: sometimes we want to input a term by adding only the ho information, and often it will be useful to pretty print just the "hard" parts of the term.

But I still have to get used to "@@". I'll see if I can think of something better.

One idea is that we can use "@" for the old meaning, and "@!" for the new one. The intuition for the new one is "make everything explicit, but them make easy arguments implicit". The old one will still have the same meaning as in Coq, and we won't have to change as much in the tutorial.

But that's just a quick thought.

leodemoura:
@avigad We also considered `@!` 
It is confusing because `!f` is the dual operation ("consume _explicit_ arguments").

leodemoura:
@avigad, I don't think it is hard to update the tutorial.
I fixed all tests in less than 10 min using a `sed` script.

fpvandoorn:
Hmmm.. Interesting. Yeah, that could be useful sometimes, but I think that most of the time that I use the old `@` I cannot use the new `@` instead. 

For this message, let me write `@@` for the old behavior and `@` for the new one (although I'd prefer to use `@` for the old behavior and a new symbol for the new behavior).

There are two reasons for me to use `@@`. The most obvious reason is to give implicit arguments explicitly. However, I think most of the time I want to give an argument explicitly it's _not_ a Pi-type. I think most of the times that I want to give arguments explicitly it's a type class argument which is too hard to infer.

A second reason to use `@@` - which is more common for me - is when I want to partially apply a function.  For example `ap` is the term with type

``` lean
ap : Π ⦃A : Type⦄ ⦃B : Type⦄ (f : A → B) {x y : A}, x = y → f x = f y
```

and maybe I want to say something about `(@@ap A B f x y : x = y → f x = f y)` (i.e. that this function is an equivalence under certain circumstances). Note that writing `ap f` is ambiguous, because that has type `?x = ?y → f ?x = f ?y`. Also, very common is something like `abbreviation iso_of_path_equiv := @@category.iso_of_path_equiv`. In both these cases, using `@` wouldn't work.

Overall, I'd guess that more than 80-90% I want to use `@@` I cannot use `@` for the same purpose. So I strongly think that we should not change the old `@` to `@@`. I do see the uses for the unelaborator and for the ability to pretty print terms using the new `@`. That sounds useful. But it sounds like the new `@` will be mostly printed, and less often typed explicitly by the user, which means it could be a harder-to-type unicode character.

leodemoura:
Ok. So, le'ts restore the previous meaning.
We keep `@` for explicit as before.
We use `@@` and some shorter unicode version for partial explicit.
@dselsam I will retract the library updates. I make the changes, ok?

dselsam:
@leodemoura Yes, and I agree with the change.

leodemoura:
Ok, I reverted it.
https://github.com/leodemoura/lean/commits/typeclass

`@` is still explicit
and
`@@` is the new operator for partial explicit

I didn't find a reasonable unicode character for `@@`. Suggestions are welcome :-)

avigad:
This sounds good to me. I agree with Floris that it is more common to want the old @, and so I am glad that is not changing.

dselsam:
How about ⌾?

leodemoura:
It has been integrated into the typeclass branch.
Feel free to reopen if I forgot something.

javra:
Do you also intend to add such a modifier making only type class arguments explicit?

leodemoura:
Jakob,

We have discussed about it, but, for now, we will use `@` and `_` when the type class argument has to be provided. The main motivation for `@@` is `eq.rec`. The rewrite tactic generates a bunch of nested `eq.rec` terms, and it is to verbose to print them in full detail.

Best,
Leo

PR #871, opened 2015-11-13
feat(library/data/complex): add complex numbers

avigad:
When working on this, I came across an annoying problem. When we use `rewrite` with expressions like `x * y`, it doesn't always work, because there multiple instances of `has_mul` that are definitionally equal but not the same.

For example, when we define the reals, we define an instance of `has_mul` for the reals. Then we make the reals an instance of `comm_ring`. When we then invoke rewriting lemmas from algebra, they refer to the instance of `has_mul` derived from the instance `comm_ring`

The problem is evident in lines 346 onwards in complex.lean. If we change `krewrite` to `rewrite`, we get the error message below.

A workaround is to use `krewrite` instead of `rewrite`. But is there a better way? Am I missing something? I have not come across this problem before, which makes me worry that I have done something wrong.

``` lean
invalid 'rewrite' tactic, rewrite step failed using pattern
  of_real (@mul real 11.source.to.has_mul_3 ?M_1 ?M_2)
matching failures in the goal
  -fail to unify equation source
    of_real (@mul real 11.source.to.has_mul_3 ?M_3 ?M_4)
  with subterm
    of_real (@mul real real_has_mul (real.of_rat a) (real.of_rat b))
proof state:
a b : rat
⊢ @eq complex (of_real (@mul real real_has_mul (real.of_rat a) (real.of_rat b)))
    (@mul complex 11.source.to.has_mul_4 (of_rat a) (of_rat b))
```

### DIFF
--- a/library/data/complex.lean
+++ b/library/data/complex.lean
@@ -1,0 +1,375 @@
+/-
+Copyright (c) 2015 Jacob Gross. All rights reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: Jacob Gross, Jeremy Avigad
+
+The complex numbers.
+-/
+import data.real
+open real eq.ops algebra
+
+record complex : Type :=
+(re : ℝ) (im : ℝ)
+
+notation `ℂ` := complex
+
+namespace complex
+
+variables (u w z : ℂ)
+variable n : ℕ
+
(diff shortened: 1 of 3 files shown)


leodemoura:
Don't worry about this kind of issue.
Daniel and I know how to address this problem in blast/simp.

avigad:
That's good to hear! I will sleep better now.

PR #894, opened 2015-11-21
algebra, lists and set quotients

fpvandoorn:


fpvandoorn:
I added some commits which complete the definition of `two_quotients`. This allows us to conveniently define HITs with 2-paths in the HoTT library. It is currently used to define two hits: reflexive quotients and the torus, but I'm also working on the groupoid quotient.

PR #899, opened 2015-11-26
Associativity of join

javra:
The proof was mostly ported from Evan Cavallo's proof in HoTT-Agda. He doesn't respond to my mails, so if @fpvandoorn or @avigad see him, could you ask him if that's okay and how we should credit him?

If the file is too slow for the standard library we could also put it somewhere else? Also you are welcome to squash the commits. It's that many because I was working very irregularly on this, and on two computers.

Some lemmas in `join.hlean` might be useful for other things, in which case one should move them to the corresponding file in `cubical`.


hott/cubical/square.hlean
```diff
@@ -76,14 +82,14 @@ namespace eq
...
-  infix ` ⬝ph `:75 := eq_hconcat
-  infix ` ⬝pv `:75 := eq_vconcat
-  postfix `⁻¹ʰ`:(max+1) := hinverse
-  postfix `⁻¹ᵛ`:(max+1) := vinverse
+  infix `⬝h`:75 := hconcat --type using \tr
+  infix `⬝v`:75 := vconcat --type using \tr
+  infix `⬝hp`:75 := hconcat_eq --type using \tr
+  infix `⬝vp`:75 := vconcat_eq --type using \tr
+  infix `⬝ph`:75 := eq_hconcat --type using \tr
+  infix `⬝pv`:75 := eq_vconcat --type using \tr
```
fpvandoorn: Please keep the spaces around these concatenations. They tell the pretty printer to print spaces around this token.

fpvandoorn:
Very nice! I'll ask Evan next time I see him, but I think what you currently do is fine.

hott/cubical/square.hlean
javra: Ah, thanks, that wasn't a concious decision but result of a merge (I put the `type using` there, before the spaces were there :wink: ).

javra:
No idea, why the last Travis check exceeds the 50 min. mark...

PR #902, opened 2015-11-30
Start of bitvectors in standard library.

joehendrix:
This contains the start of a module for bitvectors in the classical library that use the tuple data type.  This also adds additional combinators for the list and tuple data types.

### DIFF
--- a/library/data/bool.lean
+++ b/library/data/bool.lean
@@ -127,4 +127,6 @@ namespace bool
 
   theorem eq_ff_of_bnot_eq_tt {a : bool} : bnot a = tt → a = ff :=
   bool.cases_on a (λ h, rfl) (by contradiction)
+
+  definition bxor (x:bool) (y:bool) := cond x (bnot y) y
 end bool
--- a/library/data/bv.lean
+++ b/library/data/bv.lean
@@ -128,6 +128,22 @@ section arith
   definition bv_has_neg  [instance] : has_neg (bv n)  := has_neg.mk bv_neg
   definition bv_has_mul  [instance] : has_mul (bv n)  := has_mul.mk bv_mul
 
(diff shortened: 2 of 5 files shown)


leodemoura:
@joehendrix Thanks! I'm currently traveling. I will integrate your changes when I get back on Thursday.

joehendrix:
Sounds good.  There's no rush on this.  It's still missing division/modular reduction, and there are no theorems beyond those needed to admit definitions.  I made the pull request to prompt getting feedback on the style early.

avigad:
For what it is worth, the files look beautiful to me. We haven't been using camel case for identifiers much (except for bundled algebraic structures, like CommRing), but I see no good reason to avoid it here, and I can't think of a better name for `mapAccumR`. Many thanks for working on this!

robertylewis:
Agreed with Jeremy, the style looks great! One super-minor point, since I noticed you used dite a few times: there's special syntax for that.

``` lean
definition bv_shl {n:ℕ} : bv n → ℕ → bv n
  | x i :=
     if le : i ≤ n then
          (let r := dropn i x ++ replicate ff in
           let eq := calc (n-i) + i = n : nat.sub_add_cancel le in
           bv_cong eq r)
    else
          (bv_zero n) -- in the context here you have le : ¬ i ≤ n
```

Not that there's anything wrong with how you did it, but if the syntax is there, might as well make it known!

joehendrix:
Thanks for the pointers.  I didn't know about the dite syntax, so I've patched the code to use that now.  With regards to mapAccumR, I took that from an function in the Haskell base library with the same name.  I do like that Lean uses underscores mostly, but it didn't seem to work here.  mapAccumR is simultaneously performing a map and accumulate, so perhaps there's some similarity with the bundling of algebraic structures example.

PR #907, opened 2015-12-03
Unit propagation and simplifier-based preprocessing

dselsam:


library/algebra/simplifier.lean
```diff
@@ -120,4 +120,23 @@ attribute numeral_helper.mul_bit1_helper1 [simp]
 
 end numeral
 
+namespace unit_simp
+variables {A B C : Prop}
+
+lemma and_imp [simp] : (A ∧ B → C) ↔ (A → B → C) := sorry
+lemma or_imp [simp] : (A ∨ B → C) ↔ ((A → C) ∧ (B → C)) := sorry
```
leodemoura: We should remove these `sorry`'s. As we discussed it before, it is really bad for the project to have `sorry`s in the master branch. Moreover, these lemmas are very easy to prove. It looks really embarrassing to have them in the standard library.

BTW, it also looks strange to have them in `algebra`.

PR #908, opened 2015-12-03
feat(library/blast/imp_extension): imperative branch extensions

dselsam:


src/library/blast/blast.cpp
```diff
@@ -551,6 +580,62 @@ class blastenv {
...
+        lean_assert(state_id < m_imp_extension_entries.size());
+        imp_extension_entry & e = m_imp_extension_entries[state_id];
+        imp_extension_state * ext_state = e.m_ext_state.get();
+        imp_extension * ext_of_curr_state = static_cast<imp_extension*>(&curr_state().get_extension(e.m_ext_id));
+        lean_assert(e.m_ext_of_ext_state);
+        imp_extension * ext_of_ext_state = e.m_ext_of_ext_state;
+
+        list<imp_extension*> curr_state_path = get_ext_path(ext_of_curr_state);
+        list<imp_extension*> ext_state_path = get_ext_path(ext_of_ext_state);
+
```
leodemoura, line 619: You are allocating a list for each `get_ext_path`, then you reverse one of them.
You can avoid this overhead by storing the path into a buffer<imp_extension*>.
Buffers can be easily traversed forwards/backwards with no overhead.
Our `buffer` class also avoids heap allocation if the number of elements is small.

PR #910, opened 2015-12-03
fix(library/abstract_expr_manager): only squash Cast subsingletons

dselsam:


src/library/abstract_expr_manager.cpp
```diff
@@ -25,13 +25,19 @@ unsigned abstract_expr_manager::hash(expr const & e) {
...
-                h = ::lean::hash(h, hash(args[i]));
-            });
+        if (!f_congr) {
+            for (expr const & arg : args) h = ::lean::hash(h, hash(arg));
+        } else {
+            int i = -1;
+            for_each2(f_info.get_params_info(), f_congr->get_arg_kinds(),
+                      [&](param_info const & p_info, congr_arg_kind const & c_kind) {
+                          i++;
+                          if (p_info.is_subsingleton() && c_kind == congr_arg_kind::Cast) return;
```
leodemoura: I think we can simplify and avoid the `m_fun_info_manager`. We only need check whether the argument is a Cast argument or not. There is no need to check whether it is a subsingleton.

PR #919, opened 2015-12-09
Port algebra and number systems

fpvandoorn:
Overall I'm quite happy: porting didn't break too much files. There are a couple of places where I had to give the elaborator extra hints, for example because it couldn't unify `1` with `succ 0` (a couple of examples are in `homotopy/sphere.hlean`).


fpvandoorn:
Some comments:
- I changed two files in `src`: `src/frontends/lean/elaborator.cpp` and `src/frontends/lean/notation_cmd.cpp` to ensure that the HoTT library also uses the new way of parsing numerals. This might make some code obsolete, which you might want to remove, Leo. (I also commented out the code instead of deleting it; I will fix that tomorrow.)
- I made a couple of compromises for the HoTT library: 
  - Almost all algebra-stuff is in the `algebra` namespace. I didn't want to overload too many things, such as `inv_inv`, which we also have for equality. I will test at some point whether this is really necessary.
  - `+` is permanently overloaded as `add` and `sum` (the sum of two types). I think it was unacceptable to also put `+` (`add`)  in `algebra`, because then you had to open `algebra` to even write  `postfix `#`:(max+1) := foo` On the other hand, `+` is also used all the time for `sum`.
- I still have to use the classes `has_zero` and `has_one` in a couple of places where I currently define the notation `0` or `1`.

avigad:
Wow, that was fast! Was the port entirely uneventful? I see that you used the porting script, and that you use the new elaborator in the algebra files. I take it you haven't noticed any performance problems?

Here is a funny side effect of the port:

  https://github.com/fpvandoorn/lean/blob/porting/hott/algebra/field.hlean#L6

I plan to work on modules and vector spaces over the winter break. Should I do them in the standard library and port them to HoTT using the script?

PR #923, opened 2015-12-11
pp.all turns off numerals

dselsam:


tests/lean/608.hlean.expected.out
```diff
@@ -1,11 +1,11 @@
 definition id [reducible] : Π {A : Type}, A → A :=
-λ (A : Type) (a : A), a
+λ (A : Type) (H : A), H
```
dselsam, line 2: This test is broken on master on my machine. Does anybody know what happened here?

src/frontends/lean/pp.cpp
```diff
@@ -280,7 +280,7 @@ void pretty_fn::set_options_core(options const & o) {
     m_purify_metavars = get_pp_purify_metavars(o);
     m_purify_locals   = get_pp_purify_locals(o);
     m_beta            = !all && get_pp_beta(o);
-    m_numerals        = get_pp_numerals(o);
+    m_numerals        = !all && get_pp_numerals(o);
```
dselsam, line 283: This line is the only semantic change in the pull request.

tests/lean/608.hlean.expected.out
fpvandoorn, line 2: No idea, I'm quite sure that this test passed after I changed it 2 days ago.

leodemoura, line 2: I know what is going on. Floris recently defined `id` at `logic.hlean` like we do in the standard library. The `id` declaration contains the subterm `fun (a : A), a`. Internally, this subterm is represented as `lambda a #0, #0`. The `a` is just a label for pretty printing. This file also contains the term `assume H : a, H` (in the proof of `iff.refl`). Internally, this term is represented as `lambda H #0 #0`. Both terms are identical modulo the label used for pretty printing. When the `.olean` is generated we maximally share the terms. We ignore label information in the process, and both terms are merged. To avoid this kind of issue, we would have to take the names into account when performing the max-sharing step.

leodemoura, line 2: I have changed the export procedure. Now, it takes the binder names into account.

PR #928, opened 2015-12-14
[WIP] Transparent let tactic

Kha:
I have changed `let` to `note` only where necessary since it seemed like less work. However, there is a puzzling test failure left:

```
-- testing 608.hlean
2c2
< λ (A : Type) (H : A), H

---
> λ (A : Type) (a : A), a
8c8
< λ (A : Type) (H : A), H

---
> λ (A : Type) (a : A), a
```

Apparently the name of the second argument of `id` is not being correctly preserved, but I can't even begin to imagine which part of the commit could trigger this behavior.


leodemoura:
> I have changed let to note only where necessary since it seemed like less work. However, there is a puzzling test failure left:

@dselsam has also experienced this problem.
See comments here: #923 

This is not a problem with your commit. The problem happened because you have re-compiled the library. I've submitted a fix for this problem: https://github.com/leanprover/lean/commit/999f23cbc077b9a2a93d4dfbe1abb63678a95354

PR #932, opened 2015-12-18
feat(library/blast/arith): tools for arithmetic

dselsam:
No pressure to merge this yet -- there are a few commented-out sections, a few ugly proofs, and a few minor issues with TODOs. But, it passes all tests in Release mode. It is also fully rebased.


avigad:
@dselsam This looks great! 

A small comment -- our convention is to name a theorem A -> B as "B_of_A" rather than "A_of_B", so a few names in algebra/ordered_ring.lean are backwards. I'll be happy to rename things. Will anything bad happen if I do? (I.e. are any of the names hard-coded into the automation?)

dselsam:
@avigad Thank you for the feedback. I thought the takeaway from our last pull-request conversation was that I "write I see, left to right, ignoring ASTs", but I see it is more complicated than that. 

I appreciate the offer to fix it, but then I'll never learn how the naming scheme works so I'd rather try to do another pass myself. Also, these names are used in the C++ code so several spots will need to change.

avigad:
Yes, it was my fault. I was fixated on expressions and forgot about implications. The conventions are described here: https://github.com/leanprover/lean/blob/master/doc/lean/library_style.org, but the policy for implications isn't clear there either (I'll fix it).

Thanks for all the hard work. We're really excited about the automation.

PR #934, opened 2015-12-20
fix(extras/depgraph): calculate lean executable path from PATH or guess

UlrikBuchholtz:
@soonhokong: I noticed that the path to the lean executable was no longer being calculated correctly, so I've fixed that (now it also tries the $PATH first).


soonhokong:
@UlrikBuchholtz, thank you. It's merged by 71cad16c32f2e83e199b517dbff474c72275f0fe.

PR #936, opened 2015-12-21
miscellaneous housecleaning 

avigad:


avigad:
I keep adding things to this pull request. There is no rush to merge, but here are some of the things that are included:

o A start on modules and vector spaces, incorporating definitions by Nathaniel Thomas.

o Normed spaces and Banach spaces. If we develop limits and sums for those, it will subsume the reals, the complexes, R^n, C^n, function spaces, etc.

o A start on extended reals (with Jacob Gross), towards developing measures

o A start on topology (with Jacob Gross)

PR #937, opened 2015-12-22
feat(library/logic/connectives)

JacobGross:
add constructively valid DeMorgan law


JacobGross:
Whoops! My apologies.

avigad:
No problem! It may be helpful to have the implications stated separately for this and other identities. In that case, they should be named `not_and_not_of_not_or`, `not_or_of_not_and_not`, etc., and they should go in the `identities` file. But maybe we should wait until someone actually needs them.

PR #938, opened 2015-12-22
fix(library/algebra/group.lean): mark add.assoc as definition

soonhokong:
The previous version doesn't compile with `--to_axiom` option and
as a result it breaks lean.js build. Specifically, line 614 of the same file
generates an error. I found that this patch is an easy way to fix this problem.


avigad:
@soonhokong There is no need for that: the problem is just a typo. Deleting `add.assoc` (and the comma before it) from line 614 fixes the problem.

soonhokong:
@avigad, thank you for checking this. f1daf8f deleted `add.assoc` from line 614 and it solved the problem as you mentioned.

PR #939, opened 2015-12-24
Arithmetic

dselsam:
There is still a lot to be done, but it is in a consistent state, is fairly well tested and does not seem unreasonably slow. I think it is ready to merge. I am going on vacation tomorrow until 2016 and I will not touch it again until then.

A few comments:
- I have no idea how to write a state-of-the-art arithmetic and computer-algebra library, let alone a proof-producing one. I am completely winging it. I don't even have good references, and am working out tricky recursions and invariants by hand even for problems that may have standard algorithms. Any pointers to books, papers, or relevant code-bases would be much appreciated.
- The core of this pull request is a decision procedure for linear inequalities, based on the approach described at https://www.cs.utexas.edu/~moore/publications/non-linear.pdf. It can already handle fairly beastly inequalities over fields, as in https://github.com/dselsam/lean/blob/arith/tests/lean/run/blast_arith_heuristic.lean#L66-L77. Right now it is restricted to linear reasoning, but I plan to add some support for heuristic non-linear reasoning (as described in the paper) over time. Thus it currently treats `x * y` as a black-box. It is also missing a bunch of basic features. For example, it doesn't communicate new equalities to blast even when it finds them, it doesn't try to prove an inequality by contradiction, and it doesn't do case-analysis on inequalities.
- The most complicated part of the module so far is the proof-generating field-normalization (https://github.com/dselsam/lean/blob/arith/src/library/blast/arith/normalize.cpp). We wanted to be able to prove a term is equal to its normal form substantially faster than we were able to with the (much more generic) simplifier. The current implementation is faster, but unlike the simplifier, it does not track `refl` proofs, and as a result the proof terms are much bigger than they need to be. This optimization is doable but does not seem to be high priority.
- We also have a non-proof-generating normalization procedure that populates a specialized data structure for representing polynomials. The arithmetic procedure itself works with specialized data structures, and only tries to prove anything in terms of actual Lean expressions if it has found a contradiction.
- I am using a light wrapper around Rob's super-fast numeral library. I expect `div` to be marked reducible in blast, and I have been using `inv` everywhere instead. The fact that a normalized numeral can be a product (e.g. `3 * inv 5`) adds some complication throughout the module though, and I think a better approach might be to define a new, non-reducible function that either means `div` or even just `mul`, and for blast to use it to represent normalized numerals.
- I have been warned that this idea will be met with overwhelming resistance, but I feel strongly that every application of `inv` should carry with it a proof that the element is indeed invertible. We can use the same `inv` whether the inverse is total or partial:

``` Lean
structure has_inv [class] (A : Type) := (invertible : A → Prop) (inv : ∀ a, invertible a → A)
definition invertible {A : Type} [s : has_inv A] : A → Prop := has_inv.invertible
definition inv {A : Type} [s : has_inv A] : ∀ (a : A) {H : invertible a}, A := has_inv.inv

structure group [class] (A : Type) extends monoid A, has_inv A :=
(all_invertible : ∀ a, invertible a)
(mul_left_inv : ∀a, mul (@inv a (all_invertible a)) a = one)

structure division_ring [class] (A : Type) extends ring A, has_inv A, zero_ne_one_class A :=
  (nonzero_invertible : ∀ {a:A}, a ≠ zero → invertible a)
  (mul_inv_cancel : ∀(a) {H : invertible a}, mul a (inv a H) = one)
  (inv_mul_cancel : ∀(a) {H : invertible a}, mul (inv a H) a = one)
```

This may seem like it will be extremely burdensome, but we can register a tactic hint for the head `invertible` and a more mature blast should be able to discharge the obligations easily without the user needing to think about it at all.

I think one can make a convincing argument on aesthetics alone, and also on convenience for automation (right now I am tracking all the `_ ≠ 0` sub-goals that need to be solved). As Leo pointed out to me, there is an even more pressing reason to store the proofs with the applications: so we can cache them automatically. Imagine a complicated non-zero proof in the middle of a bigger proof, where the conclusion involves the tricky inverse. Then blast will not be able to use the theorem effectively since it will not be able to reprove that the term is non-zero. If we do not embed the proofs in the applications, then users will need to abstract all non-trivial non-zero proofs into stand-alone lemmas.
- I did not integrate the new lemmas used by the arithmetic module into the standard library yet. For now I am keeping them in their own file, since it is convenient for them to all be in the same place while implementing the module. The only downside right now is that one needs to `import algebra.blast` in order to use/test the module. Since this module is still a long way from being usable, and since the standard library will likely be re-factored anyway as we gradually start using blast, this does not seem pressing.
- There are a few miscellaneous rough edges, which I will iron out over time. Example: the tracing is still haphazard and does not conform to the budding standard for blast-relevant events.


src/kernel/type_checker.cpp
```diff
@@ -445,8 +446,10 @@ void check_no_metavar(environment const & env, name const & n, expr const & e, b
 }
 
 static void check_no_local(environment const & env, expr const & e) {
-    if (has_local(e))
+    if (has_local(e)) {
+        std::cout << "has locals: " << e << std::endl;
```
leodemoura, line 450: Leftover

src/library/blast/arith/heuristic.cpp
```diff
@@ -0,0 +1,552 @@
...
+
+/** Basic data-structures **/
+
+/* Proof trails */
+typedef std::function<expr()> lazy_proof;
+
+class poly;
+struct poly_parents {
+    // TODO(dhs): switch to reference counting once we have [MK_BLAST_RC()]
+    std::shared_ptr<poly> m_p, m_q;
```
leodemoura, line 112: We should not use std::shared_ptr for small objects that are created too often

src/library/blast/arith/heuristic.cpp
```diff
@@ -0,0 +1,552 @@
...
+        m_placeholders(placeholders) {}
+
+    virtual action_result resolve(expr const & pf) const override {
+        buffer<expr> proofs;
+        for (expr const & placeholder : m_placeholders) {
+            optional<expr> pf;
+            {
+                flet<state> save_state(curr_state(), curr_state());
+                curr_state().set_target(mlocal_type(placeholder));
+                pf = mk_backward_strategy()();
```
leodemoura, line 53: Why do you need to invoke backward strategy when closing a branch? Moreover, the user may have marked arbitrary lemmas for backward chaining.
It is bad practice to perform search during resolve.

src/library/blast/arith/heuristic.cpp
```diff
@@ -0,0 +1,552 @@
...
+    }
+    list<poly> get_positives() { return m_positives; }
+    list<poly> get_negatives() { return m_negatives; }
+};
+
+struct heuristic_branch_extension : public branch_extension {
+private:
+    // TODO(dhs): will this cause problems during rebalancing?
+    // Should I wrap this class and make a cheap copy?
+    typedef rb_map<std::vector<atom>, poly_pot, atoms_quick_cmp> poly_pot_map;
```
leodemoura, line 296: You have been using lists. Why did you use a vector here? The map performs copies during insertion.

src/library/blast/arith/heuristic.cpp
```diff
@@ -0,0 +1,552 @@
...
+maintaining a trace of how the new polynomial inequalities can be justified in terms of the old ones.
+
+If we do construct a polynomial inequality of the form [0 < 0, 0 < - <pos>, 0 <= - <pos>], we have
+found a contradiction, and we must start to construct a formal proof of this fact.
+*/
+
+/** Basic data-structures **/
+
+/* Proof trails */
+typedef std::function<expr()> lazy_proof;
```
leodemoura, line 107: I would not use a closure here.

src/library/blast/arith/heuristic.cpp
```diff
@@ -0,0 +1,552 @@
...
+       we store the [hypothesis_idx] of this existing hypothesis. The "linearization"
+       module will be responsible for this stage of the proof.
+       2. A polynomial inequality may be the result of resolving two existing polynomial inequalities
+       together. In this case, we store the two parents along with the scaling factors. To reconstruct
+       the proof, we use the "poly_cancel" lemmas. */
+
+    // Poor man's union
+    optional<lazy_proof>   m_lazy_proof;    // Case 1
+    optional<poly_parents> m_parents;       // Case 2
+public:
```
leodemoura, line 148: The size of an optional T is the size of T even when it is a none.

src/library/blast/arith/heuristic.cpp
```diff
@@ -0,0 +1,552 @@
...
+};
+
+static heuristic_branch_extension & get_heuristic_extension() {
+    return static_cast<heuristic_branch_extension&>(curr_state().get_extension(g_ext_id));
+}
+
+/* HEURISTIC function */
+class heuristic_fn {
+    heuristic_branch_extension &        m_ext;
+    buffer<poly>                  m_todo;
```
leodemoura, line 330: poly is not a small object. We will have to perform copies whenever inserting elements in the buffer.

src/library/blast/arith/normalize.cpp
```diff
@@ -0,0 +1,626 @@
...
+       Polynomial: a sum of monomials
+    */
+    expr_pair normalize(expr const & e) {
+        m_num_steps++;
+        if (m_num_steps > m_max_steps)
+            throw exception(sstream() << "arith::normalize failed, maximum number of steps (" << m_max_steps << ") exceeded");
+
+        expr arg;
+        if (is_numeral_expr(e)) {
+            lean_trace(*g_normalize_trace_name, tout() << "normalize::numeral: " << e << "\n";);
```
leodemoura, line 542: Suggestion: declare macro lean_trace_normalize

src/library/blast/arith/polynomial.cpp
```diff
@@ -0,0 +1,144 @@
...
+    for (monomial const & m : p.get_monomials()) {
+        e = get_app_builder().mk_add(type, e, monomial_to_expr(m, type));
+    }
+    return e;
+}
+
+
+/* Printing */
+// TODO(dhs): use [io_state_stream] so that we can pretty-print
+// issue: printing mpqs
```
leodemoura, line 116: I don't see what is the issue here. We just have to define << for mpq

src/library/blast/congruence_closure.cpp
```diff
@@ -280,7 +280,8 @@ static optional<ext_congr_lemma> to_ext_congr_lemma(name const & R, expr const &
 }
 
 static optional<ext_congr_lemma> mk_ext_congr_lemma_core(name const & R, expr const & fn, unsigned nargs) {
-    simp_rule_set const * sr = get_simp_rule_sets(env()).find(R);
+    simp_rule_sets srss = mk_simp_rule_sets(env());
```
leodemoura, line 283: The refactoring is being stuffed inside the arithmetic commit ?!?

src/library/head_map.cpp
```diff
@@ -29,7 +29,7 @@ head_index::head_index(expr const & e) {
 
 int head_index::cmp::operator()(head_index const & i1, head_index const & i2) const {
     if (i1.m_kind != i2.m_kind || (i1.m_kind != expr_kind::Constant && i1.m_kind != expr_kind::Local))
-        return static_cast<int>(i1.m_kind) - static_cast<int>(i2.m_kind);
+        return unsigned_cmp()(static_cast<unsigned>(i1.m_kind), static_cast<unsigned>(i2.m_kind));
```
leodemoura, line 32: Why did you have to make this change?

tests/lean/run/blast_ematch10.lean
```diff
@@ -1,4 +1,7 @@
-attribute iff [reducible]
+-- TODO(dhs, leo): if `iff` is set to be reducible, then `not_false_iff` becomes
+-- an invalid [simp] rule. Right now we throw an error when blast starts, though in the
+-- future we may just wish to ignore [simp] rules that are no longer valid.
```
leodemoura, line 3: Yes, we should discard invalid simp rules.

leodemoura:
I thought I have already sent you several references for arithmetic procedures, and have summarized the main ones in the slack channel. Regarding normalization, I never seen this step being described anywhere, it is considered too simple. Most articles will say things such as: "we assume all polynomials are in sum of monomials normal form", ...
I have also sent you links to the z3 code that performs these normalizations. Is the code there too confusing?

robertylewis:
I'm still looking through what you've added here, but this seems fantastic! A few preliminary questions.

> Poly pot is just a funny name for lower and upper bounds. Note that the acl2 procedure is essentially Fourier Motkin + heuristics for nonlinear arithmetic.

What was the reason to go with this method for linear problems, instead of something based on simplex/Farkas' lemma? Just that it can be extended to nonlinear problems too? (I'm not saying this is a bad reason! Just curious. I've been thinking about implementing a simplex solver for a while.) My experience using ACL2 for nonlinear inequalities was that it was very sensitive to the format of the problem and to which ACL2 books were open. How configurable do you expect your solver to be?

> The most complicated part of the module so far is the proof-generating field-normalization

Cool. What is the normal form you reduce to? Can this be made available as a tactic? i.e., given a goal `s < t` where s and t are terms in a field, normalize both? I'm happy to take a stab at this after the new year.

> I have been warned that this idea will be met with overwhelming resistance, but I feel strongly that every application of `inv` should carry with it a proof that the element is indeed invertible.

I have to think about this one a little longer, but my gut reaction is indeed resistance, and I'm not sure I see the aesthetic argument. I believe you that it would simplify things for the automation, but it feels like this could get quite confusing in manual proofs. By "complicated non-zero proof in the middle of a bigger proof," do you mean a user-supplied proof `have H : x ≠ 0, from ...`? Why wouldn't this be available as a hypothesis to blast?

leodemoura:
> What was the reason to go with this method for linear problems, instead of something based on simplex/Farkas' lemma?

I suggested the ACL2 procedure to Daniel because it is easier to implement. 
Lev's simplex module is still under development, and Daniel would not really learn much by doing it. 
It is mainly an integration task. This will be done later. Note that the ACL2 part is a small part of the commit.

It is also useful to have a Fourier-Motzkin (FM) procedure even if we have Simplex.
We can use FM to perform quantifier elimination. In Z3, FM is also use for preprocessing.

> Just that it can be extended to nonlinear problems too?

Yes.

> How configurable do you expect your solver to be?

Remark: this is one of many solvers that will be constructed.

> Cool. What is the normal form you reduce to? Can this be made available as a tactic? 

This is possible. However, we didn't implement yet a procedure that converts a blast state object back into a goal.
BTW, all blast related modules are changing too quickly. 
We already have tactics such as `simp` and `blast`, and many more variants can be constructed, but I don't want to advertise them until they are much more stable. 

> it feels like this could get quite confusing in manual proofs.

Right now, it is not feasible to move to this approach. 
We need a very robust blast/simp before we try it.
The rewrite tactic doesn't really work for `inv`/`div` with nested proofs. 
Example: the tactic `rewrite H1` would fail in the following state:

``` lean
a b : real, H1 : a = b, H2 : a ≠ 0, H3 : b ≠ 0 |- @inv real a H2 = @inv real b H3
```

These steps would have to be replaced with `simp`, or the user would have to manually invert `H3`, rewrite and reintroduce it.
After we have a robust blast, we need to go over the `nat`/`int`/`rat`/`real` libraries and simplify all proofs using it. 

That being said, the approach has many advantages. Moreover, if everything works as expected, we will have to write very few manual proofs. Anyway, I think this is not the right time to discuss/consider this change. 

BTW, there is another issue, this approach may be problematic for the HoTT library since it doesn't have proof irrelevance. 

> By "complicated non-zero proof in the middle of a bigger proof," do you mean a user-supplied proof have H : x ≠ 0, from ...? Why wouldn't this be available as a hypothesis to blast?

Suppose we have a lemma `L` whose conclusion contains `a/b`. Moreover, suppose its proof contains a `have b ≠ 0, ...`, and the proof is not simple. Now, when `blast` uses `L` it doesn't have access to the nested proof for `b ≠ 0`. The only way to make it available is to create an auxiliary lemma (i.e., factor out the nested proof).

src/library/blast/arith/heuristic.cpp
dselsam, line 112: I thought the plan was to switch to reference counting once you implemented `MK_BLAST_RC` (see the TODO). Should I implement it with `MK_LEAN_RC` in the meantime?

src/kernel/type_checker.cpp
dselsam, line 450: Yes, my mistake. I thought I had re-checked out the original but I see I did not. Obviously the `cout` was a temporary hack, but it is annoying that the exception does not print the expression. It would be nice to have the message: `Expression <expr> contains the following local constants: <local> ... <local>`. How would you recommend I implement this? With tracing?

src/library/blast/arith/heuristic.cpp
dselsam, line 53: If we do keep proofs with applications of `inv`, this issue will no longer affect `resolve`. Either way, we are going to want an extensible strategy specific to proving `! (_ = 0)`, though the strategy we choose will depend on whether we keep proofs with `inv`, since in the latter case we will want to be more robust at the possible expense of efficiency.

src/library/head_map.cpp
dselsam, line 32: This is the bug we discussed on slack. The result needs to be in `{-1, 0, 1}`.

dselsam:
> Regarding normalization, I never seen this step being described anywhere, it is considered too simple. Most articles will say things such as: "we assume all polynomials are in sum of monomials normal form", ...

What is considered simple in a paper is not necessarily simple for generating machine-checkable proofs efficiently. Example: basic arithmetic obviously does not need to be explained or justified in a paper, and yet the efficient numeral-normalizer is one of the most complex modules in blast so far. Similarly for the heuristic arithmetic paper, it is simple to implement resolving two linear inequalities, and it just happens to be much more complicated to normalize the terms efficiently with proofs.

dselsam:
> I have also sent you links to the z3 code that performs these normalizations. Is the code there too confusing?

It is my current plan to read all the relevant z3 code. We had a misunderstanding though: I thought you specifically suggested I not try to read z3 since it was not designed to be readable. Also, z3 has a much lower standard of proof-generation, and generating proofs is often the most challenging aspect of implementing these procedures in blast.

dselsam:
> Why is this kind of change stuffed inside the arithmetic commit?

It was needed since we were normalizing the simp rules. Note that you made the same change when you implemented this feature: https://github.com/leanprover/lean/commit/b117a10f82ea751c9b932e9e3006d057cc14886e#diff-1bb2394fe3a0f89e80c5e85b2bb6bf34R504

src/library/blast/congruence_closure.cpp
dselsam, line 283: There was a trade-off. It was more convenient to implement it on the arithmetic branch, since I was implementing it at the time just to be able to use `div` lemmas as simp rules for `inv`, and the tests I was using depending on various things in the arithmetic branch. The issue is that the arithmetic branch was still very much in flux, and once it stabilized it would have required some git-fu to separate out the simp rule part. It did not strike me as essential to do this at the time, but I suppose I made the wrong decision. Since you implemented the simp rule normalization yourself on master, I'll remove this part of the commit.

src/library/blast/arith/heuristic.cpp
dselsam, line 107: It is not even helpful now, since we always prove the original un-normalized `0 < ...` eagerly. However, if the linearizer is extended this flexibility may prove useful. I can get rid of it for now and just store the proof itself.

fpvandoorn:
> BTW, there is another issue, this approach may be problematic for the HoTT library since it doesn't have proof irrelevance. 

In HoTT, all inhabitants of `x ≠ 0` are propositionally equal:
https://github.com/leanprover/lean/blob/master/hott/types/pi.hlean#L309

They're not definitionally equal, as in the standard library, though.

>  After we have a robust blast, we need to go over the nat/int/rat/real libraries and simplify all proofs using it. 

One thing which worries me: if we do this and `blast` doesn't work as well in the HoTT library, then we cannot port files anymore from the standard library to the HoTT library.

leodemoura:
@fpvandoorn 

> In HoTT, all inhabitants of x ≠ 0 are propositionally equal:
> https://github.com/leanprover/lean/blob/master/hott/types/pi.hlean#L309
> 
> They're not definitionally equal, as in the standard library, though.

This is good. We have a "HoTT compatible" congruence closure procedure that will be able to handle the new `inv` if the inhabitants of `x ≠ 0` are propositionally equal.

> > After we have a robust blast, we need to go over the nat/int/rat/real libraries and simplify all proofs using it.
> 
> One thing which worries me: if we do this and blast doesn't work as well in the HoTT library, then we cannot port files anymore from the standard library to the HoTT library.

I'm also concerned about this issue. 
Daniel and I have discussed this issue many times.
For me, the main problem is that the libraries are really hard to maintain.
Whenever we make a change in the elaboration procedure, we have to go over the really painful process of fixing proofs. We have many proofs that rely on higher-order unification and are very fragile. 
By replacing them with blast, we avoid all the brittleness associated with higher-order unification.

PR #943, opened 2015-12-31
theorems for sets, and sums and products for ranges of natural numbers

avigad:
The main things in this pull request are these:
- Some more theorems for sets, bijections, finite sets, etc.
- Some properties of intervals, especially intervals of natural numbers
- Sums and products over ranges of natural numbers.

Also, I got rid of finset `singleton`, since it is confusing to have both that and `'{a}` (which is defined to be `insert a ∅`).

A small thing: with the notation for intervals, we can write `finite '(a, b)`, but we have to write `finite ('[a, b])`. See for example here: https://github.com/avigad/lean/blob/master/library/algebra/interval.lean#L135-L158. I could not figure out how to get around that.

I tried varying my proof style, adopting styles that I have seen Rob, Haitao, and others use, in particular for indenting parameters and statements. 

  https://github.com/avigad/lean/blob/master/library/data/nat/bigops.lean

On the plus side, it makes it easier to scan the list of parameters and statements. On the negative side, the indentation is less uniform as you read down the file. (I still haven't decided which I like better.)


leodemoura:
> A small thing: with the notation for intervals, we can write finite '(a, b), but we have to write finite ('[a, b]). See for example here: https://github.com/avigad/lean/blob/master/library/algebra/interval.lean#L135-L158. I could not figure out how to get around that.

You have defined the following notation at `data/set/basic.lean`

```
notation f `'[`:max a `]` := image f a
```

avigad:
Aha, of course, that explains it. That notation is useful too. So I propose we keep things the way they are.

leodemoura:
I'm not sure. I prefer to write `f '[1, 2]` and `h (image g s)` than `f ('[1, 2])`and `h (g'[s])`.
We use intervals more often than the `image`. The `f ('[1, 2])` looks really weird.
It also feels weird that `'[` occurs as a prefix and infix operator.

avigad:
Hmmm. I like the image notation. I have used it a fair amount already (do a `grep -r "'\[" .` in library), and it will come up a lot with topology, measure theory, combinatorics, ... It might be used more than intervals, and the problem only arises when we apply functions to intervals, which is even less common.

But I agree that `finite ('[a, b])` doesn't look great. Let me think about it. Maybe we can come up with some other clever notation for image.

avigad:
Other options:

``` lean
import data.set
open set

variables (A : Type) (f : A → A) (s : set A)

notation f `~[`:max s `]` := image f s

check f ~[s]

notation f `^[`:max s `]` := image f s

check f ^[s]
```

avigad:
Three more: `f :[s]`. `f ![s]` `f ' s`.

Isabelle uses a backtick: https://www.cl.cam.ac.uk/research/hvg/Isabelle/dist/library/HOL/HOL/Set.html

(And happy new year!)

leodemoura:
I like all of them but `f~[s]`. The token `:[` my create problems in the future if we start using `[` for defining  types. 

We can't use backtick like Isabelle since it is a reserved symbol.
`f ' s` looks nice and compact.

avigad:
I like `f ' s`. I'll make that change within the next few days if there are no objections.

avigad:
I spoke too soon. The single tick mark conflicts with the notation for `'(a, b)`:

``` lean
notation `'` `(` a `, ` b `)`     := Ioo a b
```

But we can use `f '' s`, which still looks o.k. to me. 

That leaves various options for inverse image, like `f -' s`, `f '- s`, `f ''- s`, `f -'' s`, `f '⁻¹`, etc.

leodemoura:
I just pushed a commit that allows us to defined tokens such as `'(`. 
https://github.com/leanprover/lean/commit/5feef27c2bc92fdc58082329ab9034b564f2ef18
Note that a redefined the interval notation

``` lean
notation `'(` a `, ` b `)`     := Ioo a b
```

The new/updated restriction is: users cannot define new tokens starting with `(` or ending with `)`.

PR #945, opened 2016-01-04
Replace tactic

robertylewis:
This adds a tactic for replacing subterms of the goal with definitionally equal terms (similar to `change`, but on subterms). This came up in email/discussion with @avigad @fpvandoorn @UlrikBuchholtz Mike Shulman and others.

Usage:

``` lean
open nat

example (H : false) : (1 : ℕ) + 2 = (3 : ℕ) :=
begin
  replace (3 : ℕ) with (2 : ℕ) + 1, -- goal : 1 + 2 = 2 + 1
  reflexivity
end

example (H : false) : (1 : ℕ) + 1 = (2 : ℕ) :=
begin
  replace (1 : ℕ) with succ 0 at {1}, -- goal : succ 0 + 1 = 2
  reflexivity
end
```

### DIFF
--- a/library/algebra/group.lean
+++ b/library/algebra/group.lean
@@ -423,6 +423,24 @@ section add_group
   theorem add_neg_eq_neg_add_rev {a b : A} : a + -b = -(b + -a) :=
   by simp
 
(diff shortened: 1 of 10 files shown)


fpvandoorn:
Great! Thanks!

PR #946, opened 2016-01-04
Add Galois connections

johoelzl:
This is a small theory to play with Lean and learning to prove in it. I want to use Galois connections to show that `generated_by` from topologies / sigma-algebras from Galois connections with the set of open sets / measurable sets.

@avigad: I tried to follow the library conventions I hope the theorem and constant names are okay and the theorems are at the right place. I'm also not absolutely sure if my usage of section/variable/parameter/include is correct. Also I'm not sure about my definition-style for structures, should I put the proofs in a new line?

`complete_lattice_set` is quite strange, but with a working simplifier this should work quite straight forward. I didn't simply unfold set and then use `complete_lattice_Prop` as then `Inf` and `Sup` are not any more definitionally equivalent to `sInter` and `sUnion`.

I tried to use blast or simp, but somehow it never worked. Maybe next month :-)

### DIFF
--- a/library/algebra/algebra.md
+++ b/library/algebra/algebra.md
@@ -10,6 +10,7 @@ Algebraic structures.
 * [interval](interval.lean)
 * [lattice](lattice.lean)
 * [complete lattice](complete_lattice.lean)
+* [galois_connection](galois_connection.lean)
 * [group](group.lean)
 * [group_power](group_power.lean) : nat and int powers
 * [group_bigops](group_bigops.lean) : products and sums over lists, finsets and sets
--- a/library/algebra/complete_lattice.lean
+++ b/library/algebra/complete_lattice.lean
@@ -278,7 +278,7 @@ open eq.ops complete_lattice
 
 definition complete_lattice_fun [instance] (A B : Type) [complete_lattice B] :
     complete_lattice (A → B) :=
-⦃ complete_lattice, lattice_fun,
+⦃ complete_lattice, lattice_fun A B,
   Inf := λS x, Inf ((λf, f x) ' S),
   le_Inf := take f S H x,
     le_Inf (take y Hy, obtain g `g ∈ S` `g x = y`, from Hy, `g x = y` ▸ H g `g ∈ S` x),
--- a/library/algebra/complete_lattice.lean
+++ b/library/algebra/complete_lattice.lean
@@ -276,7 +276,7 @@ end complete_lattice
 section
 open eq.ops complete_lattice
 
-definition complete_lattice_fun [instance] {A B : Type} [complete_lattice B] :
+definition complete_lattice_fun [instance] (A B : Type) [complete_lattice B] :
     complete_lattice (A → B) :=
 ⦃ complete_lattice, lattice_fun,
   Inf := λS x, Inf ((λf, f x) ' S),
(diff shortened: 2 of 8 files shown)


avigad:
Very nice! Welcome aboard!

Our naming conventions are described here:

  https://github.com/leanprover/lean/blob/master/doc/lean/library_style.org

They are designed to make it easy to find theorems using tab completion. Roughly, you describe the goal first, just naming what you see, as far as is needed to describe the theorem. If necessary, you add hypotheses afterwards. So a theorem `A → B → C` is named `C_of_A_of_B`.

The names can look funny at times, but the scheme works really well. I rarely need to browse theories to find what I am looking for; tab completion is almost always enough. The most important thing is to make the prefix obvious.

With this scheme, your theorems `A_implies_B` would be renamed to `B_of_A`. As horrible as it sounds, `is_lub_unique` would be `eq_of_is_lub_of_is_lub`. (Though "unique" comes up a lot; maybe we should adopt the convention of using `f_unique` for this purpose.) 

We use `mem` for set membership. Would it be o.k. to rename `upper_bounds_image` to `mem_upper_bounds_image`? I would name `is_lub_Sup_iff` to `is_lub_iff_Sup_eq` (or even `is_lub_iff`, if you think that's enough).

Do you think users will `open galois_connection` when they want to use galois connections? Then it might make sense to use `protected lemma compose` and `protected lemma id` to avoid overloading `compose` and `id`. (This means users will have to write `galois_connection.compose` and `galois_connection.id` even if the namespace is open.)

The use of sections, parameters, and include looks good to me!

If you want, we can merge this to the repository, and then I will do a pull request with suggested name changes.

robertylewis:
Cool, thanks!

The proof of complete_lattice_set will create an instance of `has_lt (set A)`, which will allow the notation < to be used for subset. The pretty printer will automatically use the last defined notation for a definition:

``` lean
open nat
theorem foo (a b : ℕ) : lt a b := sorry
check foo -- a < b
infix `is_lt`:50 := lt
check foo -- a is_lt b
check λ a b : ℕ, a < b -- a is_lt b
```

I'm not sure in which order the notations for lt and subset are defined, so just to check: does this affect pretty-printing subsets?

(Also, I think the parameter A to complete_lattice_set should be made explicit.)

avigad:
@rlewis1988: `lt a b` and `subset a b` are definitionally equal, but they are different expressions. So theorems that are stated with the subset notation will be printed with the subset notation, and theorems that are stated with < will be printed with <.

robertylewis:
Oh, right, good point! Never mind then.

johoelzl:
Okay, I will be incorporating your suggestions, i.e. renaming theorems, using `private`, and make the parameter to `complete_lattice_set` explicit. I assume the workflow now is to close this pull request, change the history and submit a new one. This is also an exercise for me to learn git's history rewriting :-)

avigad:
That works for me!

For the history rewriting, you can use "git rebase -i HEAD~3", or example, to revise the last three commits. The message that comes up in the editor is informative -- it will let you merge ("squash") or rename commits. Then when you push to your repo, you have to use "git push -f origin master" -- the "f" is needed to force the overwrite of your previous commit history.

Kha:
> I assume the workflow now is to close this pull request, change the history and submit a new one.

There shouldn't be any need for that, pull requests work quite well with force pushes to their respective branch. It doesn't, however, notify participants of new pushes, so you should leave a comment after that.

In my opinion, the best workflow consists of addressing review comments via additional commits (so they can be reviewed iteratively), then squashing them when the review is completed.

johoelzl:
@Kha thanks for the hint, I will try this workflow.

johoelzl:
@avigad So, I renamed most theorems. I didn't mention compose, i.e. `increasing_u_l: increasing (u ∘ l)` or even worse: `u_l_u_eq_u : u ∘ l ∘ u = u`.

@rlewis1988 I also changed the implicit parameters for the fun type class instances to explicit parameters.

If everything is fine I will incorporate the changes into the first two changes, and then force push.

avigad:
Looks good to me. Leo manages the pull requests to the library; my vote is :+1: 

The naming scheme sounds good. The only point is to make it easy to find the names with tab completion; we can always adjust later if necessary.

PR #951, opened 2016-01-13
Fix for cl-lib, cl.el issues

syohex:
- Use cl-lib functions/macros instead of cl where cl-lib.el is loaded
- Wrap cl.el with `eval-when-compile` for suppressing byte compile warning


soonhokong:
@syohex, thanks for the contributions! This PR is merged by f749461.

PR #954, opened 2016-01-14
Don't use deprecated function, string-to-int

syohex:
string-to-int was deprecated since Emacs 22.
string-to-number should be used instead of string-to-int.


soonhokong:
Merged by 885c62648f5d7d175873b2a206a2f0c689d343d7

PR #958, opened 2016-01-14
Use s-join instead of string-join

syohex:
Because string-join was introduced at Emacs 24.4 and subr-x.el must be
loaded for using it.


soonhokong:
Merged by f2eef7aa1b7c9b7ed8a3147acb62567263dc789a

PR #959, opened 2016-01-15
reorganize analysis library

avigad:
Now everything is in an "analysis" namespace. The reals are an instance of a Banach space, which is an instance of a metric space, and theorems are inherited appropriately. I compared compile times with the previous version, and they seem to be comparable.

Some minor things:

When we open `real`, then `real.inf` and `real.sup` overload the algebra `inf` and `sup.` We may want to put the latter in a `lattice` namespace. 

At one point, I had to change a `rewrite` to a `krewrite`, I think because now negation on the reals can be found in two ways -- it can be the Banach space minus, or the field minus. It only happened once, though.

### DIFF
--- a/library/algebra/module.lean
+++ b/library/algebra/module.lean
@@ -18,7 +18,7 @@ structure left_module [class] (R M : Type) [ringR : ring R]
   extends has_scalar R M, add_comm_group M :=
 (smul_left_distrib : ∀ (r : R) (x y : M), smul r (add x y) = (add (smul r x) (smul r y)))
 (smul_right_distrib : ∀ (r s : R) (x : M), smul (ring.add r s) x = (add (smul r x) (smul s x)))
-(smul_mul : ∀ r s x, smul (mul r s) x = smul r (smul s x))
+(mul_smul : ∀ r s x, smul (mul r s) x = smul r (smul s x))
 (one_smul : ∀ x, smul one x = x)
 
 section left_module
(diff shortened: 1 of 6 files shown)


johoelzl:
Regarding `inf` and `sup` that's something we should talk about @ CPP or in CMU: Lattices are very basic and I would suggest to move them to the algebra name space, and instantiate lattice for all total_orders.

Another idea: introducing complete_lattices in look-step with sets.

avigad:
Good point. Linear orders are already instances of lattice, with sup = max and inf = min. We should define conditional lattices and instantiate the reals, as you did in Isabelle, and use notation Sup and Inf for sup and inf over sets.

We already have complete lattices. We haven't yet instantiated sets to that, but we should do that, too.

johoelzl:
Oh yes, conditionally complete lattices are nice too!

I already added the instance to `complete_lattice.lean`. With look-step definition I meant, that we would need to define parts of `set`, then introduce `complete_lattice` and then continue with `set`. My proposal c.f. `complete_lattice` for `set` is the following:

The order `⊆` is as early as possible a complete lattice, i.e. it is an abbreviation of `has_le` for `set`. And `b`/`s`/`Union`, are special instances of `Sup` and `SUP`. Where `(SUP x : S, f x) = Sup (f ' S)`. and `SUP x, f x = SUP x : univ, f x`. (dual for `b`/`s`/`Inter` and `Inf` / `INF`).

Also `bounded_exists` and `bounded_forall` are very similar to `SUP`, `INF` on `Prop`.

The problem with this approach is, that `complete_lattice` is defined in terms of `set`. So we need to define `set` and `∈` then `complete_lattice` only after this the order properties of `set`. Also `set` is only classically a `complete_lattice`, this might be also a problem...

Or is it possible to extend the type class instances of a Type later on, i.e. introduced for set 
the following instances:
   `weak_order (set A)`
and later on a instance of
  `complete_lattice (set A)`
I assume this problematic (i.e. theorems about `set1 <= set2` would carry the wrong type class instance).

avigad:
Yes, it is certainly possible to extend instantiations incrementally. We do that often. As you say, an issue is that the operations that class resolutions "finds" look slightly different, though they are definitionally equal.

We all need to talk about this more. An advantage to the current definitions for sets is that they unfold easily to the logical specifications. BTW, I think that right now "set" is the only part of the core library where  classical logic is used substantially. We should think about how important it is to separate the constructive and nonconstructive parts.

We still have to work out some basic methodology. We have been thinking of Lean as either "Isabelle + dependent types" or "Coq + automation", but these are not exactly the same. In the Coq approach, you rely on computational reduction a lot. In the Isabelle approach, you don't unfold definitions automatically, and rely on explicit theorems to translate. As the automation starts coming to life, we have to think about how we balance these two approaches.

PR #961, opened 2016-01-16
fix(src/util/file_lock): ignore failure to create locks on read-only file systems

gebner:
Right now, `file_lock` ignores the `EACCES` error when creating lock files, assuming that the file to be locked is part of a system-wide installation.  On NixOS however, the file system containing system packages is even mounted read-only, so `open` returns `EROFS`.  This results for example in the following error:

```
/home/gebner/lean/library/examples/ex.lean:4:0: error: failed to lock file '/nix/store/fxc0cvnv0j6kc1yc2ygngz7c805n10j5-lean-20160116/lib/lean/library/init/default.olean'
```

This PR ignores `EROFS` in addition to `EACCES`.


leodemoura:
Thanks for making the pull request.

PR #964, opened 2016-01-20
choice

skbaek:
Added `is_trunc_equiv` to `types/equiv.hlean`, and lemmata 3.8.2 & 3.8.5 of the book to a new file `choice.hlean`

### DIFF
--- a/hott/choice.hlean
+++ b/hott/choice.hlean
@@ -28,13 +28,13 @@ section
   (to_inv !sigma_pi_equiv_pi_sigma) (H X A P HX HA HP HI))
 
   -- AC_cart can be derived from AC' by setting P := λ _ _ , unit.
-  definition AC'_to_AC_cart : AC'.{u} -> AC_cart.{u} :=
+  definition AC_cart_of_AC' : AC'.{u} -> AC_cart.{u} :=
   λ H X A HX HA HI, trunc_functor _ (λ H0 x, (H0 x).1) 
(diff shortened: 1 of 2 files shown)


fpvandoorn:
Thank you very much for polishing this into a pull request! The help is very much appreciated.

> not_is_hprop_bool_eq_bool and is_hset_bool are general facts about bool, so it might be better to move them to a different file for easier access.

Yes, we have a file `types.bool` where that would be perfect. (the reason why `is_hset_bool` was not yet in the library is because Lean can solve it by type class inference, which means that you can proof it by `_`)

> Not sure what to name X_must_be_hset. That name is what 3.8.5 means, but there may be a better alternative that follows Lean's naming convention.

This is a hard one: I don't think our naming scheme works well for naming counterexamples. I also have a hard time to find a good name for this.

Is it OK if I make some changes to your files before we integrate it to the library? I will then also move the facts about `bool` to `types.bool`.

skbaek:
> Is it OK if I make some changes to your files before we integrate it to the library? I will then also move the facts about bool to types.bool.

Yes, of course! I'm not sure how the process works (I'm fairly new to GitHub), but if I need to make additional changes on my side please let me know.

fpvandoorn:
That's not needed. But do look at the changes I made, and ask any questions if you're unsure why I made them.
This PR is superseded by #966.

PR #992, opened 2016-02-09
feat(init/wf): port from standard library to HoTT library

fpvandoorn:
Porting `wf.lean` to the HoTT library for Gershom.

After this commit we need some more advanced theorems in init/wf, notably function extenstionality.
For this reason I had to refactor the init folder a little bit.
To keep the init folders in both libraries similar, I did the same refactorization in the standard library, even though that was not required for the standard library.


fpvandoorn:
Never mind, I just noticed a simpler solution.

fpvandoorn:
Ok, not it can be merged :-)

PR #1000, opened 2016-02-13
feat(library/defeq_simplifier): new simplifier that uses only definit…

dselsam:
…ional equalities


robertylewis:
A milestone: issue number 1000!

PR #1002, opened 2016-02-14
feat (library/theories/topology/basic) : add separation theorems

JacobGross:
add T0, T1, T2 separation theorems and add closed singleton theorem for T1 spaces

### DIFF
--- a/library/theories/topology/basic.lean
+++ b/library/theories/topology/basic.lean
@@ -152,8 +152,15 @@ structure T0_space [class] (X : Type) extends topology X :=
 namespace topology
   variables {X : Type} [T0_space X]
 
-  theorem T0 {x y : X} (H : x ≠ y) : ∃ U, Open U ∧ ¬(x ∈ U ↔ y ∈ U) :=
-  T0_space.T0 H
+  theorem separation_T0 {x y : X} : x ≠ y ↔ ∃ U, Open U ∧ ¬(x ∈ U ↔ y ∈ U) :=
+  iff.intro
+    (T0_space.T0)
+    (assume H, obtain U [OpU xyU], from H,
+     suppose x = y,
+     have x ∈ U ↔ y ∈ U, from iff.intro 
+       (assume xU, this ▸ xU) 
+       (assume yU, this⁻¹ ▸ yU),
+     absurd this xyU)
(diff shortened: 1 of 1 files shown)


robertylewis:
@avigad @leodemoura Is there an established convention for implicit vs explicit variables in theorems like this?

``` lean
theorem separation_T0 {x y : X} : x ≠ y ↔ ∃ U, Open U ∧ ¬(x ∈ U ↔ y ∈ U) 
```

I don't think we're consistent about this in the library. I generally prefer implicit like Jacob has done, since I'm almost always using iff theorems in the form `iff.mp separation_T0 H`. But I think this goes against our stated convention, and there are theorems in algebra that don't follow this. If we agree on one style I'll change them as I see them.

(In general though, I prefer to have two one-directional implications around -- they're easier to search for.)

avigad:
The convention I have been following is to make these variables explicit (and also to state the one-directional versions). I'm also willing to adjust this after the pull request is merged.

robertylewis:
For one instance of the opposite convention being used, see https://github.com/leanprover/lean/blob/master/library/logic/identities.lean

I'm not sure if changing those would break a lot. But I can at least add the one-directional versions.

spl:
I'll add to the record this example – one of several – of contrasting implicit styles:
- [`init.nat.le_trans {n m k : ℕ} (H1 : n ≤ m) : m ≤ k → n ≤ k`](https://github.com/leanprover/lean/blob/master/library/init/nat.lean#L99)
- [`algebra.order.le_trans : ∀a b c, le a b → le b c → le a c`](https://github.com/leanprover/lean/blob/master/library/algebra/order.lean#L17)

fpvandoorn:
I would also make `x` and `y` explicit here.

> - [`algebra.order.le_trans : ∀a b c, le a b → le b c → le a c`](https://github.com/leanprover/lean/blob/master/library/algebra/order.lean#L17)

While this is indeed the "wrong" style, you won't use this projection in practice. The one you use in practice has the right style:
[`le.trans : ∀{a b c : A}, a ≤ b → b ≤ c → a ≤ c`](https://github.com/leanprover/lean/blob/master/library/algebra/order.lean#L28)

PR #1004, opened 2016-02-16
renamings and additions in HoTT

fpvandoorn:
- Additions to the library of pointed types.
- Changes in `init.path` to use the `induction` tactic instead of `eq.rec`, also added some missing theorems.
- Major renamings as sent to lean-user:

is_hprop -> is_prop
is_hset -> is_set
(this included some C++ files)

hprop -> Prop
hset -> Set
Pointed -> pType

Currently if `foo` is a type, we currently use `Foo` for its corresponding pointed type.
New naming convention: use `pfoo`. So for example `Susp` will be renamed to `psusp`.
Similarly for pointed maps. E.g. `Susp_functor` will become `psusp_functor`


fpvandoorn:
I fixed (the only) one broken test. It was funny why it broke: it broke because `+` was _not_ overloaded anymore.

fpvandoorn:
Adding two more commits.

PR #1006, opened 2016-02-17
various additions

avigad:
- renamed complement -> compl in sets and finsets
- did various housecleaning
- tinkered with typeclass instances in the reals to help the unifier a bit
- add preimages of sets
- improved complete_lattices
- improved filters
- added the order topology, from Jacob Gross
- added material on filters, from Jacob Gross


avigad:
I added to this pull request the definition of a real inner product space, and proofs that it instantiates a normed space (via the Cauchy Schwartz inequality). I also used this to instantiate the complex numbers as a real normed space, which will allow us to define the exponential function for the reals and complex numbers (and in fact any real Banach algebra) simultaneously.

Some notes:
- I used `⟨v, w⟩` for the inner product. This overwrites the notation for dependent pairs, but I don't think this will cause a problem.
- A few times I came across the unification issues that Leo has been worrying about, e.g. with diamonds in the inheritance graph. For example, I proved some facts about squares and square roots, but my proofs broke when I moved these facts elsewhere in the library, and I had to change some `rewrite`s to `krewrite`s and provide more arguments explicitly. The workarounds are not terrible, but Leo is right that we need to solve this problem (so I am glad that he and Daniel are working on this).
- The library has modules over an arbitrary ring; vector spaces over the reals are a special case. The elaborator and rewrite sometimes have a hard time figuring out the relevant ring, so I created instances of some facts for the reals: https://github.com/avigad/lean/blob/master/library/theories/analysis/normed_space.lean#L25-L50.
  It is not a big deal to use these, but we should keep an eye on this: if the elaborator and unifier improve, they may not be necessary later on.

PR #1008, opened 2016-02-23
Fix markdown links

Kha:
I stumbled over a broken link in library.md, so I wrote a quick Python script for checking all of them: https://gist.github.com/Kha/d8b4b0006db50f2e90c7

I can add the script as a test case if you think that would be helpful. When the documentation is more complete, a reverse test - find files that aren't mentioned in .md files - would probably also be interesting.

### DIFF
--- a/hott/algebra/category/limits/limits.md
+++ b/hott/algebra/category/limits/limits.md
@@ -3,6 +3,5 @@ algebra.category.limits
 
 * [limits](limits.hlean) : Limits in a category, defined as terminal object in the cone category
 * [colimits](colimits.hlean) : Colimits in a category, defined as the limit of the opposite functor
-* [complete](complete.hlean) : Categories which are (co)complete or constructions which preserve (co)completeness
 * [functor_preserve](functor_preserve.hlean) : Functors which preserve limits and colimits
 * [adjoint](adjoint.hlean) : the (co)limit functor is adjoint to the diagonal map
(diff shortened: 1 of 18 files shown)


hott/algebra/category/limits/limits.md
```diff
@@ -3,6 +3,5 @@ algebra.category.limits
 
 * [limits](limits.hlean) : Limits in a category, defined as terminal object in the cone category
 * [colimits](colimits.hlean) : Colimits in a category, defined as the limit of the opposite functor
-* [complete](complete.hlean) : Categories which are (co)complete or constructions which preserve (co)completeness
```
Kha, line 6: I couldn't find any revision of this file in the git logs.

leodemoura:
@Kha This is great!
Do you mind including the Python script in the pull request?
We have a `script` sub-directory for this kind of tool.

avigad:
I agree -- this is very nice.

fpvandoorn:
This is very nice! I knew that I would occasionally move files and forget to change the markdown file, but I hoped I had made fewer mistakes doing this. It would be very nice to have this in the test suite.

> I couldn't find any revision of this file in the git logs.

I think I created this file and then renamed it before I even committed it. The contents are now in [set.hlean](https://github.com/Kha/lean/blob/master/hott/algebra/category/limits/set.hlean) in that folder (which should indeed be added to that folder).

> When the documentation is more complete, a reverse test - find files that aren't mentioned in .md files - would probably also be interesting.

Yes, that would also be really useful. I often forget to add newly created files to the markdown files, so I think there are also lots of files missing.

Kha:
Ok, added the script to `script/`. It turns out I had way too much fun writing Python code again after quite some time, so the script now optionally also checks http links (can be slow) and the reverse test (you probably only want to use this from `library/` and `hott/`), and returns a non-zero status code in case of any warning.

PR #1009, opened 2016-02-23
feat(script/check_md_links.py): make python2/3 compatible

soonhokong:
@Kha, I needed to make some change to run the code with python2. I've checked that it works with both of Python-2.7.11 and Python-3.5.1. Could you check that it looks OK? If so, I'll merge this commit to the blessed repo.

Related link: http://python3porting.com/noconv.html#import-errors

### DIFF
--- a/script/check_md_links.py
+++ b/script/check_md_links.py
@@ -4,8 +4,12 @@ import argparse
 import collections
 import os
 import sys
-import urllib.error
-import urllib.request
+try:
+    from urllib.request import urlopen
+    from urllib.parse import urlparse
+except ImportError:
+    from urlparse import urlparse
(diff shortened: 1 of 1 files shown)


Kha:
Ah, I didn't check for that. You'll probably want to add a `from __future__ import print_function`, otherwise the output will be mangled.

soonhokong:
@kha, I added it. I also put copyright form.

fpvandoorn:
>  I also put copyright form.

You made a (copy-paste) error with the name in the copyright form.

soonhokong:
@fpvandoorn, oh I made a mistake :-( 
I just fixed it. Thanks for the notice!

PR #1012, opened 2016-02-29
feat(hott) order categories and type functors

javra:


fpvandoorn:
In this pull request I get the same error. Did you push after rebasing/fixing?

javra:
Thanks ^^ I wonder why the travis check went through anyway... doesn't it built the library anymore?

PR #1015, opened 2016-03-03
additions to HoTT

fpvandoorn:
Add theorems about pointed equivalences, nat/sphere_index/trunc_index, and a little more homotopy theory.

This also removes the `[multiple_instances]` attribute.

I've also formalized the long exact sequence of homotopy groups in Lean (section 8.4 of the HoTT book), but it needs some cleanup before I add it to the HoTT library. Currently it can be found [here](https://github.com/fpvandoorn/Spectral/blob/personal/homotopy/LES_of_homotopy_groups.hlean).


fpvandoorn:
Added one more commit, which replaces `assert` by `have` in my previous commits, and also merges the namespace `equiv.ops` into `equiv`.

PR #1020, opened 2016-03-07
Hopf fibrations

UlrikBuchholtz:
The main purpose of this commit is to add the complex and quaternionic Hopf fibrations. The latter is done via the modified Cayley-Dickson construction using imaginaroids.

A further highlight is the Main Lemma from Licata and Finster's paper on Eilenberg-Mac Lane spaces, which gives a delooping of a K(G,1); we'll need this later for our construction of EM spaces and spectra.

Also included as biproducts: improvements to join; functoriality of pushout and susp (and conditions for when we get an equivalence), and the flattening lemma for susp.

Finally, the connectedness namespace is renamed to is_conn.

Cc @fpvandoorn and @javra

### DIFF
--- a/hott/book.md
+++ b/hott/book.md
@@ -139,14 +139,14 @@ Chapter 7: Homotopy n-types
 Chapter 8: Homotopy theory
 ---------
 
-Unless otherwise noted, the files are in the folder [homotopy](homotopy/homotopy.md)
+Every file is in the folder [homotopy](homotopy/homotopy.md)
 
-- 8.1 (π_1(S^1)): [homotopy.circle](homotopy/circle.hlean) (only the encode-decode proof)
-- 8.2 (Connectedness of suspensions): [homotopy.connectedness](homotopy/connectedness.hlean) (different proof)
-- 8.3 (πk≤n of an n-connected space and π_{k<n}(S^n)): [homotopy.homotopy_group](homotopy/homotopy_group.hlean)
+- 8.1 (π_1(S^1)): [circle](homotopy/circle.hlean) (only the encode-decode proof)
+- 8.2 (Connectedness of suspensions): [susp](homotopy/susp.hlean) (Theorem 8.2.1, different proof)
+- 8.3 (πk≤n of an n-connected space and π_{k<n}(S^n)): [homotopy_group](homotopy/homotopy_group.hlean)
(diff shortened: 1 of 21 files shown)


UlrikBuchholtz:
This probably will conflict with https://github.com/leanprover/lean/pull/1019 – I can update this request when the other one is closed.

fpvandoorn:
Very nice. That is a lot of useful results! Thanks for offering to resolve merge conflicts. Some questions/comments:
- Could you update `book.md` for the contents of section 8.2. I agree with the move.
- Can you also update `homotopy/homotopy.md`?
- Very interesting that you need the computation rule of the elimination principle of the 1-point circle for the constructors of the 2-point circle. Does that mean it is actually useful for you that those two circles are the same type?
- We missed the same results about moving equivalences in homotopies. I added these recently to [init.equiv (L218-228 and 381-398)](https://github.com/leanprover/lean/blob/master/hott/init/equiv.hlean#L218-L228). Can you merge that with what you've done?

UlrikBuchholtz:
> Could you update book.md for the contents of section 8.2. I agree with the move.

Done.

> Can you also update homotopy/homotopy.md?

Done. See also attached [dependency graph](https://github.com/leanprover/lean/files/161618/treddepshtpy.dot.pdf).)

> Very interesting that you need the computation rule of the elimination principle of the 1-point circle for the constructors of the 2-point circle. Does that mean it is actually useful for you that those two circles are the same type?

Yes, for the imaginaroid construction on the zero-sphere, we need the interaction between complex conjugation on the circle and circle multiplication, where conjugation is defined using the suspension structure of the circle. But it's easier to prove things in the base-loop description, so a comparison is needed. See [here](https://github.com/UlrikBuchholtz/lean/blob/e87850e11b1ff8ba130525a3d8b8aa6cae7c27f1/hott/homotopy/quaternionic_hopf.hlean#L45-L57).

> We missed the same results about moving equivalences in homotopies. I added these recently to init.equiv (L218-228 and 381-398). Can you merge that with what you've done?

Yes, will do when making the updated pull request once your request is pulled. (@leodemoura I'll leave this open for now, but go ahead and close when you have merged #1019 )

leodemoura:
#1019 has been merged

UlrikBuchholtz:
Closing because we now have #1021

PR #1022, opened 2016-03-09
a general theory of limits, based on filters

avigad:
... following Hölzl, Immler, and Huffman, "Type classes and filters for mathematical analysis in Isabelle/HOL". This pull request includes work by Jacob Gross.

Some of the material in this file will eventually go into the filters file in data.sets, but I will avoid making any changes to core parts of the library that Leo is using while refactoring. (Actually, I will probably later move that file to `theories/topology`.) I will also later rename `linear_strong_order_pair` to `linear_order`.

@rlewis1988 @JacobGross I think I have settled on nice notation for limits, described in the header comments to library/theories/topology/filterlim.lean.

There is still a lot to do:
- define the neighborhood filter and "within" modifier, which will give limits in topological spaces
- define pointwise continuity, and pointwise continuity within a set.
- define the metric space instantiations, and generalize the theorems that are already in the analysis library.
  At least things are moving along now.

Regarding the second, @johoelzl suggested that limits continuity with respect to a set can be handled by moving a subspace topology, but I am not convinced it will make our lives easier.

### DIFF
--- a/library/theories/topology/filterlim.lean
+++ b/library/theories/topology/filterlim.lean
@@ -1,0 +1,503 @@
+/-
+Copyright (c) 2016 Jacob Gross. All rights reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: Jacob Gross, Jeremy Avigad
+
+Limits, following Hölzl, Immler, and Huffman, "Type classes and filters for mathematical
+analysis in Isabelle/HOL".
+
+We are experimenting with notation like this:
+
+  f ⟶ y [at x]
+
+... which is actually a composition of two notations: the first describes the function and
+the target filter, and the second desribes the source filter.
+
+Variations on the first can include:
+
+  f ⟶ ∞
+  f ⟶ -∞
+  f ⟶+ y          convergence from the right
+  f ⟶- y          convergence from the left
+  f ⟶ y within s  convergences within a set
+
+Variations on the second can include:
+
+  [at ∞]
+  [at -∞]
+  [at x]
+  [at+ x]
+  [at- x]
+  [at x within s]
+
+When written consecutively, we get filterlim f F₂ F₁, where the first notation gives f and F₂,
+and the second gives F₁.
+
+Note that lambdas are needed for functions given by expressions:
+
+  (λ x, x^2) ⟶ 4 [at 2]
+  (λ x, x^2) ⟶ ∞ [at -∞]
+
+If making `[at` a token is problematic, we can add a tick mark or something.
+-/
+import data.set data.nat algebra.order algebra.interval
+open set function set.filter interval
+
+
+-- TODO: move to data/set/basic
+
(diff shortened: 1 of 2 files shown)


robertylewis:
The notation looks very nice!

johoelzl:
Yeah, the notation is really nice!

We gave `filterlim` and `filtermap` actually the wrong names. I'm not sure about Lean's naming convention, but in Isabelle/HOL we try to stick to the convention `<const>_<type>`. So nowadays I would name them `lim_filter` and `map_filter`. The theory name could be `filter_limits`? Also, for the future it would be surely nice to set up a `monad` type class, to have a general `map` constant. Maybe also products, etc...?

## Continuity on sets

I'm unsure about continuity on a set. For the development of topology itself, it is enough to just think about continuous functions in the entire set. Also e.g. compact spaces are associated with the type and I wouldn't care about compact sets. But working with real analysis and upwards one practically needs to have continuity and compactness on subsets.

I think we need to keep in mind that this is also possible: develop your theory in a **type-based** style and **then** use `subtype`-topologies (and other `subtype` typeclass instances) to derive more practical **set-based** rules which are easier to use. There is quite some boilerplate in Isabelle's topology (and also measure theory) as we often forced development in a set-based style (e.g. the functions ranges are still in the right set etc).

avigad:
In the pull request, I used `mapfilter` instead of `filtermap`. I decided not to add an underscore, because with our naming conventions it is less confusing when identifiers are not split into two words. (`map_filter_...` would seem to say something about `map (filter x)`).

I kept `filterlim` instead of `limfilter`, because it isn't really a limit of filters. It isn't really a filter of limits either, but, rather, a certain kind of limit, which may as well be called a "filter limit". 

I considered changing `filterlim f F1 F2` to `limit f F1 F2` or `converges f F1 F2` but both of these have downsides as well and don't suggest the level of generality. So I kept `filterlim` (again, to avoid the underscore). I can be talked out of it, though. I think it is a good idea to change the name of the theory to `filter_limits`, since that is clearer.

I still don't know whether it will be easier to use subtype topologies or just to use subsets everywhere. I will experiment.

johoelzl:
Ah sorry, I overlooked that you already changed `mapfilter`. A better name for `filterlim` would be surely good: `converges` sounds quite good to me.

avigad:
@johoelzl I avoided `converges f F1 F2` because `converges f F`  sounds like it should mean `∃ y, (f → y) F`. Maybe it would be better to use `converges_to f F1 F2` or `approaches f F1 F2`.

What do you think of `approaches`? (Also @rlewis1988?)

robertylewis:
I agree, I don't like `converges f F1 F2`. `approaches` sounds reasonable, as does `converges_to` -- I think we've been using `converges_to` for the special-case definitions in the library right now.

johoelzl:
I also like `approaches`. It sounds more general and less _topological_ compared to `converges`.

Historically the name `filterlim` was chosen to be similar to `LIMSEQ` for limits of sequences and `LIM` for neighbourhood limits. So for me it _is_ somehow the same limit concept just for different instances.

PR #1025, opened 2016-03-20
feat(library): introduce exists.intro2

spl:


avigad:
@spl Thanks for this! An exists.intro2 is a good idea. But I am getting compile errors with the current repository.

Leo is doing a major refactoring of the elaborator. The idea is to make it less general but much more stable and predictable. He is adapting parts of the core library in a separate branch, so now is not the best time to introduce changes there -- let's wait for the new elaborator to settle.

leodemoura:
@spl This pull request still doesn't work on top of the current standard library.
I'm closing the pull-request. Please, feel free to reopen it if you manage to fix it.

Thanks,
Leo

PR #1038, opened 2016-04-06
Fix emacs info commands in project subdirs

Kha:
The `lean-show-*` commands did not respect the current project for the value of `--dir`, which resulted in empty output.

### DIFF
--- a/src/emacs/lean-company.el
+++ b/src/emacs/lean-company.el
@@ -84,8 +84,6 @@
 (defun company-lean--import-candidates (prefix)
   (let* ((cur-dir (f-dirname (buffer-file-name)))
          (parent-dir (f-parent cur-dir))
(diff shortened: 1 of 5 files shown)


Kha:
It seems I misunderstood the purpose of `--dir`. Now setting the working directory for `start-process` instead.

@soonhokong I changed `lean-project-find-root` to return the path to the directory instead of the `.project` file. If anything is amiss, please tell me - these are my very first lines of Lisp code :) .

soonhokong:
@Kha, thanks! I'm traveling now but will check it this evening.

Kha:
Today I encountered a related issue for the `lean --server` process (auto-completion failed in a project subdir), so I have extended the fix to `lean-server-create-process`. It's not a perfect solution - you'll have to restart the process when switching projects - but I think it's a good compromise if we don't want to extend the server protocol.

PR #1040, opened 2016-04-12
fix(frontends/lean/elaborator): incorrect assertion

dselsam:


leodemoura:
merged in the `lean3` branch

PR #1042, opened 2016-04-16
Structure check

dselsam:
Note that this pull request simply disallows structures whose result universe is not definitely zero but might be zero, when the environment is impredicative. The check for inductive types that satisfy this property is conservative and rejects if there is a single intro-rule and it takes a (non-parameter) argument. Since the unique intro rule of structures must take at least one argument, the policy in this pull request is not any more conservative than the one for general inductive types.


leodemoura:
merged in the `lean3` branch

PR #1043, opened 2016-04-17
Option to disable printing of proofs that appear as arguments to functions

dselsam:
When I use the "safe" programming style, I end up with huge proof terms that are not implicit arguments that make the proof state very hard to read. This pull request introduces a new option `pp.proofs` which defaults to `true`, but when disabled will, for arguments to functions that are proofs, print the theorem statement instead of the proof term itself. Note that since the macro/pp interface is not completed yet, I took a shortcut and re-used the `inaccessible` annotation for the theorem statement. See `pp_no_proofs.lean.expected.out` below for an example.

In the process of adding this feature, I noticed a bug that prevented coercions that take more than one argument from being hidden (37281d5) when `pp.coercions = false`.

Note: I chose to only alter the rendering of proof terms that appear as function arguments, whereas I could have made both the implementation and the semantics simpler by inferring the type of each expression no matter where it occurs. I chose the first approach for two reasons: (a) performance, since we do not need to infer the type of a (huge) proof term in order to ignore it, and (b) the type one sees by type-inference may not be pretty in general. If others would prefer the second approach, I would be happy to switch.


fpvandoorn:
> In the process of adding this feature, I noticed a bug that prevented coercions that take more than one argument from being hidden (37281d5) when pp.coercions = false.

Very nice! This is indeed a bug I've noticed for a long time, but which seemed to small to report.

leodemoura:
merged in the `lean3` branch

PR #1047, opened 2016-04-25
fix(frontends/lean/inductive_cmd): remove universe placeholder from p…

dselsam:
…arameters

Note that the following less helpful error is triggered for the test file on master:

```
error: invalid reference to undefined global universe level '14.u'
```

The issue is that even though the global universe placeholder starts out only as the resultant level, it may appear elsewhere after unification.


leodemoura:
merged in the `lean3` branch

PR #1048, opened 2016-04-25
Major additions to the HoTT library

fpvandoorn:
This pull requests adds:
- truncated 2-quotients
- groupoid quotients
- Eilenberg MacLane spaces `K(G,n)` (no proof of uniqueness of such spaces yet)
- Chain complexes
- The long exact sequence of homotopy groups
- The Freudenthal Suspension Theorem
- Whitehead's principle
- The computation of homotopy groups of almost all spheres which are known in HoTT. (Only Guillaume's proof of `π₄(S²)` is missing (but that has also not been formalized anywhere else yet).

Also various changes and fixes to the HoTT library. 
@avigad: I added the induction principle for `image`, so if you have `v : image f b` you can use the tactic `induction v with a p` to get `a : A` and `p : f(a) = b` (as long as you're proving a mere proposition).

### DIFF
--- a/hott/algebra/algebra.md
+++ b/hott/algebra/algebra.md
@@ -16,9 +16,10 @@ The following files are [ported](../port.md) from the standard library. If anyth
 * [ordered_field](ordered_field.hlean)
 * [bundled](bundled.hlean) : bundled versions of the algebraic structures
 
-Files which are HoTT specific:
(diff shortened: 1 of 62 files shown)


fpvandoorn:
Added a commit which updates `book.md`

PR #1050, opened 2016-05-06
new group theory library

avigad:
This pull request includes an update to the Emacs mode documentation, but the main component is a new library for group theory. The latter includes subgroups, homorphisms, kernels, normalizers, centralizers, and quotients. This is joint work with Andrew Zipperer.

Here are the main differences from Haitao's theories:
- it is based on `set`, rather than `finset`
- it incorporates "local" versions of all notions and constructions, e.g `hom_on f G` to say that f is a homomorphism on G, `ker_in f G` to denote the kernel of f in G, `normal_in H G` to say that H is normal in G, etc.

Otherwise, we are still playing "catch up" to Haitao.

We overload notation, using `a * H` and `H * a` for left and right cosets, and `G / H` for quotient groups. It works well, and looks really nice.

We also use class inference quite aggressively, to infer that things are subgroups, homomorphisms, normal, etc. All this works surprisingly well also. One day, special purpose automation may provide some improvements. For example, if `hom_on f G` and `H ⊆ G`, then `hom_on f H`, but it seems hard to deal with that well using class inference.

At one point, I hit a wall, where a theorem I was working on took 7 seconds to elaborate. The probablem appeared when I added a line with trivial inference.  

  https://github.com/avigad/lean/blob/master/library/theories/group_theory/quotient.lean#L459-L522

I think that problem was that with an "obtain" somehow the elaborator got lost in a crazy search. I worked around it by writing the proof in tactic mode.

Otherwise, there were no serious performance problems. Of course, things can probably be made more efficient, but it doesn't make sense to invest a lot of time on optimization until we move to Lean 3. 

During the refactoring, rather than make additions / changes to the core library, I continue to add things to `library/theories/move.lean`.


avigad:
@soonhokong I don't understand what went wrong with the Travis check. If there is something I need to fix, please let me know.

soonhokong:
It seems that nothing is really wrong. For `clang` + DEBUG Build combinations, we had timeouts.

PR #1055, opened 2016-05-14
Implement a compiler to C++, and link to the existing runtime

jroesch:
This set of commits implements a compiler from Lean's expression type to C++. This is still very much a prototype and needs more testing and features.

I am opening this pull request now, even though it is not done because I was able to write and successfully link a few example programs with IO, and it would be great to merge it now. I am working on finishing a POPL submission, and will probably be less free in the coming weeks. 

I would also appreciate getting some code review, and general feedback since I'm pretty rust-y at C++. 

I have a couple more things I want to add before its merged, but would like to start the conversation in a structured fashion, and not just over email. 

In terms of specifics of this patch, @leodemoura it would be great to get some feedback on the runtime, since I believe I found some bugs. I "fixed" a few of them, but have questions about a couple more, and will comment on the appropriate code below.  

My future wish list for this is:
- user provided optimizations in the form of proofs
- a rich typed pointer API for interacting with foreign code
- a structured mechanism for reasoning about partiality
- lots of optimizations both on `expr` and the `simple_expr` level
- runtimes and code generators for different output languages (specifically Rust, since it would
  allow verified code to interact with low-level systems code with stronger properties than C++)
- possibly emitting LLVM IR

A couple issues I think are important to resolve before merging:
- [ ] integrate running tests into the testing infrastructure
- [ ] automatically select and invoke the compiler programmatically
- [ ] clean up the emitted code
- [ ] add a bunch more tests
- [ ] add more configuration

I'll try to address the above issues tomorrow if time permits.

Also shoutout to @lukenels who I bounced a bunch of ideas at, and who helped with some of the IO code.


src/runtime/cpp/lean_runtime.cpp
```diff
@@ -202,11 +216,17 @@ void obj_cell::dealloc() {
             case obj_kind::Closure:
                 DEC_FIELDS(it, todo);
                 it->~obj_cell();
-                g_obj_pool->recycle(it, sz+1);
+                // trying to recycle closures and raw ptrs causes crashes
```
jroesch, line 219: Trying to recycle either a Closure or my new RawPtr type causes a bad access to be trigger when attempting to clean up the free list.

src/runtime/cpp/lean_runtime.cpp
```diff
@@ -94,6 +100,7 @@ obj_cell::obj_cell(obj_cell const & src, obj const & a1):
     copy_fields(src);
     void ** f = field_addr();
     new (f + src.m_size) obj(a1);
+    *fn_ptr_addr() = src.fn_ptr();
```
jroesch, line 103: I believe this was a bug where copying closures would lose the fn_ptr, causing a segfault when using `apply(..)`.

src/runtime/cpp/lean_runtime.cpp
```diff
@@ -31,13 +31,13 @@ class obj_pool {
     }
 
     void * allocate(unsigned n) {
-        if (n < LEAN_NUM_OBJ_FREE_LISTS && m_free_list[n] != nullptr) {
-            void * r = m_free_list[n];
-            m_free_list[n] = *(reinterpret_cast<void **>(r));
-            return r;
-        } else {
+        // if (n < LEAN_NUM_OBJ_FREE_LISTS && m_free_list[n] != nullptr) {
```
jroesch, line 34: I need to add this code back, commented out from a debugging session.

src/backends/simple_expr.h
```diff
@@ -0,0 +1,168 @@
...
+        SVar,
+        Let,
+        Call,
+        Switch,
+        Error,
+        Project,
+        ClosureAlloc,
+    };
+
+    struct simple_expr {
```
jroesch, line 27: I didn't implement the intrusive smart pointer here out of a desire to move quickly, if you think we should do that let me know.

src/backends/simple_expr.cpp
```diff
@@ -0,0 +1,24 @@
...
+#include "kernel/inductive/inductive.h"
+#include "util/name.h"
+#include "util/name_set.h"
+
+namespace lean {
+    void simple_expr::display(std::ostream& os) const {
+        os << "hello world";
+    }
+
+    std::ostream& operator<<(std::ostream& os, const simple_expr & se) {
```
jroesch, line 20: Displaying the new IR is something left unfinished.

src/backends/c_emitter.h
```diff
@@ -0,0 +1,25 @@
...
+#include <iostream>
+#include "kernel/environment.h"
+#include "backend.h"
+#include "simple_expr.h"
+
+namespace lean  {
+    // static const char * LEAN_OBJ_TYPE = "lean::obj";
+    // static const char * LEAN_ERR = "lean::runtime_error";
+
+    class c_emitter {
```
jroesch, line 18: This class is currently empty, but the goal is to clean up the backend by moving most of the formatting code to the emitter, and then just having the backend use the emitter to write out the actual strings.

Kha:
Nice! I love that it already supports monadic IO.

May I ask why you decided on C++ as the (first) target language? Even with the runtime being shared with lean3's VM, I would have assumed a 'managed' language like OCaml to be a much easier target. I suppose it would be interesting to have both at some point and compare their performance.

src/backends/backend.cpp
```diff
@@ -0,0 +1,616 @@
...
+        std::stringstream ss;
+        ss << "internal_error: should not be trying to compile a theorem (";
+        ss << d.get_name();
+        ss << ")";
+        throw new backend_exception(ss.str().c_str());
+    } else {
+        std::stringstream ss;
+        ss << "unhandled declaration: ";
+        ss << d.get_name();
+        throw new backend_exception(ss.str().c_str());
```
MarisaKirisame, line 124: I think you should throw by value and catch by reference. First of all you are leaking memory, and more importantly your catch look uglier because it deals with two catch doing essentially the same - only one work on backend_exception*, one work on backend_exception. see http://www.informit.com/articles/article.aspx?p=373339

jroesch:
@Kha I met with Leo sometime last month, and it seemed that he was interested in generating C++ code coupled with my interest in being able to model and program against low level system APIs. 

Another goal of mine was to design this like a compiler and not an extraction mechanism. I didn't want to copy Coq's approach since frankly its not very robust, and configurability comes in the form of stringly typed code. I want take a approach similar to GHC and design APIs and optimizations that allow programmers performance and control without resorting to string replacement. 

With that said modulo a little bit of generalization it should be easy to write backends for other languages. Specifically I'm interested in targeting Rust, I believe it will be a better candidate for implementing a safe, high performance runtime.

leodemoura:
@jroesch Thanks for making the pull request.
I'm currently on vacation. I will be back next Saturday. I will integrate your changes as soon as I get back.
BTW, I will also merge your commits with the `lean3` branch.
The efficient interpreter is showing signs of life in this branch.
Some of the preprocessing steps used in the interpreter may also be useful for C++ code extraction.

leodemoura:
> In terms of specifics of this patch, @leodemoura it would be great to get some feedback on the runtime, since I believe I found some bugs. I "fixed" a few of them, but have questions about a couple more, and will comment on the appropriate code below.

No problem. I will take a look when I get back.

jroesch:
@leodemoura sounds great, enjoy your vacation! there are a couple lurking issues that I will try to iron out before you get back, busy working on our POPL submission so may not find time this week, but will let you know.

jroesch:
Was just reading over the lean3 branch, it looks like your infrastructure can replace some of the functionality I have. 

I should be able to use the computational irrelevance analysis, and inliner, if not more. 

My current view is that we should mark terms for erasure, do inlining on the `expr` representation, then lower to the `proc`/`simple_expr` since its closure converted/ANF. 

Finally we can then do lower-level imperative optimizations, and output to C++ code or even LLVM if we are so inclined, I also generalized the interface for code generation so we should be able to have multiple target backends. I think it would be good to have some permutation of C++/Rust/OCaml/Haskell.

jroesch:
I looked around a little more tonight, and it seems that this whole chunk of code should be able to replace the first part of the backend: https://github.com/leanprover/lean/blob/3b47d91b3ed6486aa6d88fc67e17571a9db90ab2/src/library/compiler/preprocess_rec.cpp#L116. 

I have my changes rebased on top of the lean3 branch, but the standard library doesn't build due to all the `sorry`s, and a couple of my tests rely on it. 

Branch here: https://github.com/jroesch/lean/tree/lean3-cg.

leodemoura:
> Branch here: https://github.com/jroesch/lean/tree/lean3-cg.

@jroesch This is great. We can discuss the details and remaining issues when we meet in person.

jroesch:
Closing this PR in favor of my new one, which is also a WIP. https://github.com/leanprover/lean/pull/1090

PR #1056, opened 2016-05-19
limits

avigad:
The file `library/theories/topology/limit` now contains a nice general theory of limits, modeled after Isabelle's. I made some executive decisions regarding names and notation, described in detail in the comments to that file.

I installed the new limits in the analysis folders (metric_space, normed_space, real_limits), but those need to be refactored.

To do:
- in `library/theories/topology`, the theory of continuity should work with the new limits
- more things should be done in `topology`, in full generality
- the proofs in analysis should be simplified


johoelzl:
Nice!

What we recently also introduced in Isabelle/HOL is quantifier syntax for eventually (and its deMorgan-dual frequently): sth like `∀_f x \in F, ...`. Then the statements read like 'for all x in the neighbourhood of z, ...'.

avigad:
@johoelzl Good point. I just tried this:

``` lean
notation `∀₀` binders ` ≈ ` F `, ` r:(scoped:1 P, P) := eventually r F

section
  variables (F : filter ℝ) (s : set ℝ) (f : ℝ → ℝ) (y ε : ℝ)

  check ∀₀ x ≈ F, x ∈ s
  check ∀₀ x ≈ [at ∞], x > 0
  check ∀₀ x ≈ [at y], f x ∈ s
  check ∀₀ x ≈ [at y+], f x ∈ s
end
```

It looks pretty good. On the other hand, it is not that much clearer than

``` lean
  check eventually (λ x, x ∈ s) F
  check eventually (λ x, x > 0) [at ∞]
  check eventually (λ x, f x ∈ s) [at y]
  check eventually (λ x, f x ∈ s) [at y+]
```

and users have to learn the new notation, and get used to associating it to the word "eventually". It might not be worth it, given that users should have to deal with "eventually" all that often. I'll think about it. (If anyone else has an opinion, please share!)

Another concern: right now, to use limits effectively in analysis, we have to open `set` `set.filter` and `topology`. It might help to consolidate namespaces.

robertylewis:
This looks great!

@soonhokong (and others who know Git better than I do): I have local work that I need to adjust based on these commits. Suppose I rebase my local commits on avigad:master and make those adjustments. Will there be problems merging that into leanprover:lean later on, after this pull request and others are merged?

If so, @leodemoura , could I beg you to merge this one as soon as you can? :)

soonhokong:
screen shot 2016-05-19 at 3 53 55 pm

@rlewis1988, I don't see a problem. `avigad/master` is [fast-forwardable](http://ariya.ofilabs.com/2013/09/fast-forward-git-merge.html) from `blessed/master`. So, if your commits are based on `avigad/master`, it should be all good.

avigad:
@soonhokong @leodemoura I changed the link in README.md from the pygments Latex method to the lstlisting method.

johoelzl:
@avigad regarding the forall-syntax for eventually: I think it appears more and more often in Isabelle. Also, from time to time we also change some `exists delta s.t. for all x, P x`. And when the user interface allos to just clicks on the quantifier to go to the `notation` line it shouldn't be too bad to learn a new syntax.

avigad:
@johoelzl O.k., I am convinced!

Issue #694, opened 2015-06-23
heuristic for insterting [unfold-c] and [constructor]


fpvandoorn:
It is annoying to tag everything with the `[unfold-c _]` and `[constructor]` attributes. I'm adding them mostly when I'm actually using them, but still they occur already a lot in the HoTT-library.

We should have good heuristics for adding them automatically. This also requires a method to remove either of these attributes from a definition, because no heuristic is perfect.

For `[unfold-c]`: If we define a definition and only a single `[recursor]` occurs in the definition and the major premise of that recursor is an argument of the definition, then add the attribute `[unfold-c n]`, where `n` marks the place of the major premise. 

The idea is that the unfolded version is "simpler" because we can reduce the recursor. However, the actual unfolded term might still be much longer than the folded definition. 

Here is a conservative variant:
Same as above, but also require that all minor premises are either variables (terms of the form `(fun a_1 a_2 ... a_n, x` where `x` is a variable in the local context (can be `a_i` or another variable)) or have a `[constructor]` as head. 

Example: in `hott/init/path`, `concat` would fall in the first category, and `inverse` in the second. Even though this is a very restrictive condition, loads of definitions fall into this conservative variant (e.g. most definitions in `hott/init/path`)

---

For `[constructor]`: if the head of the definition is a `[constructor]` then this definition is also a `[constructor]`. The idea is that the new definition is a "special case" of the old definition, so if you would want to unfold terms with the previous definition, then you also want to do it with the new one.

The only counterexamples I've found so far is in the algebraic hierarchy, where we don't want to add this attribute from definitions defining an parent `structure` from an extended `structure`. We could add the exception that automatically generated definitions like this do not have this attribute.

leodemoura:
I'm planning to mark this feature as `postponed` and close it. Unless someone volunteers to implement it.
I will not work on this kind of feature in the near future.

Kha:
I applied `[unfold]` to quite a few definitions defined by pattern matching today. In general, if a definition starts by applying a function with an `unfold` attribute (or a primitive recursor), I think it would be a safe heuristics to just copy that attribute. If this sounds reasonable, I can give it a try.

Kha:
Ah, I should have read the the opening post more closely. Floris's heuristic makes much more sense and shows that the implementation is probably not that trivial. I'd start with definitions created by the definitional package then, that way I shouldn't have to worry about major and minor premises.

Issue #776, opened 2015-08-06
folding fails


fpvandoorn:
In the following example there are two places where folding fails with `esimp`.

``` lean
-- HoTT
open nat unit eq equiv

  definition code [reducible] [unfold 1 2] : ℕ → ℕ → Type₀
  | code 0 0               := unit
  | code (succ n) 0        := empty
  | code 0 (succ m)        := empty
  | code (succ n) (succ m) := code n m

  definition refl [unfold 1] : Πn, code n n
  | refl 0 := star
  | refl (succ n) := refl n

  definition encode [unfold 3] {n m : ℕ} (p : n = m) : code n m :=
  p ▸ refl n

  definition decode [unfold 1 2] : Π(n m : ℕ), code n m → n = m
  | decode 0 0               := λc, idp
  | decode 0 (succ l)        := λc, empty.elim c _
  | decode (succ k) 0        := λc, empty.elim c _
  | decode (succ k) (succ l) := λc, ap succ (decode k l c)


  definition nat_eq_equiv (n m : ℕ) : (n = m) ≃ code n m :=
  equiv.MK encode
           !decode
           begin
             revert m, induction n, all_goals (intro m;induction m;all_goals intro c),
               all_goals try contradiction,
               induction c, reflexivity,
               try (esimp; state; now), exact sorry
           end
           begin
             intro p, induction p, esimp [encode], induction n,
               reflexivity,
               esimp, state, exact sorry
           end
```

I get the expected output when I remove the `unfold` attribute from `refl` and `decode` and unfold them manually:

``` lean
open nat unit eq equiv

  definition code [reducible] [unfold 1 2] : ℕ → ℕ → Type₀
  | code 0 0               := unit
  | code (succ n) 0        := empty
  | code 0 (succ m)        := empty
  | code (succ n) (succ m) := code n m

  definition refl : Πn, code n n
  | refl 0 := star
  | refl (succ n) := refl n

  definition encode [unfold 3] {n m : ℕ} (p : n = m) : code n m :=
  p ▸ refl n

  definition decode : Π(n m : ℕ), code n m → n = m
  | decode 0 0               := λc, idp
  | decode 0 (succ l)        := λc, empty.elim c _
  | decode (succ k) 0        := λc, empty.elim c _
  | decode (succ k) (succ l) := λc, ap succ (decode k l c)


  definition nat_eq_equiv (n m : ℕ) : (n = m) ≃ code n m :=
  equiv.MK encode
           !decode
           begin
[...]

PR #1059, opened 2016-05-24
[WIP] Automatically add [unfold] hints to equations definitions

Kha:
This is a preliminary implementation of https://github.com/leanprover/lean/issues/694#issuecomment-220469553. It applies `[unfold ...]` to any definition that is non-mutually defined by pattern matching. It can handle implicit arguments in the pattern and section variables.

Some observations:
- The attribute is applied to 187 definitions in the stdlib.
- I was able to remove 11 manual applications of the attribute (all in `hott/`, I guess the attribute isn't used widely yet) and some explicit unfold steps in proofs.
- There were 2 'false positives' where the attribute complicated proofs. I modified the parser so that `[unfold]` without arguments would be accepted and remove the unfold hint (which may not be the most readable solution).
- I applied the parameter-less `[unfold]` to 4 more recursive definitions because, unlike with an explicit unfold step, the recursive occurences weren't folded back. @leodemoura might this be something easy to change?

### DIFF
--- a/hott/homotopy/LES_of_homotopy_groups.hlean
+++ b/hott/homotopy/LES_of_homotopy_groups.hlean
@@ -515,7 +515,7 @@ namespace chain_complex
   definition str_of_nat {n : ℕ} : ℕ → ℕ × fin (succ n) :=
   λm, (m / (succ n), mk_mod n m)
 
-  definition nat_of_str_3S [unfold 2] [reducible]
+  definition nat_of_str_3S [reducible]
     : Π(x : stratified +ℕ 2), nat_of_str x + 1 = nat_of_str (@S (stratified +ℕ 2) x)
   | (n, fin.mk 0 H) := by reflexivity
   | (n, fin.mk 1 H) := by reflexivity
--- a/hott/types/fin.hlean
+++ b/hott/types/fin.hlean
@@ -315,7 +315,7 @@ end
 definition mk_pred (i : nat) (h : succ i < succ n) : fin n :=
 mk i (lt_of_succ_lt_succ h)
 
-definition succ : fin n → fin (succ n)
+definition succ [unfold] : fin n → fin (succ n)
 | (mk v h) := mk (nat.succ v) (succ_lt_succ h)
 
 lemma val_succ : Π (i : fin n), val (succ i) = nat.succ (val i)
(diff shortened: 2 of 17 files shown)


fpvandoorn:
Here are some remarks about the `[unfold _]` attribute.
- The folding back of recursive definitions with `esimp` is discussed [here](https://github.com/leanprover/lean/issues/776#issuecomment-128988348). Because `unfold` doesn't work well with `esimp`, I typically don't mark recursive definitions with the `unfold` attribute.
- [Here](https://github.com/fpvandoorn/lean/commit/f62af59f36136c54e57763eced2435b6a65ab907#commitcomment-12455450) and [here](https://github.com/leanprover/lean/pull/754#issuecomment-126696860) is a discussion about another potential problem of adding the `unfold` attribute to recursive definitions.
- The `unfold` attribute is widely used in the HoTT library (for example, [this](https://github.com/leanprover/lean/blob/master/hott/init/path.hlean) file contains 43 unfold hints). The reason you could only remove few manual applications is that pattern matching is not widely used.

fpvandoorn:
Another comment: some functions compute only on `constructors` of a particular kind. For example, a function `X × Y × Z → W` might only compute if the second component of the pair is itself again a pair. Here is a dummy-example:

``` lean
definition f : (Σ(x y : ℕ), x = y) → unit
| ⟨x, y, p⟩ := star
```

There is currently no way to specify that this definition only reduces on triples (and I don't think it is a good use of time to implement this).

These examples do come up in practice, but it has never been a problem for me. I just give such definitions the `unfold` attribute, and in all cases where I use such definitions in practice and want them to reduce, they are applied to constructors of the right form so that they actually reduce.

Kha:
@fpvandoorn Thank you for the helpful links. I'll have to see how to exclude recursive definitions; there is some code for that in the equations compiler, but it isn't straightforward to adapt.

About the nested constructors, it's good to hear that they shouldn't be a problem in practice. Should a pattern ever prove to be so, one can always apply the parameter-less `[unfold]` manually.

leodemoura:
@Kha I will merge this pull request in the `main` branch. In the `lean3` branch we have support for unification hints, and the `esimp` command will behave differently. So, I will not merge it in the `lean3` branch for now. We can decide later.

Kha:
> the esimp command will behave differently.

Oh, I see! In that case, it may be easier not to merge the PR; I still haven't added a check for recursive definitions. Even for `master`, the benefits would be quite limited.

PR #1060, opened 2016-05-24
Support C++14 sized delete

Kha:
I experimented with actually making use of the size information, but there was no noticeable speedup.


leodemoura:
@Kha and @soonhokong I will merge this pull request in the `main` and `lean3` branches later today.

PR #1062, opened 2016-05-25
[WIP] Add option to hide binder types

Kha:
I'm trying to make Lean's flycheck output a bit more clear, which is sorely needed with the kinds of proofs I will presumably be tackling in the next six months:
2016-05-25-134119_785x415_scrot

This patch is a first step. It tries to make the pretty printer behave more like an actual user by hiding the types of binders by default, except in type signatures:

``` lean
-- not the actual definition
notation `do` binder ` ← ` x `; ` r:(scoped f, id f x) := r

definition f (x : nat) := λy {z}, do w ← x - y + z; w ≠ x
print definition f

set_option pp.binder_types true
print definition f
```

```
print result:
 definition f : ℕ → ℕ → (Π {z : ℕ}, Prop)
 λ x y {z}, do w ← x - y + z; w ≠ x (lean-checker)
print result:
 definition f : ℕ → ℕ → (Π {z : ℕ}, Prop)
 λ (x y : ℕ) {z : ℕ}, do w : ℕ ← x - y + z; w ≠ x (lean-checker)
```

This is especially useful for binding notations like the monadic do notation I'm using. What do you think, would this generally be desirable?


fpvandoorn:
So if I write `print f` then in the type of `f` all binder types are still displayed? 

In the terms I usually do not care about the type of binders, so I think it would be good to not print them by default.

Kha:
Yes, `print f` would produce the same output as the first `print definition` above (except for an additional `:=`, curious).

leodemoura:
@Kha this option is useful. We have a similar option in the `lean3` branch.
We also have builtin support for the `do` notation. Examples:
https://github.com/leanprover/lean/blob/lean3/tests/lean/run/IO1.lean
https://github.com/leanprover/lean/blob/lean3/tests/lean/run/IO3.lean
https://github.com/leanprover/lean/blob/lean3/tests/lean/run/IO4.lean

The `do` notation works for any `monad` instance:
https://github.com/leanprover/lean/blob/lean3/library/init/monad.lean

leodemoura:
@Kha I will try to merge this pull request in the `main` and `lean3` branches.

Kha:
I will have to update the PR before it's ready to merge, this is just the bare-bones implementation for demonstration. I also haven't updated any tests yet.

Kha:
> We also have builtin support for the do notation.

Ah, that's interesting. My notation is specialized to the `option` monad currently to model nontermination, but it would be interesting to transparently replace it with other monads, e.g. a state monad that counts the number of `bind`s, which should be an accurate measure of asymptotic runtime.

Kha:
I took a look at the `lean3` branch and noticed that the `pp.hide_binder_types` option there is indeed _very_ similar - that's quite some convergent evolution :) . The differences I noticed are:
- defaults to not hiding
- also hides binder info (I don't mind these since they don't take up much space)
- no special case for definitions' type signatures, which makes sense with the negated default

So, with which version should we go? I haven't noticed any problems with hiding the types by default yet.

leodemoura:
> So, with which version should we go?

I am planning to keep yours since you keep the binder info.

Kha:
Ok, cleaned up code, added a test, and updated existing tests (using `for f in **/*.produced.out; do cp $f ${f%.produced.out}.expected.out; done`).

PR #1063, opened 2016-05-25
Update analysis theory to use new limits

robertylewis:
This pull request includes all of @avigad 's commits in #1056  . 

Various things have happened here:
- The Frechet derivative is defined on normed vector spaces, and some specializations to the reals are shown
- Rules about continuity in analysis are added
- The analysis theory is adapted to the new theory of limits in topology, which means that a lot of theorems are significantly more general

There is still a bit of cleaning to be done, but I think that things are named correctly and in the right places....


avigad:
Nice! 

I noticed some naming inconsistencies: we have constants `derivative_at` and `is_frechet_deriv_at`, but also theorem names that refer to `deriv_at` and `frechet_derivative_at`! Also, the name `is_frechet_deriv_at` doesn't fit well with the order of arguments, e.g. `is_frechet_deriv_at f g x` says that `g` is the frechet derivative of `f` at `x` rather than (as the name suggests) saying that `f` is the frechet derivative of `g` at `x`.

I think my vote is to use `has_deriv_at f d x` for "`f` has derivative `d` at `x`", and `has_frechet_deriv_at f g x` for "`f` has frechet derivative `g` at `x`." This keeps the order of arguments, and involves only changing the name.

If we later decide to have function versions (I would argue that we should do this with differentiability as a precondition), we could use `deriv_at f x` for the derivative of `f` at `x`, and `frechet_deriv_at f x` for the frechet derivative. (I suppose the corresponding predicates should be `differentiable_at f x`).

What do you think? I would be glad to do these renamings one day.

I think ultimately we will also need `differentiable_at_within` and `frechet_differentiable_at_within`. For example, many theorems from calculus say e.g. that `f` is differentiable on an interval `'[a, b]`, which amounts to saying that `f` is differentiable at `x` within `'[a, b]` for every `x` in `'[a, b]`. In my experience, such local versions come up all the time, and are unavoidable.

Will that generalization be hard? I think it is mostly a matter of changing the limits "at" to "at_within" everywhere.

robertylewis:
> What do you think? I would be glad to do these renamings one day.

Yes, those make sense. I renamed a bunch of things in analysis, but for some reason didn't even look at the names in that file. I'll take care of it.

> Will that generalization be hard? I think it is mostly a matter of changing the limits "at" to "at_within" everywhere.

That's next on my to-do list. I just wanted to make this stage public sooner rather than later!

avigad:
It looks really great. It is exciting that the analysis library is finally showing signs of life!

robertylewis:
I've added a commit to fix some names.

Other questions/random thoughts that come up here:
- The definition `definition has_frechet_deriv_at (f A : V → W) [is_bdd_linear_map A] (x : V)` is a little inconvenient because of the inferred proof of `is_bdd_linear_map A`. For example, if `A = B`, we can't directly substitute B into `has_frechet_deriv_at f A x`. The "unbundled" approach gets us into trouble here. Since that hypothesis isn't needed for the definition, we could remove it and just require it for the appropriate theorems, but maybe that could get ugly?
- I think the proof of the chain rule `has_frechet_derivative_at_comp` is the longest single proof I've done, but there's no obvious refactoring -- I need to think about that a bit more. Surprisingly, it takes less than a second to elaborate. But I use `simp` a handful of times, and if I don't clear some hypotheses before doing that, the elaboration time more than doubles. (To be specific: I'm only clearing four to eight irrelevant hypotheses.) Is this expected behavior for simp?

avigad:
The first is an interesting question.

`has_frechet_deriv_at f A x` is a property of `f`, `A`, and `x`, and it feels morally wrong to make it depend on something else as well. So my gut reaction would be to avoid the dependence. I don't know whether we need `has_frechet_deriv_at f A x` to entail that `A` is a bounded linear map, but if we do, it seems to me that the natural thing to do is to make that part of the definition. For example, it could be a Prop-valued structure extending `is_bdd_linear_map A`.

But what exactly is the problem? I just tried this, and it works:

``` lean
import theories.analysis.bounded_linear_operator
open analysis

variables {V W : Type}
variables [normed_vector_space V] [normed_vector_space W]
variables f A B : V → W
premise   [is_bdd_linear_map A]

example (H : A = B) : is_bdd_linear_map B :=
by rewrite -H; assumption
```

avigad:
Oops, wait, I got my example wrong...

avigad:
I see the problem now. There is no way to make this rewrite work:

``` lean
import theories.analysis.bounded_linear_operator
open analysis

variables {V W : Type}
variables [normed_vector_space V] [normed_vector_space W]
variables f A B : V → W
variable  x : V
premises  [is_bdd_linear_map A] [has_frechet_deriv_at f A x]
premises  [is_bdd_linear_map B]

example (H : A = B) : has_frechet_deriv_at f B x :=
begin rewrite -H end
```

It seems that this sort of problem could come up a lot. So we need to find the ideal solution. I think it is just to eliminate the dependence.

robertylewis:
Yes, you can do that, but you can't do this:

``` lean
import theories.analysis.bounded_linear_operator
open analysis

variables {V W : Type}
variables [normed_vector_space V] [normed_vector_space W]
variables f A B : V → W
variable x : V
premise   [is_bdd_linear_map A]
premise H : has_frechet_deriv_at f A x

example (H : A = B) : has_frechet_deriv_at f B x :=
by rewrite -H; assumption
```

or this:

``` lean
import theories.analysis.bounded_linear_operator
open analysis

variables {V W : Type}
variables [normed_vector_space V] [normed_vector_space W]
variables f A B : V → W
variable x : V
premise   [is_bdd_linear_map A]
premise [is_bdd_linear_map B]
premise H1 : has_frechet_deriv_at f A x

example (H : A = B) : has_frechet_deriv_at f B x :=
begin rewrite -H; assumption end
```

I don't think you can extend `is_bdd_linear_map` to a Prop-valued structure, since it isn't prop-valued itself.  `is_bdd_linear_map A` carries data, the operator norm of A.

avigad:
Our messages crossed. But regarding this:

```
I don't think you can extend is_bdd_linear_map to a Prop-valued structure, since it isn't prop-valued itself. is_bdd_linear_map A carries data, the operator norm of A.
```

I think anything that starts with `is` should be Prop-valued and carry no data. It should be proof irrelevant; we don't want expressions that depend on it to depend on the particular bound. 

In this case, the class should just say that there exists a bound. If there are places you need to use a bound, you can use exists elim, or, if a definition depends on it, `some`.

fpvandoorn:
I think it's good to eliminate unnecessary dependences like that one. 

The rewrite-problem can also occur in places where it's not possible to remove such dependencies. In that case, this is a workaround:

``` lean
import theories.analysis.bounded_linear_operator
open analysis

variables {V W : Type}
variables [normed_vector_space V] [normed_vector_space W]
variables f A B : V → W
variable  x : V
premises  [H1 : is_bdd_linear_map A] [H3 : has_frechet_deriv_at f A x]
premises  [H2 : is_bdd_linear_map B]
include H2 H3
example (H : A = B) : has_frechet_deriv_at f B x :=
begin revert [H2], rewrite [-H], intro, exact H3 end
```

robertylewis:
> I think anything that starts with is should be Prop-valued and carry no data. It should be proof irrelevant; we don't want expressions that depend on it to depend on the particular bound.

Fair enough. This change was easy to make.

> I think it's good to eliminate unnecessary dependences like that one.

I'm conflicted. This seems like a shortcoming of the bundled approach. Really, in the definition

``` lean
definition has_frechet_deriv_at (f A : V → W) [is_bdd_linear_map A] (x : V) :=
  (λ h : V, ∥f (x + h) - f x - A h ∥ / ∥ h ∥) ⟶ 0 [at 0]
```

we should have something like `A : bdd_linear_map V W`. Eliminating the "unnecessary" dependency means that you can prove things like `∀ f x, has_frechet_deriv_at f (λ h, f (x + h) - f x) x`, which makes me uncomfortable, even if all the derivative theorems add the requirement [is_bdd_linear_map A].

But this seems better to me than this:

> For example, it could be a Prop-valued structure extending is_bdd_linear_map A.

Then we'd need to use an intro rule every time we wanted to prove something has a derivative, and get a subgoal that the derivative is linear. I think that would get frustrating fast.

johoelzl:
In Isabelle `has_frechet_derivative f Ax` includes the assumption that `A` is a bounded linear function, i.e. similar to:

``` lean
structure has_frechet_deriv_at (f A : V → W) (x : V) :=
  (linear : is_bdd_linear_map A)
  (lim : λ h : V, ∥f (x + h) - f x - A h ∥ / ∥ h ∥) ⟶ 0 [at 0])
```

Also when Fabian introduced a type of bounded linear functions we thought about:

``` lean
definition has_frechet_deriv_at (f : V → W) (A : bdd_linear_map V W) (x : V) :=
  λ h : V, ∥f (x + h) - f x - A h ∥ / ∥ h ∥) ⟶ 0 [at 0]
```

But `bdd_linear_map` is hard to do well in Isabelle, I would assume it works better in Lean with a category type classes.

avigad:
Both approaches seem reasonable. We have had endless discussions about whether it is better to bundle things like `bdd_linear_map` or leave them unbundled, as in `is_bdd_linear_map`. We never seem to come to a conclusion, but, at least at one point, Leo felt strongly that the unbundled version will work better with automation. He and Daniel have worked hard on indexing techniques to quickly find relevant theorems using predicates like `is_bdd_linear_map`, and they seem to have solved that problem. On the other hand, there is no mechanism to easily recognize that an unbundled object could be bundled to apply a bundled theorem.

In algebra, so far I have been happy with the unbundled `is_hom` instead of a bundled `hom`, so my inclination would be to try that here first.

@rlewis1988 I think the biggest downside is having to type `has_frechet_deriv_at_intro`. If the `is_bdd_linear_map` part can typically be inferred by class inference, then we can tag that as an implicit argument to the intro rule. After applying the intro rule, then, the goal will be exactly the same as what you have now. (And if the `is_bdd_linear_map` can't be inferred, you'll get the same error message you have now.)

On the other hand, other ways of showing `has_frechet_deriv_at` will be able to synthesize the information from other sources. For example, composition rules and derivative rules for fixed functions "know" that the derivatives are bdd linear maps by construction, and so they can save the class inference (and potentially the user) the trouble of finding proofs of that fact. So it seems to me that there are advntages to keeping it together with what it means to be a frechet derivative.

Finally, don't hesitate to experiment and mess around with naming. If typing `frechet_deriv` all the time is really annoying, we can shorten it to `fderiv`. So, for example, we would have `has_fderiv_at_intro`.

I think our real goal is to get automation good enough so that users never have to type that at all...

PR #1065, opened 2016-05-30
Compute correct flycheck message width

Kha:
Subtracts the widths of all other columns from the buffer width. Somewhat relatedly, we might want to hide that unused 'ID' column.

I begin to regret never having learned Lisp.


arademaker:
> I begin to regret never having learned Lisp. 

@Kha , why ? Never is to late! http://www.gigamonkeys.com/book/

PR #1071, opened 2016-06-10
Allow shadowing variables in patterns

Kha:
When it comes to shadowing variables, Lean can currently be a bit inconsistent:

``` lean
example (xs xs : list ℕ) : -- works
  list ℕ → list ℕ → list ℕ
| [] := λxs, xs -- works
-- | (x::xs) := begin -- 'invalid recursive equation'
| (x::xs') := begin
  intro xs, -- works
  cases xs with x xs, -- actually x_1 xs_1, but that's an issue for another day
end
```

The inability to shadow variables in patterns presented a problem for my current project, where I generate tuple unpacking code like the following:

``` lean
let x := 1 in
do tmp ← f x; match tmp with -- maybe even `do (r, x) ← f x; ...` in the future...?
(r, x) := ...
end
```

Having to rename that second `x` would add quite some additional complexity to my transpiler, so it was easier to lift the restriction in Lean.

Now, the ability to shadow variables in patterns can often be confusing to beginners in functional programming, but I believe most people eventually embrace it, which I've even done subconsciously in my own proofs after implementing the feature. As far as I know, it was mostly a technical restriction anyways.

### DIFF
--- a/src/frontends/lean/decl_cmds.cpp
+++ b/src/frontends/lean/decl_cmds.cpp
@@ -441,13 +441,8 @@ static bool is_curr_with_or_comma_or_bar(parser & p) {
 }
 
 /**
-   For convenience, the left-hand-side of a recursive equation may contain
-   undeclared variables.
-   We use parser::undef_id_to_local_scope to force the parser to create a local constant for
-   each undefined identifier.
-
-   This method validates occurrences of these variables. They can only occur as an application
-   or macro argument.
+   This method validates occurrences of local variables (i.e., variables bound in the pattern).
+   They can only occur as an application or macro argument.
 */
 static void validate_equation_lhs(parser const & p, expr const & lhs, buffer<expr> const & locals) {
     if (is_app(lhs)) {
--- a/src/frontends/lean/decl_cmds.cpp
+++ b/src/frontends/lean/decl_cmds.cpp
@@ -542,42 +542,45 @@ static void parse_equations_core(parser & p, buffer<expr> const & fns, buffer<ex
     while (true) {
         expr lhs;
         unsigned prev_num_undef_ids = p.get_num_undef_ids();
-        buffer<expr> locals;
-        {
-            parser::undef_id_to_local_scope scope2(p);
-            buffer<expr> lhs_args;
-            auto lhs_pos = p.pos();
-            if (p.curr_is_token(get_explicit_tk())) {
-                p.next();
-                name fn_name = p.check_decl_id_next("invalid recursive equation, identifier expected");
-                lhs_args.push_back(p.save_pos(mk_explicit(get_equation_fn(fns, fn_name, lhs_pos)), lhs_pos));
-            } else {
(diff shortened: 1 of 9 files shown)


src/frontends/lean/decl_cmds.cpp
```diff
@@ -542,42 +542,45 @@ static void parse_equations_core(parser & p, buffer<expr> const & fns, buffer<ex
...
-                    lhs_args.push_back(p.save_pos(mk_explicit(fns[0]), lhs_pos));
-                    lhs_args.push_back(first);
-                } else {
-                    throw parser_error("invalid recursive equation, head symbol in left-hand-side is not a constant",
-                                       lhs_pos);
-                }
+                throw parser_error("invalid recursive equation, head symbol in left-hand-side is not an equation function",
+                                   lhs_pos);
             }
+        }
```
Kha, line 559: This refactoring turned out to be unnecessary after all, but I think it simplifies the function's logic

src/frontends/lean/parser.cpp
```diff
@@ -107,6 +107,8 @@ parser::undef_id_to_const_scope::undef_id_to_const_scope(parser & p):
     flet<undef_id_behavior>(p.m_undef_id_behavior, undef_id_behavior::AssumeConstant) {}
 parser::undef_id_to_local_scope::undef_id_to_local_scope(parser & p):
     flet<undef_id_behavior>(p.m_undef_id_behavior, undef_id_behavior::AssumeLocal) {}
+parser::local_and_undef_id_to_local_scope::local_and_undef_id_to_local_scope(parser & p):
+    flet<undef_id_behavior>(p.m_undef_id_behavior, undef_id_behavior::AssumeLocalAndAlsoDefinedLocals) {}
```
Kha: I'm very open to naming suggestions :)

tests/lean/run/eq23.lean
```diff
@@ -10,7 +10,7 @@ with tree_list :=
 namespace tree_list
 
 definition len {A : Type} : tree_list A → nat
-| len (nil A)    := 0
```
Kha, line 13: I believe this _should_ already have been illegal under the old rules, but `validate_pattern` didn't catch it.

Kha:
I've added a second commit that also allows shadowing definitions that are not constructors, see the [lean-user thread](https://groups.google.com/forum/#!topic/lean-user/DtZShjFAvX4).

leodemoura:
I'm currently spending all my time on the `lean3` branch. I didn't have time to read the lean-user thread yet. Let's discuss the problems and this pull request in our meeting in the beginning of July.

leodemoura:
```
do tmp ← f x; match tmp with -- maybe even `do (r, x) ← f x; ...` in the future...?
(r, x) := ...
```

The notation `do (r, x) ← f x` is supported in the `lean3` branch. Example:
https://github.com/leanprover/lean/blob/lean3/library/init/meta/injection_tactic.lean#L44

PR #1072, opened 2016-06-20
refactor(library/data/pnat): make pnat a decidable_linear_order

gebner:
This makes the K₂_symm theorem 10 lines shorter, as it is now just the commutativity of max.  A few other lemmas in pnat.lean are now subsumed by general ones about linear orders as well.

However, I could not get the instance declaration to work without resorting to tactics mode:

``` lean
  le_refl         := begin apply pnat.le_refl end,
```

With just `le_refl := pnat.le_refl`, I get the following error:

```
/home/gebner/lean/library/data/pnat.lean:98:0: error: type mismatch at application
  @decidable_linear_order.mk pnat pnat.le pnat.le_refl
term
  pnat.le_refl
has type
  ∀ a,
    @le pnat pnat_has_le a a
but is expected to have type
  ∀ a,
    pnat.le a a
```


robertylewis:
Cool, thanks! I think the pnat file was done before the general theory of min/max was implemented, so this is nice to see.

I'm not sure why that error is coming up. This isn't a fix, but to be very slightly briefer, you could write `by apply pnat.le_refl` instead of using `begin ... end`. (`by t` is syntax to apply a single tactic.)

gebner:
Thanks for the pointer!  It's a bit less verbose now, but I still have no idea why it's required at all.

PR #1076, opened 2016-06-23
van Kampen theorem

fpvandoorn:
- Many small changes in the HoTT library. Most notably I've proven the Seifert-van Kampen Theorem (the full version with a type of basepoints).
- Changes which are most likely to break proofs: `pathover_of_eq` now has a second explicit argument, and I've renamed `apo011[1] -> apd011[1]` and `apd011[1...] -> apdt011[1...]`
- The commits 05f8777 and 65eae47 are changes in the `extras` folder and are also useful for the `lean3` branch.

### DIFF
--- a/extras/depgraph/README.md
+++ b/extras/depgraph/README.md
@@ -3,7 +3,11 @@ Generating Dependency Graph
 
 Usage:
 
-    leandeps.py [options] dir/file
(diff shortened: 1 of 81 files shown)


fpvandoorn:
Added one more commit.

PR #1077, opened 2016-06-24
refactor(simplifier): port skeleton to new tactic framework

dselsam:
Note: many features are missing still, and the contextual cache is not working yet.


leodemoura:
merged in `lean3` branch

PR #1081, opened 2016-07-02
Simplifier: basic support for extensions and canonicalizing instances

dselsam:


src/library/defeq_canonicalizer.cpp
```diff
@@ -11,37 +11,37 @@ Author: Leonardo de Moura
 #include "library/cache_helper.h"
 
 namespace lean {
-struct defeq_canonize_cache {
+struct defeq_canonicalize_cache {
     environment           m_env;
     /* Canonical mapping I -> J (i.e., J is the canonical expression for I).
        Invariant: locals_subset(J, I) */
-    expr_map<expr>        m_C;
+    expr_struct_map<expr>        m_C;
```
dselsam, line 18: This one took me a while to track down.

src/library/defeq_canonicalizer.cpp
```diff
@@ -116,6 +116,7 @@ struct defeq_canonize_fn {
             if (get_weight(e) < get_weight(*new_e) && locals_subset(e, *new_e)) {
                 replace_C(*new_e, e);
                 replace_M(*h, *new_e, e);
+                insert_C(e, e);
```
dselsam, line 119: Is there a reason you were not caching here?

src/library/defeq_canonicalizer.cpp
```diff
@@ -11,37 +11,37 @@ Author: Leonardo de Moura
 #include "library/cache_helper.h"
 
 namespace lean {
-struct defeq_canonize_cache {
+struct defeq_canonicalize_cache {
```
leodemoura, line 14: Canonizer is the name commonly used:

http://link.springer.com/chapter/10.1007%2F978-3-540-31862-0_27
http://arxiv.org/pdf/1207.3262.pdf
https://books.google.pt/books?id=B_tX0BdH9ScC&pg=PA442&lpg=PA442&dq=Shostak+canonizer&source=bl&ots=RnC-WVphDL&sig=4rN_JrRuw0ar5YlkE8jqkRpWQbU&hl=en&sa=X&ved=0ahUKEwjv-ePd89XNAhUqDMAKHQz3CMY4ChDoAQgsMAI#v=onepage&q=Shostak%20canonizer&f=false

src/library/defeq_canonicalizer.cpp
leodemoura, line 119: Just forgot it.

library/init/meta/tactic.lean
```diff
@@ -297,6 +297,20 @@ do (new_target, Heq) ← target >>= simplify,
...
+               match expr.is_eq tgt with
+               | option.none := fail "simplifier extension expects a goal of the form [ctx |- l = ?r]"
+               | option.some (start, res) := do r ← simp_ext start,
+                                                unify res (prod.pr1 r),
+                                                unify g (prod.pr2 r)
+               end
+   | _ := fail "simplifier extension expects a goal of the form [ctx |- l = ?r]"
+   end
+
+
```
leodemoura, line 313: The new `do-match` notation can be used to simplify the code above. The `exact` and `assert` tactics can also be used to avoid low-level primitives such as `get_goals`.

```
meta_definition mk_eq_simp_ext (simp_ext : expr → tactic (expr × expr)) : tactic unit :=
do (lhs, rhs)     ← target >>= match_eq,
   (new_rhs, Heq) ← simp_ext lhs,
   unify rhs new_rhs, 
   exact Heq
```

leodemoura:
Integrated, but dropped the "canonize => canonicalize" change.

dselsam:
For what it is worth:
https://en.wikipedia.org/wiki/Canonicalization
https://en.wikipedia.org/wiki/Canonization_(disambiguation)

leodemoura:
> For what it is worth:
> https://en.wikipedia.org/wiki/Canonicalization
> https://en.wikipedia.org/wiki/Canonization_(disambiguation)

I'm aware of this wiki page.
My point is that there are hundreds of papers in automated reasoning that use `canonize`/`canonizer`.
I never seen a single paper in this area using `canonicalizer`.

avigad:
I also like the idea of nominating Lean terms for sainthood.

PR #1084, opened 2016-07-06
test(tests/lean/run): add three perf tests

dselsam:


dselsam:
Should I add the script that produces these instead/as-well?

leodemoura:
> Should I add the script that produces these instead/as-well?

I think it would be nice to also add the script.

PR #1087, opened 2016-07-06
[WIP] Generalized and user-defined attributes

Kha:
- [x] use `scoped_ext` for generic attribute storage
  - currently breaks `print [attr]`, which will have to be revised anyway as soon as attributes aren't keywords anymore
- [ ] improve attribute syntax
  - [ ] switch to `definition [attr] name` to make syntax unambiguous
  - [ ] support `[attr1, attr2]`
  - [ ] some syntax for removing attributes?
- [ ] ~~support 'global by default' attributes (`instance`)~~
- [x] **make all attributes global**
- [ ] actually use `attribute_decl.m_descr` somewhere
- [ ] support user-defined attributes
  - [ ] with custom validators


Kha:
Some more progress: a more or less stable version without metaclasses and persistent scopes. I realized quite late that `export` is the one thing that should still behave like an old `scoped_ext`, so I had to reimplement a simple version of it, tracking a map of all `export_decls` in a plain `environment_extension` and a list of active `export`s in a (new) `scoped_ext`.

Regarding making attribute declarations go in front of the definition name, @avigad mentioned that he would much prefer the current syntax. It seems to me like that shouldn't be too hard to implement even with user-defined attributes using a small look-ahead - check if the id behind `[` is the name of a registered attribute.

leodemoura:
@Kha Thanks for the update.

> Regarding making attribute declarations go in front of the definition name, @avigad mentioned that he would much prefer the current syntax.

There are many issues. The change is not just motivated by avoiding the look-ahead. 

1- Error messages are not going to be nice. Example: `definition [smp] foo ...` produces the nice error message `"smp is not an attribute"`. With the current syntax `definition foo [smp] ...`, we will perform the lookahead and check that `smp` is not an attribute. Then, we will try to parse it as a binder, and get back a weird error message about invalid binder declaration. 

2- We want to support user defined attributes. There will be not easy way to syntax highlight these attributes in Emacs if we keep the current syntax. Emacs syntax highlighting code would have to invoke Lean to decide whether something is an attribute name or not.

3- User confusion between attributes and instance implicit arguments. The syntax is identical. The confusion doesn't happen in Lean2 because we have syntax highlight and the set of attributes is fixed and small.

I strongly believe we have to change.

fpvandoorn:
My 2 cents: I wouldn't mind writing `definition [instance] foo` instead of `definition foo [instance]`.

avigad:
I have no objection. My complaint was aesthetic -- as you scan down a file, the theorem names will jump around more. But mostly it is an aversion to change.

Leo's arguments are compelling, and I am sure I will get used to it. (Probably in the long run I will even prefer it.)

dselsam:
I'll throw out one other option. If I recall, only one or two attributes actually need to be provided before the type and value are elaborated. We could have special commands for them, and then only accept regular attributes at the end.

Example:

``` Lean
class has_le ...
instance nat_has_le ...
definition mul_one ...
attributes mul_one [simp, ematch, intro, ...]
```

Kha:
Just to name an alternative, we could also change the delimiter tokens for attributes (or even type classes?) themselves, to something like `⟦simp⟧` or `@[simp]`. That feels like an even bigger change, but shouldn't be harder to refactor. It would also make it _very_ clear to the user that these are two separate features and not arbitrarily placed in front of and behind the declaration name.

If the token is not a valid expression start, we could even prefix the complete declaration with the attributes, like in some other languages:

``` lean
@[simp, intro]
lemma foo [group A] ...
```

Though that does somehow look less fitting to Lean.

avigad:
We could also use notation like

``` lean
lemma foo [attribute: simp] [group A] ...
```

or

``` lean
lemma foo [attr: simp] [group A] ...
```

leodemoura:
> I'll throw out one other option. If I recall, only one or two attributes actually need to be provided before the type and value are elaborated. We could have special commands for them, and then only accept regular attributes at the end.

We discussed this option when Sebastian visited MSR. 
The problem is that `class` can be a definition, an inductive datatype or a structure.

leodemoura:
I think all suggestions have problems. Daniel's suggestions is almost there, but it has the problem I described in the previous post. One workaround is to use: 
- `class foo ...` for `structure foo [class] ...` (the most common case)
- `class inductive foo ...` for `inductive foo [class] ...`
- `class definition foo ...` for `definition foo [class] ...`
  We actually considered this option when Sebastian was visiting MSR, but we thought it was too ad hoc.
  The notation `instance foo ...` for `definition foo [instance] ...` is fine. Actually, I would be quite happy to use it.
  I'm happy to set all other attributes using the `attribute` command.

I think Sebastian's suggestion is fine for attributes such as `simp` and `intro`. I think it would be weird for `class` and `instance`.

Kha:
Remind me again, was there some proposed change for lean3 that would prevent tagging declarations with `class/instance/coercion`(?) after the fact? If so, it would certainly make sense to remove that attribute and introduce some dedicated syntax. If not, I would probably find a special syntax confusing.

leodemoura:
> Remind me again, was there some proposed change for lean3 that would prevent tagging declarations with class/instance/coercion(?) after the fact? If so, it would certainly make sense to remove that attribute and introduce some dedicated syntax. If not, I would probably find a special syntax confusing.

Only `class` has this restriction when declaring structures. When we mark a structure as a `class`, we automatically add definitions showing that the structure is an instance of the parent structures. The automatically generated projections are also different: the structure argument is instance implicit (i.e., `[...]`).

There is no restriction for `instance` and `coercion` attributes.

Kha:
Going back to the original `definition [attrs] id` proposal, we could also increase readability by extending the `attribute` keyword to also accept new declarations:

``` lean
attribute [constructor, ...] -- less central attributes preferably go here
definition[instance] foo ...
```

jroesch:
One feature I would like to request is the ability to define attributes that take arguments. In particular I would like to add an `extern` attribute which allows the user describe foreign code existing in external libraries (`.a`, `.dylib`, `.so`, etc). 

Rust has an easy to use version of this which allows you to specify a set of symbols belonging to an external library. The external library is identified by the name passed in the `#[link(name = ??)]` attribute (see below example). A similar feature would allow our build tooling to automatically provide the correct link arguments when compiling Lean programs to executables, or for dynamic loading if we support execution of such functions inside the VM.

``` rust
#[link(name = "snappy")]
extern {
    fn snappy_compress(input: *const u8,
                       input_length: size_t,
                       compressed: *mut u8,
                       compressed_length: *mut size_t) -> c_int;
    fn snappy_uncompress(compressed: *const u8,
                         compressed_length: size_t,
                         uncompressed: *mut u8,
                         uncompressed_length: *mut size_t) -> c_int;
    fn snappy_max_compressed_length(source_length: size_t) -> size_t;
    fn snappy_uncompressed_length(compressed: *const u8,
                                  compressed_length: size_t,
                                  result: *mut size_t) -> c_int;
    fn snappy_validate_compressed_buffer(compressed: *const u8,
                                         compressed_length: size_t) -> c_int;
} 
```

In Lean I think attributes are a perfect fit for implementing this system. 

For example if we want to interface with file system APIs: 

``` lean
constant open : path -> options -> IO file_descriptor 

attribute open extern[name = "open", link = "libc", wrapper = "true"]
```

I'm not attached to a particular syntax, but would love to do something like the above example. I think this could be generally useful to others too, Rust has made heavy use the attribute system for marking deprecation, warnings, documentation, metadata and more.

Kha:
The last commit introduces an abstract `attribute` base class together with subclasses for 'basic' attributes, priority attributes and ones that hold a `list<unsigned>`. I haven't touched the existing API yet because I'd first like to have some feedback on this design and especially the implementation. This is the very first time I've designed a class hierarchy in C++.

If the design is acceptable, I would next refactor the current API to make use of the hierarchy so that e.g.
`set_attribute(env, get_dummy_ios(), "unfold", n, LEAN_DEFAULT_PRIORITY, idxs, persistent)`
becomes
`get_unfold_attribute().set(env, get_dummy_ios(), n, idxs, persistent)`
and delegate parameter parsing to a virtual method. At that point, it should be usable for @jroesch's needs, so it would probably make sense to do a first merge.

src/library/attribute_manager.h
```diff
@@ -7,65 +7,149 @@ Author: Leonardo de Moura
...
 
 #ifndef LEAN_DEFAULT_PRIORITY
 #define LEAN_DEFAULT_PRIORITY 1000u
 #endif
 
 namespace lean {
+
+struct attr_data {
+    virtual unsigned hash() const {
+        return 0;
```
dselsam, line 21: You probably want this to be a pure-virtual method: 
`virtual unsigned hash() const = 0`

src/library/attribute_manager.h
```diff
@@ -7,65 +7,149 @@ Author: Leonardo de Moura
...
+        return set_core(env, ios, n, attr_data_ptr(new attr_data), persistent);
+    }
+};
+
+// TODO(sullrich): Let the attribute handle priority parsing itself by introducing a custom
+// syntax (something like [attr:prio]?)
+class prio_attribute : public attribute {
+protected:
+    virtual void write_entry(serializer &, attr_data const &) const override final {}
+    virtual std::unique_ptr<attr_data> read_entry(deserializer &) const override final { return {}; }
```
dselsam: Why not make these pure-virtual?

src/library/attribute_manager.cpp
```diff
@@ -7,121 +7,161 @@ Author: Leonardo de Moura
...
 
-void register_param_attribute(char const * attr, char const * descr, set_param_attribute_proc const & setter,
-                              has_attribute_proc const & tester, get_attribute_param_proc const & getter) {
-    lean_assert(!is_attribute(attr));
-    param_attribute_decl decl;
-    set_core_fields(decl, attr, descr, setter, tester);
-    decl.m_token  = std::string("[") + attr;
-    decl.m_getter = getter;
-    g_param_attrs->push_back(decl);
+class legacy_prio_attribute : public prio_attribute {
```
dselsam: Are you planning to keep this, and to keep the name "legacy"?

src/library/attribute_manager.cpp
```diff
@@ -7,121 +7,161 @@ Author: Leonardo de Moura
...
-    std::string        m_id;
-    std::string        m_descr;
-    std::string        m_token;
-    has_attribute_proc m_tester;
-};
 
-struct default_attribute_decl : public attribute_decl {
-    set_attribute_proc m_setter;
-};
+static name_map<attribute_ptr> * g_attributes;
```
dselsam: Why is this a name? At least currently, you seem to only be storing strings here. Also, you may want to make it a hash table instead of a functional map.

src/library/attribute_manager.h
Kha, line 21: For simplicity, even a `basic/prio_attribute` allocates an `attr_data` right now. Would you suggest changing this so that `attr_record.m_data` is an empty `unique_ptr` for those attributes (in which case I'd probably have to add that information to the serialization)? Or should I just move the `hash` implementation into a new `basic_attr_data` class?

src/library/attribute_manager.h
```diff
@@ -7,65 +7,149 @@ Author: Leonardo de Moura
...
+public:
+    prio_attribute(char const * id, char const * descr) : attribute(id, descr) {}
+    virtual environment set(environment const & env, io_state const & ios, name const & n, unsigned prio,
+                            bool persistent) const;
+};
+
+template<typename Data>
+class typed_attribute : public attribute {
```
dselsam, line 90: It is up to you, but you might consider making all attributes just template-instances of this one, perhaps even without the base class. The `write_entry` and `read_entry` could even be implemented here, and just serialize and de-serialize the data field.

src/library/attribute_manager.h
Kha: Wouldn't that prevent me from instantiating the class?

src/library/attribute_manager.cpp
Kha: If it's still used after the API refactoring, I'll think of a better name.

src/library/attribute_manager.h
Kha, line 90: For `typed_attribute`, the data class is now responsible for serialization, which reduces `unsigned_params_attribute` to a typedef. But `basic_attribute` and `prio_attribute` still aren't `typed_attribute`s because they have different `on_set` signatures.

Kha:
@dselsam There's a single last failing test, `simp3.lean`: `tactic failed, result contains meta-variables`. Could this be due to some new instances in the scope?

leodemoura:
I merged the changes made so far.

PR #1092, opened 2016-07-24
A frontend for most of SMT2

dselsam:
The first commit adds a new macro type for numerals of any of the concrete types `nat`, `int`, and `real`. It also adds an `arith_instance_manager` to bundle partial applications and other instances for arithmetic types, with instances for concrete types hard-coded. Note that the manager has not been fully tested yet except for the concrete types `int` and `real`, which are all that are used by the SMT2 frontend. The rest of the functionality will be used heavily next week when integrating the arithmetic normalizer.

The second commit adds a scanner, parser, and elaborator for most of the SMT2 standard. It supports the following theories: `Core`, `Ints`, `Reals`, `Reals_Ints`, and `ArraysEx`. Out of those, `Core` and the arithmetic ones are integrated tightly with the library (though `ints` and `reals` have not actually been ported to Lean3 yet). The placeholder file `library/smt2.lean` includes constants for everything that is used and is not yet part of Lean3. For arrays, we currently map the three symbols `Array`, `select` and `store` to undefined constants of the appropriate types. Note that it would be problematic to use the new `map` type in `data.map` because of the `option B` return type. One approach is to require `[inhabited]` instances, but then we would need to synthesize them on the fly and it seems unnecessarily complicated.

A note on elaboration: all elaboration is local to a single n-ary application. Although the standard requires all real numerals to have decimal points, some files in the benchmarks do not abide by this, and so we sometimes need to coerce `int` to `real`. I could not find an `Array` file that requires coercing e.g. `Array Int Int` to `Array Real Real` so I did not implement that feature.

Since we do not have an SMT solver yet, `(check-sat)` triggers a tactic that will always fail. Next week when I am working on the arithmetic normalizer, I will make the default tactic "simplify and normalize with everything", which may already be able to prove some of the goals.

Note: to call it, you need to pass the argument `--smt2` or else Lean will treat `.smt2` as an unknown extension.


library/algebra/smt2.lean
```diff
@@ -0,0 +1,47 @@
+import algebra.ordered_field
```
leodemoura: Could you move the file to `library/smt`?

library/algebra/smt2.lean
```diff
@@ -0,0 +1,47 @@
...
+constants (real_has_lt : has_lt real)
+constants (real_has_le : has_le real)
+
+attribute [instance] real_has_zero real_has_one real_has_add real_has_mul real_has_sub real_has_neg real_has_div real_has_le real_has_lt
+
+constants (real.of_int : int → real) (real.to_int : real → int) (real.is_int : real → Prop)
+
+-- Arrays
+-- Note: we don't use map because of the [option] in the return type
+universe variables l₁ l₂
```
leodemoura: It is easy to define the array theory and prove the lemmas used in SMT.
https://github.com/leanprover/lean/blob/c50c2dde685cc0d276a2b66191f47c19c3b30d44/library/smt/array.lean

src/frontends/smt2/parser.cpp
```diff
@@ -0,0 +1,691 @@
...
+
+        bool ok = true;
+        try {
+            parse_commands();
+        } catch (exit_exception const & ex) {
+            return ok;
+        } catch (throwable const & ex) {
+            ok = false;
+            legacy_type_context tc(m_env, m_ios.get_options());
+            auto out = regular(m_env, m_ios, tc);
```
leodemoura: Why are you using the legacy_type_context? We are going to delete it as soon as we have the new elaborator.

src/library/arith_instance_manager.cpp
```diff
@@ -0,0 +1,489 @@
...
+arith_instance_info & get_arith_instance_info_for_int() {
+    return *g_int_instance_info;
+}
+
+arith_instance_info & get_arith_instance_info_for_real() {
+    return *g_real_instance_info;
+}
+
+arith_instance_info & get_arith_instance_info_for(type_context & tctx, expr const & type) {
+    expr_struct_map<arith_instance_info> & cache = get_arith_instance_info_cache_for(tctx);
```
leodemoura: It would be nice if this procedure returns `g_nat_instance_info` when `type` is `nat`. Same for `int` and `real`. I didn't check the code carefully, but it doesn't seen to be the case.

src/library/arith_instance_manager.h
```diff
@@ -0,0 +1,74 @@
...
+*/
+#pragma once
+#include "kernel/expr.h"
+#include "library/type_context.h"
+
+namespace lean {
+
+class arith_instance_info {
+    type_context * m_tctx_ptr;
+
```
leodemoura: This looks wrong. `type_context` is usually a transient object that lives on the stack.
`arith_instance_info` is stored in thread local storage, right?

src/library/arith_instance_manager.cpp
```diff
@@ -0,0 +1,489 @@
...
+namespace lean {
+
+static arith_instance_info * g_nat_instance_info  = nullptr;
+static arith_instance_info * g_int_instance_info  = nullptr;
+static arith_instance_info * g_real_instance_info = nullptr;
+
+class arith_instance_info_cache {
+private:
+    environment m_env;
+    expr_struct_map<arith_instance_info> m_cache;
```
leodemoura: `arith_instance_info` is not a small object. Try `sizeof(arith_instance_info)`. `expr_struct_map` will have to copy these objects when inserting elements, etc.

library/algebra/smt2.lean
leodemoura: The file `library/smt/array.lean` assumes the index type has decidable equality. 
We can avoid this requirement by using `open classical`. Then all types will have decidable equality.
I think this is reasonable, since all SMT users and benchmarks are assuming classical logic anyway.

src/library/arith_instance_manager.cpp
```diff
@@ -0,0 +1,489 @@
...
+}
+
+arith_instance_info & get_arith_instance_info_for_real() {
+    return *g_real_instance_info;
+}
+
+arith_instance_info & get_arith_instance_info_for(type_context & tctx, expr const & type) {
+    expr_struct_map<arith_instance_info> & cache = get_arith_instance_info_cache_for(tctx);
+    auto it = cache.find(type);
+    if (it != cache.end()) {
```
leodemoura: The approach used for caching looks incorrect, and may produce incorrect results.
Consider the following corner case scenario. 
Invoke `get_arith_instance_infor_for(tctx, A)` where the local context in `tctx` is 
`(A : Type) [comm_ring A] ...`
Now, we invoke `get_arith_instance_info_for(tctx, A)` again with the same `A` and same environment, but the rest of the local context is different (e.g., `(A : Type) [field A] ...`). In this scenario, the cache would contain incorrect results.
I realize this is not a common scenario, but it may generate problems in the future.
To fix this problem we have to use an approach similar to the one used in `type_context`.
It is more expensive, since we have to check whether two different local contexts are compatible or not. when we are accessing the cache.

src/library/arith_instance_manager.cpp
dselsam: Good point. I found the magic incantation to emplace the `arith_instance_info` on the map:
https://github.com/dselsam/lean/blob/smt2/src/library/arith_instance_manager.cpp#L484-L490

src/library/arith_instance_manager.cpp
dselsam: Great catch. 

I refactored `type_context::compatible_local_instances` to take a local context as argument, instead of hard-coding it to the local context in its cache:
https://github.com/dselsam/lean/blob/smt2/src/library/type_context.cpp#L2090-L2116
Then I store a local context along with the cache entry:
https://github.com/dselsam/lean/blob/smt2/src/library/arith_instance_manager.cpp#L21-L27
Finally, when retrieving from the cache, I check for compatibility and then update the cached local context:
https://github.com/dselsam/lean/blob/smt2/src/library/arith_instance_manager.cpp#L503-L510

PR #1097, opened 2016-07-29
Combination of last three pull requests

dselsam:
As of this writing, this branch is rebased against lean3. Whether we merge this now or not, I am going to do all my subsequent work on top of these commits.


src/library/CMakeLists.txt
```diff
@@ -14,7 +14,6 @@ add_library(library OBJECT deep_copy.cpp expr_lt.cpp io_state.cpp
   attribute_manager.cpp error_handling.cpp unification_hint.cpp
   local_context.cpp metavar_context.cpp type_context.cpp export_decl.cpp delayed_abstraction.cpp
   fun_info.cpp congr_lemma.cpp defeq_canonizer.cpp scope_pos_info_provider.cpp
-  local_context.cpp metavar_context.cpp type_context.cpp export_decl.cpp lazy_abstraction.cpp
```
dselsam: This was erroneously added while rebasing a previous commit. It is not worth rebasing everything again, but note that the commits between this one and the one that inserted it will not build.

PR #1099, opened 2016-08-03
feat(simplifier/simp_lemmas): take arbitrary list of attributes and c…

dselsam:
…ache.

### DIFF
--- a/library/init/meta/simp_tactic.lean
+++ b/library/init/meta/simp_tactic.lean
@@ -11,17 +11,18 @@ open list nat
 
 meta_constant simp_lemmas : Type₁
 
-/- Create a data-structure containing all lemmas tagged as [simp].
+/- Create a data-structure containing a simp lemma for every constant in the first list of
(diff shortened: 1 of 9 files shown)


dselsam:
Note: this is a weird solution. Right now I am just caching the `simp_lemmas` for each namespace individually, and then when `mk_simp_lemmas` is called, I get the individual `simp_lemmas` from the cache and insert them all into a new `simp_lemmas` structure. Thus even if everything is in the cache, we still need to re-insert the simp-lemmas into a functional map every time.

It would be better to cache the whole `simp_lemmas` structure itself indexed by the list of names (actually the pair, list of simp names and list of congr names). I did not see an easy way to make a `std::unordered_map<pair<list<name>, list<name>>>` though. If the lemma insertion is ever found to be a bottleneck, I will try this approach. On Jeremy's truth table example at least, the percentage of time in `mk_simp_lemmas` went from 15% to 1%.

dselsam:
Actually, I will do this optimization in the morning, and open a new PR.

PR #1100, opened 2016-08-03
feat(simplifier/simp_lemmas): take arbitrary list of attributes and c…

dselsam:
…ache


dselsam:
For now I just optimized for the special case when there is a single simp attribute given. We can revisit this issue later if it is ever a bottleneck. There are many different ways of bringing down the cost.

PR #1106, opened 2016-08-08
refactor(simplifier): many fixes, extensions, and tests

dselsam:
This pull request contains dozens of minor fixes to the simplifier, new support for n-ary associative operators, new macros for encapsulating proofs, and hooks for theory solvers to be plugged in. Note: `eq` is always used by default, and `propext` is assumed throughout. All macros are given high enough trust level that they are always unfolded, except for the propositional theory module for which expansion has not been implemented yet. Reason: the necessary lemmas are only in the library for `iff`, and it didn't seem worth proving the propext versions until we settled on actually using `propext`. It is still fairly easy to revert to `iff`.

I am switching projects today, and I think that even if the simplifier gets extensively refactored again it is still better to merge this PR and start from here.

Issues:
1. I am not sure the point of LEAN_BELIEVER_TRUST_LEVEL. The result of those checks in `unfold_macros.cpp` was that untrusted macros were not expanded before giving them to the kernel, and then the kernel would fail because of the macros. I simplify removed the checks.

Outstanding items:
1. Expanding the prop-simplifier macro.
2. Better canonization of subsingletons (see https://github.com/dselsam/lean/blob/simp_refactor/tests/lean/run/simplifier_canonize_subsingletons.lean#L71-L80).


dselsam:
The use of `propext` is still bugging me, and not just because some things will no longer compute. Users will want to prove `iff` theorems and tag them as `[simp]`, but these would be ignored when `simp` is called with `eq` as it would be by default. The user would need to prove the `propext` version of the theorem as well or instead.

Right now the only impediment to removing `propext` is tracking the ability to flatten an n-ary application, which requires an equivalence relation, an associativity lemma for the operator over that relation, and a `congr_arg` lemma for that relation/op combination. We want `and` and `or` to be flattened.

The fully general solution is to deal with all of using type classes that take or store all of the necessary facts:

```
structure equiv_relation [class] {A : Type} (rel : A → A → Prop) :=
(refl : ∀ x, rel x x)
(symm : ∀ x y, rel x y → rel y x)
(trans : ∀ x y z, rel x y → rel y z → rel x z)

structure is_associative [class] {A : Type} (rel : A -> A -> Prop)
                                                  [equiv_relation rel] 
                                                  (op : A -> A -> A) :=
(congr : forall x y z, rel y z -> rel (op x y) (op x z))
(assoc : forall x y z, rel (op (op x y) z) (op x (op y z)))
```

The only concern here is performance for the common cases. An intermediate solution is to have one `is_associative` for `eq` that only takes `assoc`, and one for `iff` that takes `assoc` and `congr_arg`, that users rarely need to consider.

src/library/unfold_macros.cpp
```diff
@@ -119,8 +119,6 @@ class unfold_untrusted_macros_fn {
 };
 
 static bool contains_untrusted_macro(unsigned trust_lvl, expr const & e) {
-    if (trust_lvl > LEAN_BELIEVER_TRUST_LEVEL)
-        return false;
     return static_cast<bool>(find(e, [&](expr const & e, unsigned) {
```
leodemoura, line 124: This was a performance optimization. 
`contains_untrusted_macro` is invoked when importing files. 
When we are importing hundreds of files, the following `find` consumes a significant amount of time. It was affecting the `flycheck` response time.
The idea of the statement above was: if the trust level is high enough trust all macros.

src/library/unfold_macros.cpp
```diff
@@ -143,7 +141,7 @@ expr unfold_all_macros(environment const & env, expr const & e) {
 }
 
 static bool contains_untrusted_macro(unsigned trust_lvl, declaration const & d) {
-    if (trust_lvl > LEAN_BELIEVER_TRUST_LEVEL || !d.is_trusted())
+    if (!d.is_trusted())
         return false;
```
leodemoura, line 147: Same here.

src/library/unfold_macros.cpp
dselsam, line 124: This version (`(unsigned trust_lvl, expr const & e)`) is only used when defining new definitions. The other one (`(unsigned trust_lvl, declaration const & d)`) is used during importing, and the optimization is fine. I think it is important to remove the check in this one though, because we currently pass the untrusted macros to the kernel without unfolding them, and the kernel throws an error.

leodemoura, line 124: > we currently pass the untrusted macros to the kernel without unfolding them, and the kernel throws an error.

If this is the case, then removing this line is just masking the real problem. 
The kernel object should have been created using the same `trust_lvl` we are checking here. 
So, if the kernel is rejecting the macro, there is a bug somewhere. 
The first possibility that comes to my mind is: a macro was defined with `trust_lvl` greater than  `LEAN_BELIEVER_TRUST_LEVEL`.

PR #1107, opened 2016-08-08
Make `inline` an attribute

Kha:
Quick PR to minimize merge conflicts


src/frontends/lean/parser.cpp
```diff
@@ -2027,18 +2027,20 @@ class lazy_type_context : public abstract_type_context {
 void parser::parse_command() {
     lean_assert(curr() == scanner::token_kind::CommandKeyword);
     m_last_cmd_pos = pos();
-    name const & cmd_name = get_token_info().value();
+    name cmd_name = get_token_info().value();
```
Kha, line 2030: Note: This fixes a bug where the 'unknown command' error below would print the wrong token.

PR #1108, opened 2016-08-10
Create modules and scaffolds for inductives

dselsam:


dselsam:
Note: this PR does not change any functionality. The old `inductive` command still works and is just in a different location.

dselsam:
Re-organizing library.

PR #1109, opened 2016-08-11
Scaffold for new inductive compiler

dselsam:


dselsam:
I don't need to merge regularly while working on the compiler itself, but it would be great for this PR to merge soon so I don't need to keep rebasing the changes to the `CMakeLists` and `init_module`s.

PR #1111, opened 2016-08-16
linja: Add option to hard-code the path to ninja

shlevy:


leodemoura:
We are currently working on a new version (branch `lean3`). It does not depend on `ninja` anymore.

PR #1112, opened 2016-08-16
Add expression to build with the nix package manager

shlevy:
On top of #1111

### DIFF
--- a/bin/linja.in
+++ b/bin/linja.in
@@ -107,7 +107,7 @@ g_phony_targets    = ["all", "clean", "tags", "clear-cache"]
 g_project_filename = ".project"
 g_lean_path        = "USE DEFAULT" # System will search automatically
 g_leantags_path    = "USE DEFAULT" # System will search automatically
-g_ninja_path       = "USE DEFAULT" # System will search automatically
+g_ninja_path       = "@NINJA_PATH@" # USE DEFAULT = System will search automatically
 g_flycheck_header  = "FLYCHECK_BEGIN"
 g_flycheck_footer  = "FLYCHECK_END"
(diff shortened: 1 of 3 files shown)


gebner:
FWIW, we have a [lean package in nixpkgs](https://github.com/NixOS/nixpkgs/blob/master/pkgs/applications/science/logic/lean/default.nix), and you can override it to build current master:

```
nix-build -E 'with import <nixpkgs> {}; lean.override { src = ./.; }'
```

However that doesn't work on the lean3 branch as `src/cmake/Modules/CompBase.cmake` tries to fetch something from github.  In any case, the NINJA_PATH is definitely nicer than the wrapper.

shlevy:
@gebner I prefer building packages without (recent) official releases directly from source checkouts, rather than adding them to nixpkgs and periodically updating them. If this is merged here I'll try to get it working on `lean3` as well :smile:

gebner:
My point was that you can easily use the nixpkgs expression to build a local checkout.  But of course, this only works as long there are no new dependencies or build system changes.

soonhokong:
@gebner,

> However that doesn't work on the lean3 branch as src/cmake/Modules/CompBase.cmake tries to fetch something from github. In any case, the NINJA_PATH is definitely nicer than the wrapper.

There is a workaround. Copy `src/cmake/Modules/cpp14_lang` directory in `master` branch to `lean3` branch. FYI, here is a [link](https://github.com/leanprover/lean/tree/master/src/cmake/Modules/cpp14_lang).

leodemoura:
We are currently working on a new version (branch lean3). It does not depend on ninja anymore.

PR #1116, opened 2016-08-17
Basic user-defined attributes

Kha:
I haven't added a test for `attribute.get_instances` yet, because I didn't find a way to get the current environment outside of a proof.

### DIFF
--- a/src/library/attribute_manager.h
+++ b/src/library/attribute_manager.h
@@ -100,9 +100,9 @@ protected:
         return attr_data_ptr(data);
     }
 public:
-    typed_attribute(char const * id, char const * descr, on_set_proc on_set) : attribute(id, descr),
+    typed_attribute(name const & id, char const * descr, on_set_proc on_set) : attribute(id, descr),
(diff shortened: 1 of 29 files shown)


src/library/attribute_manager.h
```diff
@@ -155,20 +155,31 @@ template class typed_attribute<indices_attribute_data>;
...
-attribute const & get_attribute(name const & attr);
-void get_attributes(buffer<attribute const *> &);
+class user_attribute_ext {
+public:
+    virtual name_map<attribute_ptr> const & get_attributes(environment const & env);
+    virtual void write_entry(serializer &, attr_data const &) {}
+    virtual attr_data_ptr read_entry(deserializer &) {
+        return attr_data_ptr(new attr_data);
+    }
+};
```
Kha, line 165: @leodemoura I made this into a class after I noticed that I needed multiple callbacks

leodemoura, line 165: It makes sense.

PR #1122, opened 2016-08-23
Attribute removal

Kha:
Adds the syntax `[-simp]` for removing attributes.

Existing system attributes that register a callback for setting but not for unsetting cannot be removed. I assume that this is the expected behavior for a few attributes like `instance`.

@leodemoura I haven't disallowed persistent attributes removal so far since it seems to work out of the box.


leodemoura:
@avigad Sebastian already added attribute removal :)

avigad:
I saw! @Kha, when I am back in Pittsburgh, let's work to set up a nice user front-end for using auto, simp, etc. @leodemoura and I have already started discussing the syntax and mechanisms.

PR #1126, opened 2016-09-06
doc(make): change g++-4.9 to g++

swr1bm86:
now if use `g++4.9` as compiler flag on OSX, it will produce error message
that

```
CMake Error at CMakeLists.txt:2 (project):
  The CMAKE_CXX_COMPILER:

    g++-4.9

  is not a full path and was not found in the PATH.

  Tell CMake where to find the compiler by setting
  either the environment variable "CXX" or the CMake cache entry
  CMAKE_CXX_COMPILER to the full path to the compiler, or to the
  compiler name if it is in the PATH.
```

so just use `g++` is cool enough, i can build Lean use this flag.


leodemoura:
merged in the `lean3` branch

PR #1128, opened 2016-09-10
feat(inductive_compiler): support for mutually inductive types

dselsam:


jroesch:
monster PR 👍

library/init/datatypes.lean
```diff
@@ -394,23 +394,23 @@ definition default_has_sizeof_eq (A : Type) (a : A) : @sizeof A (default_has_siz
 rfl
 
 attribute [instance]
-definition nat_has_sizeof : has_sizeof nat :=
+definition nat.has_sizeof : has_sizeof nat :=
```
leodemoura: This is bad, we will get unintended overloads when we open namespaces. One option is to mark it as protected. The other (simpler) option is to revert this change.

dselsam:
This doesn't handle nested inductive types yet, just mutually inductive ones. The mutuals are vastly simpler to compiler. I have a design on paper for the nested compiler, and a big hairball on a different branch that implements about 3/4 of the necessary functionality. It has taken much longer than expected but I think I'll be able to PR the nested next week.

library/init/datatypes.lean
dselsam: Right now I am generating `<ind_name>.has_sizeof` for every inductive type, and I am relying on finding these declarations in the C++ by appending `has_sizeof` to the name of the inductive type in question. I think we should handle this the same way we handle `.cases_on`, and protect to avoid the overloads. I will fix it tomorrow morning.

leodemoura: > I am relying on finding these declarations in the C++ by appending has_sizeof to the name of the inductive type in question

`has_sizeof` is a type class. So, we can use type class resolution to find the `has_sizeof` instance for a type `T`.

> I think we should handle this the same way we handle .cases_on, and protect to avoid the overloads.

`.cases_on` is different. First, there is no top level `cases_on` declaration. Second, `cases_on` is not a type class.

In general, the names of type class instances do not matter. For example, we don't need any naming convention for handling `has_add` instances in the automation.

If you need a naming convention, please use something like `.has_sizeof_inst`. 
Anyway, it looks weird that the code only works if the instance for `nat` is called `nat.has_size_of`.

src/library/util.cpp
```diff
@@ -200,6 +200,16 @@ bool is_inductive_predicate(environment const & env, name const & n) {
     return is_zero(get_datatype_level(env.get(n).get_type()));
 }
 
+bool can_elim_to_type(environment const & env, name const & n) {
+    if (!is_standard(env))
+        return false;
```
leodemoura: It should be `return true` for `!is_standard(env)`, and this statement should be after line 207.

PR #1129, opened 2016-09-12
Pure function caching for user-defined attributes

Kha:
I wasn't actually planning to implement this myself, but was curious to at least take a look at the current caching infrastructure. I was pleasantly surprised by its simplicity, making this a fairly quick and easy addition.

### DIFF
--- a/library/init/meta/attribute.lean
+++ b/library/init/meta/attribute.lean
@@ -10,8 +10,16 @@ meta_constant attribute.get_instances : name → tactic (list name)
 meta_constant attribute.fingerprint : name → tactic nat
 
 structure user_attribute :=
+(name : name)
 (descr : string)
 
 /- Registers a new user-defined attribute. The argument must be the name of a definition of type
-   `user_attribute`. -/
+   `user_attribute` or a sub-structure. -/
 meta_constant attribute.register : name → command
+
+structure caching_user_attribute extends user_attribute :=
+(Cache : Type)
+(cache : list declaration → Cache)
+
+meta_constant caching_user_attribute.get_cache :
+  Π(attr : caching_user_attribute), tactic (caching_user_attribute.Cache attr)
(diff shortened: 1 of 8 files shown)


library/init/meta/attribute.lean
```diff
@@ -10,8 +10,16 @@ meta_constant attribute.get_instances : name → tactic (list name)
 meta_constant attribute.fingerprint : name → tactic nat
 
 structure user_attribute :=
+(name : name)
```
Kha, line 13: The attribute name now comes directly from the structure instead of the definition's name so that `caching_user_attribute.get_cache` can access it.

src/library/tactic/user_attribute.cpp
```diff
@@ -117,6 +160,16 @@ vm_obj attribute_fingerprint(vm_obj const & vm_n, vm_obj const & vm_s) {
     return mk_tactic_success(mk_vm_nat(h), s);
 }
 
+vm_obj caching_user_attribute_get_cache(vm_obj const & vm_attr, vm_obj const & vm_s) {
+    auto const & s = to_tactic_state(vm_s);
+    auto const & n = to_name(cfield(vm_attr, 0));
+    auto const & cache_handler = cfield(vm_attr, 2);
```
Kha, line 166: Finding out the field index was trial and error, I'm not quite clear yet at what point the `Cache` field / the dependent pair vanishes.

leodemoura: The code generator erases all fields that are:
1) proofs (since they are irrelevant)
2) types (also computationally irrelevant).
The following function returns this information:
https://github.com/leanprover/lean/blob/lean3/src/library/util.h#L87

Kha: I see, thanks!

PR #1132, opened 2016-09-19
Make local_context::get_unused_name O(log(n)) and use it in the intro tactic

gebner:
Right now, calling the `intro` tactic on two binders with the same name `a` results in two local constants `a` and `a` with the same (user-facing) name.  This is unsatisfactory, as it is now not only unclear what `a` resolves to (it refers to the one introduced last), it also makes it impossible to refer to one of the `a`s at all.

This PR resolves this by using the existing `local_context::get_unused_name` function to obtain a name that is not yet used in the current goal, i.e. we now get the hypotheses `a` and `a_1`.  As discussed on slack, I have also implemented an asymptotically faster version of `get_unused_name`, it should now be O(log(n)) instead of O(n^2).

The function `get_unused_name` now has the following specification: given a name `foo` it returns the first name out of the following list that is not yet used as a user-facing name in a hypothesis: `foo`, `foo_1`, `foo_2`, ...

### DIFF
--- a/library/init/meta/tactic.lean
+++ b/library/init/meta/tactic.lean
@@ -231,6 +231,7 @@ meta_constant infer_type    : expr → tactic expr
 meta_constant get_local     : name → tactic expr
 /- Return the hypothesis in the main goal. Fail if tactic_state does not have any goal left. -/
 meta_constant local_context : tactic (list expr)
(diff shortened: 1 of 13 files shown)


leodemoura:
Thanks

PR #1137, opened 2016-09-28
type_context: add more information to exceptions

gebner:


leodemoura:
I've just submitted https://github.com/leanprover/lean/commit/4a5d881c50a8652138cd34bf0a666acfddae9ccb
The main differences are:
- It is the `formatter` object 
- It is uses thunks. We had performance problems in the past when terms are too big. It was particularly bad when the error message is not even used (e.g., `tac1 <|> tac2`, where `tac1` fails and `tac2` succeeds).

leodemoura:
BTW, @dselsam added a `pp` option that is useful for debugging tactics.

```
set_option pp.locals_full_names true
```

It forces the pretty printer to display the unique id associated with a local constant.

PR #1139, opened 2016-09-29
Emacs commits from master

Kha:
Commits from `git log lean3..master src/emacs` that didn't look out of date


leodemoura:
One issue: Many keybinds have been deleted.
Example: auto completion is gone.

Kha:
Yes, I wasn't sure whether to delete them. But nobody should be too surprised about missing features right now, so I left it that way for now.

Kha:
I've amended the last commit, commenting out all unimplemented keybindings.

PR #1150, opened 2016-10-11
feat(library/tactic): add back notation for `by assumption`

Kha:
I didn't realize that with a less special syntax, we can just implement this in Lean... at least, I think we can. Error messages at least are already nicer than in lean2: compare the test output with

```
invalid tactic, term
  A
has type
  Type
but is expected to have type
  A
proof state:
A : Type
⊢ A
```


avigad:
That works beautifully, and I like the notation. Are we happy with the keystrokes, \flq and \frq?

\la and \ra (for left-angle and right-angle) are a little shorter and easier to remember.

Kha:
I would be fine with `\la`. I've already mapped `«` to `\<<` locally (for my escaped names patch), but I'm sure we'll want to keep `\<` for anonymous constructors. So for consistency's sake, perhaps something like `\f<` and `\f<<`? Assuming that 'f' in the old bindings actually stands for 'French', at least to me that would provide a nice mnemonic to distinguish them from `⟨` and `⟪`.

avigad:
I like that proposal. At least, in my old age, I think I can remember those.

PR #1155, opened 2016-10-14
[WIP, RFC] Compilation server

gebner:
This PR should bring support for incremental compilation of lean files within emacs, i.e. if you edit a proof at the end of a file, only the last proof will be redone--everything above it is cached.

Status: incremental compilation works, next step will be auto-completion.

The implementation works roughly in a similar way to the lean-server in lean 2, with a few differences:
- Instead of transferring only changed lines, we always transfer the whole buffer.
- Flycheck errors are done via lean-server as well.
- JSON instead of a custom text protocol.
- One server per buffer (see below).

Further changes:
- Error messages are now stored in the parser as `message` objects.  This allows us to 1) easily resend them to the client, and 2) also relieves us of parsing headaches (linja silently broke some flycheck errors).
- Lean now outputs a warning if the olean file of any imported file is out-of-date compared to the lean file.

To do:
- auto-completion via company-mode (@Kha is currently working on this)
- go-to-definition (I guess we need to extend `declaration` objects and olean files to include file/line information)
### Line protocol

Communication with `lean --server` happens via stdin and stdout.  One JSON object on a single line per request, and one JSON object on a single line per response.  A typical exchange looks as follows:

```
< {"content": "run_command (tactic.trace \"hello, world\")\n", "file_name": "demo.lean", "command": "sync"}
> {"message":"no file changes","response":"ok"}
< {"command": "check"}
> {"is_ok":true,"messages":[{"caption":"trace output","file_name":"demo.lean","pos_col":0,"pos_line":1,"severity":"information","text":"hello, world\n"}],"response":"ok"}
```

For JSON support in C++, I added a small MIT-licensed library consisting of a single header file (that's the official release).  For now, I put it into `src/utils`.

I picked JSON for the protocol format mainly because of readily available, reliable, and tested libraries for it.  Possible alternatives would be s-expressions or a plain text format.  S-Expressions are unfortunately not standardized, and almost no premade parsers/formatters exist, requiring us to roll our own solution (and making sure that it's consistent across lean and all possible clients).  A custom text format as in lean 2 (or flycheck in lean 3) has similar issues: again we would need to implement and test a parser for each possible front-end--and there have been bugs even in the heavily-used linja tool.
### Project handling

I think the biggest issue with this PR is the handling of projects.  Before, each flycheck invocation called linja to rebuild a temporary copy of the current buffer _and_ its dependencies.  So if you have two lean files and modify one that is the dependency of the other, it will be recompiled (and saved to olean) as soon as you 1) save it and 2) edit the other one.

Right now the situation in this PR is the following: on each flycheck invocation, we only recompile the current file.  Should any olean dependencies have changed, we use the new olean files and recompile from the beginning.  If however only the lean file of the dependency has changed, then you only get a warning and need to linja yourself.

I have added the `lean-run-linja` command to run linja from within emacs.  Having a "compile the whole project" function seems to be useful regardless of the approach we take here, and it's very nice: we even hyperlink the errors in the linja output.

I'm a bit uncertain as how to deal with this issue.  From my point of view, the following would be nice to have:
- When I open two files in emacs, and I change something in one file, then the other should be rechecked if it has the first one as a dependency.  Probably even if I don't save the first one.  Across project boundaries as well?
- If a file has errors, I should still be able to use the correct parts from other files.
- If I depend on a lean port of the (let's say) four color theorem, then I don't wait for it to compile.
- There should be a way to have projects outside of the lean repo which depend on other projects.  Right now I don't believe this is easily possible, as linja will only search in the standard library in addition to the current directory.  Any solution we pick for the compile server should be compatible with this situation.

I'm mentioning these points as we almost have them in lean: parallel compilation/delayed proof checking is implemented, even though currently disabled AFAICT.  I just don't want to spend too much effort on a solution that will only ever work with the current 81+35 files, that we can completely check in under a minute.

Possible approaches:
- Require semi-manual linja calls for now.
- Keep the current behavior of the flycheck-mode, i.e. call linja on every check.  We could either make linja wrap the lean server (emulating the current implementation where linja calls lean), or call linja from within emacs.
- Move linja into lean, i.e. make lean compile the dependencies by itself.  With this approach, we could leverage lean's existing parallel proof checking capabilities, providing a similar user experience to Isabelle/PIDE: you could begin to work on a new project while the proof-checking happens concurrently in the background.

Sebastian suggested the last one, and I'm starting to like it as well.


leodemoura:
@gebner @Kha This is awesome!
I will merge it right now.

> Sebastian suggested the last one, and I'm starting to like it as well.

Me too. 

BTW, another advantage of getting rid of `linja` is that it depends on `ninja` and many users had problems with this dependency.

leodemoura:
Question: what does it happen if the lean server crashes? Does emacs automatically restart it?

gebner:
> Question: what does it happen if the lean server crashes? Does emacs automatically restart it?

Every time we send a request to the server, we first make sure that it is running.  If it crashed, it will be restarted.

PR #1156, opened 2016-10-14
Bring back basic completion

Kha:
This is mostly just reviving the old code and adjusting it to the new server API. Currently, only completing declarations is supported. I'd like to extend that to a unified interface for many different kinds of completions in the future.

As an experiment, I've also enabled _idle completion_ which will asynchronously kick in (after typing at least 5 characters) without bringing your typing to a halt; there can be some flickering, however. When you press `tab`, you're back to the old, synchronous behavior.

### DIFF
--- a/src/shell/server.cpp
+++ b/src/shell/server.cpp
@@ -276,8 +276,8 @@ json server::handle_complete(json const & req) {
         std::vector<pair<std::string, name>> selected;
         bitap_fuzzy_search matcher(pattern, max_errors);
         env.for_each_declaration([&](declaration const & d) {
-            /*if (is_projection(env, d.get_name()))
-                return;*/
+            if (is_projection(env, d.get_name()))
+                return;
             if (auto it = exact_prefix_match(env, pattern, d)) {
                 exact_matches.emplace_back(*it, d.get_name());
             } else {
--- a/src/shell/server.cpp
+++ b/src/shell/server.cpp
@@ -2,15 +2,35 @@
 Copyright (c) 2016 Microsoft Corporation. All rights reserved.
 Released under Apache 2.0 license as described in the file LICENSE.
 
-Author: Gabriel Ebner
+Authors: Gabriel Ebner, Leonardo de Moura, Sebastian Ullrich
 */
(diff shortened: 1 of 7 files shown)


src/shell/server.cpp
```diff
@@ -146,4 +168,172 @@ json server::handle_check(json const &) {
...
+        environment const & env = snap->m_env;
+        options const & opts = snap->m_options;
+        unsigned max_results = get_auto_completion_max_results(opts);
+        unsigned max_errors = get_fuzzy_match_max_errors(pattern.size());
+        std::vector<pair<name, name>> exact_matches;
+        std::vector<pair<std::string, name>> selected;
+        bitap_fuzzy_search matcher(pattern, max_errors);
+        env.for_each_declaration([&](declaration const & d) {
+            /*if (is_projection(env, d.get_name()))
+                return;*/
```
Kha: @leodemoura Was there a particular reason for excluding projections? I've often missed them when using autocompletion in lean2.

leodemoura: @kha I agree that blocking all projections is a bad idea. This hack was motivated by the algebraic hierarchy. For example, `add_assoc` is a projection for many algebraic structures, and we have the top-level lemma `add_assoc`. When trying to use auto-completion, we would get the `add_assoc` for all algebraic structures in Lean. It was a very long list.

Kha: I see, I suspected something like that. It would be nice to only hide inherited projections (that haven't been renamed), but I don't think we expose that information currently. I'll revert the change for now.

leodemoura: > It would be nice to only hide inherited projections

I'm not sure it would cover all use-cases. Suppose we have the following two structures:

```
structure A := (x : nat)
structure B extends A := (y : nat)
```

They are not classes. If we have `b : B`, we probably want to use `B.x` instead of `A.x`. To use `A.x` we will first have to coerce `b` into `A`.

Kha: Ok, so it should probably be an explicit annotation like `private` (whatever that would exactly mean).

PR #1158, opened 2016-10-15
fix(shell/server): last snapshot may not be valid any more

Kha:
Also avoid copying the vector, but that probably doesn't change much.

### DIFF
--- a/src/shell/server.cpp
+++ b/src/shell/server.cpp
@@ -137,29 +137,27 @@ json server::handle_check(json const &) {
         m_only_checked_until = optional<pos_info>(0, 0);
 
     if (m_only_checked_until) {
-        snapshot_vector new_snapshots;
-        for (snapshot const & snap : m_snapshots) {
-            if (snap.m_pos.first < m_only_checked_until->first) {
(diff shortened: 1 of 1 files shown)


gebner:
@Kha Do you remember any examples where this mattered?  I'd like to add them as test cases.

Kha:
@gebner The end-of-file snapshot would be one example when appending to a file. I think if we are careful about that one and otherwise save snapshots only at the beginning of command keywords, we can revert this change (i.e. go back to comparing line numbers).

gebner:
Yeah, the final snapshot gave me some trouble.  Turns out that `get_first_diff_pos` was pretty broken on files with newlines at the end: for `inductive foo` and `inductive foo\n`, it returned that they differ on the second line...  I then tried to reuse the final snapshot, which was at the end of the first line.

I have fixed this now: https://github.com/gebner/lean/commit/482a5fd69151059c8b2bcde4f6311e88d807ed56

But thanks for reminding me to recheck the cases where we save a snapshot that is not at a command keyword.  This happens for example with doc comments, which behave somewhat like commands.

PR #1161, opened 2016-10-17
[WIP] try to improve elab error message

dselsam:
``` Lean
inductive Foo : Type → Type*
| fixed : bool -> Foo bool
| flex  : Pi (X : Type), X -> Foo X

set_option pp.all true

def rig : Pi (X : Type), decidable_eq X → Foo X → X
| .bool _ (Foo.fixed b) := b
| .X    _ (Foo.flex X x) := x
```

was giving the following annoying error message:

```
/home/dselsam/projects/lean/gdb/gdb.lean:21:11: error: type mismatch at application
  Foo.flex X
term
  X
has type
  Type
but is expected to have type
  Type
```

This (not-polished) commit changes it to the following:

```
/home/dselsam/projects/lean/gdb/gdb.lean:21:11: error: type mismatch at application
  Foo.flex X
unable to unify actual argument
  X : Type
with argument placeholder
  ?m_1 : Type
```

Note: I have yet to figure out why the example doesn't work or whether there is a workaround.

### DIFF
--- a/src/frontends/lean/elaborator.cpp
+++ b/src/frontends/lean/elaborator.cpp
@@ -707,6 +707,18 @@ format elaborator::mk_app_type_mismatch_error(expr const & t, expr const & arg, 
     return msg;
 }
 
+format elaborator::mk_app_arg_mismatch_error(expr const & t, expr const & arg, expr const & arg_type,
+                                             expr const & expected_mvar, expr const & expected_type) {
+    auto pp_fn = mk_pp_ctx();
+    format msg = format("type mismatch at application");
+    msg += pp_indent(pp_fn, t);
+    msg += line() + format("unable to unify actual argument");
+    msg += pp_indent(pp_fn, arg) + space() + colon() + space() + pp_fn(arg_type);
+    msg += line() + format("with argument placeholder");
+    msg += pp_indent(pp_fn, expected_mvar) + space() + colon() + space() + pp_fn(expected_type);
+    return msg;
+}
+
 format elaborator::mk_too_many_args_error(expr const & fn_type) {
     auto pp_fn = mk_pp_ctx();
     return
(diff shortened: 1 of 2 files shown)


dselsam:
I found the issue and a workaround.

Recall the problem:

``` Lean
inductive Foo : Type → Type*
| flex  : Pi (X : Type), X -> Foo X

set_option pp.all true

def rig : Pi (X : Type), decidable_eq X → Foo X → X
| .X    _ (Foo.flex X x) := x
/-
error: type mismatch at application
  Foo.flex X
unable to unify actual argument
  X : Type
with argument placeholder
  ?m_1 : Type
-/
```

The metavariable `?m_type2` for the _type_ of the `_` placeholder gets assigned to `decidable_eq ?m_1`, and in doing so, the context of `?m_1` is restricted, since `?m_type2` does not have the `X` pattern local in its local context (since it has not been seen yet proceeding from left to right).

The workaround is to put the `decidable_eq X` argument _after_ the `Foo X` argument. 

I think this behavior will confuse other users, but unfortunately I do not know the equation compiler well enough yet to assess if this would be easy to address. I'll keep looking into it.

leodemoura:
>  think this behavior will confuse other users, but unfortunately I do not know the equation compiler well enough yet to assess if this would be easy to address. I'll keep looking into it.

This problem is not related with the equation compiler.
There are two problems in the example above. One is minor: incorrect error message being displayed. The suggest fix tries to address this one. 
The real problem is how to communicate unification errors to users.
The example above makes it clear that sometimes is hard to understand why the unification problem `t =?= s` failed IF we just print `t` and `s`. In the example `?m_1 =?= X`, there is one crucial piece of information missing: the context for `?m_1`.
One possibility is to display the context of each meta-variable occurring in `t =?= s`. However, most of the time this information is irrelevant. We can try to minimize the amount of information by tracking which information might be relevant when processing `is_def_eq`.

leodemoura:
When I deleted the `lean3` branch, it also closed this pull request :(

PR #1164, opened 2016-10-19
feat(frontends/lean): support binary, octal and hex literals

joehendrix:
Scanner will interpret numeric literals prefixed with `0b`, `0o`, and `0x` as binary, octal, and hex decimal values respectively.

The prefix character (`b`, `o`, `x`) may be upper or lower case.


leodemoura:
Thanks!

Issue #1169, opened 2016-10-22
lean-server: usability of "show goal"


gebner:
Consider the following incomplete proof:

``` lean
lemma add_comm' (n m : nat) : n + m = m + n :=
begin induction n end
```

With the current `show-goal-at-pos`, you can only see the tactic state before the induction tactic.  There seems no to be no point where we show the (failed) tactic state after the induction tactic.

In addition, we should probably clear the `*lean-info*` buffer if we cannot find a tactic state at a given point.

leodemoura:
> In addition, we should probably clear the _lean-info_ buffer if we cannot find a tactic state at a given point.

The commit above doesn't address this one.
Any suggestions?
Should we have another exception (e.g., `show_goal_no_found_exception`)?

gebner:
Ah, that's something purely on the emacs side.  We do get the information that there is no goal at that point, but tight now we just show an error message at the bottom of the screen.  However the last tactic state still remains visible in the `*lean-info*` buffer.

PR #1170, opened 2016-10-26
Emacs info buffer

gebner:
This fixes the rest of issue #1169.  Now when you press C-c C-g at a point where there is no goal, then the _lean-info_ buffer is updated as well, preventing confusion.

I've also updated the next-error-mode to reuse the same buffer, and added a convenient `with-output-to-lean-info` macro.


gebner:
~~Please don't merge yet, we just noticed a serious performance regression in the next-error-mode.~~

gebner:
The performance regression in next-error-mode should be fixed now.

PR #1178, opened 2016-11-03
[prototype] Parallel compilation

gebner:
Unfortunately, this is far from being production-ready, even though I've already completely rewritten it once.  I'm putting the current version here to solicit feedback, ideas, and general comments.

What should it do?  Put simply, improve the user experience when interactively working with large formalizations.  Kind of like Isabelle/PIDE.  More concretely, I have been focusing on the following three goals:
 - *On-demand elaboration:*  If you modify the "first" definition in the formalization and want to continue working on the "last" theorem, you would currently have to elaborate all the proofs in the middle.  This incurs a significant delay as elaboration of proofs is costly.  Yet we almost never need the actual content of the proofs, so it is a reasonable idea to just skip the proofs in the middle completely.  This elaboration cost will only increase as we integrate more automation in Lean.  For some operations however, notably `print`, `print axioms`, and `reveal`, we still need the proofs: here it would be preferable to only elaborate the proofs that are actually necessary.
 - *Incremental error messages:*  If you have a file with 100 theorems, each of which takes just one second to prove, then even if you elaborate the proofs in parallel, you would still have to wait almost a minute to see the first error message (numbers based on my laptop).  This is complaint I've heard from @Kha many times, and it bugs me as well when working on the `automation/super/examples.lean` file.  It would be nice if error messages (from elaboration) were visible as soon as possible (or at least not delayed until the end of the file.)
 - *Dependency tracking and recompilation from within Lean:*  Right now these functions are performed by linja.  For the on-demand elaboration to work well, I think lean-server will need to handle multiple files at the same time (right now there is one lean-server process per emacs buffer), and also handle dependencies between these files as well as recompilation.  If we have this functionality in lean itself, we can probably drop the linja wrapper and the ninja dependency.

This prototype achieves these three goals.  However the implementation is fragile and incomplete, and the design is awkward at best.

How does the implementation work?  Let me split the description into two parts: first, I'll detail the changes to the compilation process, and then the implementation of the parallelization.

The basic technique uses *delayed theorems*, just as in Lean 2.  When declaring a theorem, we do not put the theorem with the elaborated proof into the environment, but instead put an axiom there.  
When we later produce the elaborated proof, we replace the axiom by the freshly proven theorem.  In Lean 2, at the end of a file we would wait for all delayed theorems to finish proving and then export the olean files with the elaborated proofs.

In this prototype, we process multiple files in the same lean process.  When we process an import command at the beginning of a file, we only wait for the dependent files to finish *parsing*, the proofs in the dependent files do not have to be elaborated yet.  We import the delayed theorems in the dependent files as axioms (by exporting the environments with the unreplaced axioms to an in-memory olean file, and then replaying it in the current environment).  The axioms are only replaced when processing a `reveal` command (not implemented yet), or when writing the final olean file.  Delayed theorems for dependencies do not have to elaborated at all.  So far, this first part works pretty well: we can compile the whole standard library using this approach.

Many of the tasks that we want to perform in parallel have dependencies that are unknown before starting the task.  When parsing a file, we do not know what other files it depends on (ok, this is something we could figure out beforehand), but we really do not know what proofs it depends on.  Even more so, `print axioms` needs to recursively go through the references delayed theorems and gather their proofs (not implemented yet), here we really do not know what dependencies we have.
Hence I've decided to implement a task-based abstraction for delayed, asynchronous values.  (Kind of like `std::async`, but I don't think you can hook them up to a thread pool--probably for a good reason, see note on thread-local variables below.)  There is a global task queue that manages worker threads to process tasks asynchronously.  These `task<T>` values (see `task_queue.h` file) can be synchronously joined, or passed to the task queue for async execution.  Dependencies are expressed by joining the dependent task inside another task.  As an optimization (to reduce waiting and the number of running threads), when we join a not-yet-executing task, we execute it on the current thread.
The actual file loading, dependency handling, and invalidation is handled by the `module_mgr`, which provides a `get_module` function that returns a task for the latest version of the required module.  Updates from the editor are provided via a `module_vfs` object that provides the content of the editor buffer, on each update the `module_mgr` needs to be notified to invalidate the reverse dependencies.

Immediate problems with the implementation:
 - Modules are identified by their absolute file names (and the dependent modules in the olean files are absolute file names as well).
 - Delayed theorems do not work with implicit universe parameters.
 - Missing `reveal`, `print`, `print axioms` functionality.
 - Export-to-text, C API, and SMT parser are commented out.
 - Incremental error reporting in emacs is pretty much just a proof-of-concept.

General points in no particular order:
 - *Combining environments:*  One central operation here consists of combining two or more environments, when importing the dependent modules.  Right now, the only way to do this is by serializing the (newly added) elements in the dependent module to olean, and then deserializing and importing the olean to the current environment.  However, deserializing is *slow*.  On the standard library, we spend 40% of the time during import (if I read the `perf` output correctly), 10% alone on just `deserialize_expr`.
 - *Build time of standard library:*  On a related note, the standard library doesn't benefit at all from this kind of parallelization at the moment: it takes about 5s real time (10s user time) to compile on my laptop--no matter whether using linja or the prototype.  This is probably to be expected since it mostly consists of definitions at the moment.
 - *Thread-local variables:*  Parallelizing lean is difficult as there are many thread-local variables that you need to replicate exactly.  For theorem elaboration, you need to set at least `scoped_expr_caching`, `scope_global_ios`, `scope_trace_env`, `scope_pos_info_provider`, and `scope_trace_as_messages`--and I'm not sure these are all.  At first I thought it would be a good idea to reuse threads--i.e. if you parse a file and require the proof of a theorem, elaborate it right on the same thread.  But this makes the value of the thread-local variables even more unpredictable.
 - *Identity of modules:*  When handling multiple files, we do not want to have multiple different copies of the same file.  That is, we should not handle three different files `data/nat/basic.lean`, `basic.lean`, and `/home/gebner/lean/library/data/nat/basic.lean`--this should be just one file, as far as Lean is concerned.  Right now, parsing and importing get file names.  The command `import .core` gets the file name `data/nat/basic.lean` and just drops the last path component and adds `core.lean`.  I think that we should use module names instead, i.e. use `data.nat.basic` and relative imports are then defined in terms of the name components.  The dependency section in the olean file would only include the fully-qualified module names.  This would make it easy to recognize identical modules.
 - *Incremental compilation:*  There is no support for caching previous elaboration results, all reverse dependencies are always invalidated on every change.  My first idea was to skip invalidation if the environment did not change (i.e. if you only modified a proof), but I could not figure out how to compare environments for equality.
 - *Memory consumption:*  In the common case, all source files, environments for each delayed theorem, intermediate olean files, are all stored in memory.  This is not a problem for the standard library, and I don't think it is a problem for most formalization projects either.  However it might be problematic for software verification, I remember some talk about huge verification conditions--can anybody elaborate?


leodemoura:
Thanks for the update. It looks very promising.
I will write an appropriate response tomorrow morning.

leodemoura:
> It would be nice if error messages (from elaboration) were visible as soon as possible (or at least not delayed until the end of the file.)

Does "elaboration" here mean "elaborator - tactic execution"?
If yes, we can perform a first pass where every tactic that produces a proof is skipped and replaced with a `sorry`. So, we can "quickly" check for parsing errors, typing errors, type class resolution problems, etc.

leodemoura:
> Modules are identified by their absolute file names (and the dependent modules in the olean files are absolute file names as well).

Why is this a problem?

leodemoura:
> There is no support for caching previous elaboration results, all reverse dependencies are always invalidated on every change. 

In Lean2, we had the following cache: https://github.com/leanprover/lean/blob/master/src/library/definition_cache.h

It is essentially a mapping from pre-terms (expr before elaboration) to term (fully elaborated expression).
We would also store a fingerprint for the configuration options and attributes. 
So, if we add a `set_option` or `attribute` command in the beginning of the file the whole cache would be invalided. We can add an option that ignores the fingerprint.

leodemoura:
> My first idea was to skip invalidation if the environment did not change (i.e. if you only modified a proof), but I could not figure out how to compare environments for equality.

We can add an equality env1 == env2 that holds iff the environments contain the same declarations. This is the easy part. We would also have to extend environment_extension with an equality predicate.

leodemoura:
> Right now, the only way to do this is by serializing the (newly added) elements in the dependent module to olean, and then deserializing and importing the olean to the current environment. However, deserializing is slow. 

I can investigate and try to optimize the deserialization code.

BTW, in Lean2, I tried the following experiment: deserialization on demand. 
When importing a file, we would not deserialize the "values" associated with definitions and theorems. The value would only be deserialized when the method `declaration::get_value` is invoked for the first time. Most values are never accessed. 
This approach has a drawback, the .olean files are bigger. For example, if two declarations contain the same sub-term `t`, then `t` is serialized twice in the "deserialization on demand" approach. 
The pros are: it is faster to import .olean files, and we save a considerable amount of memory. I don't remember the exact numbers but we reduced memory consumption by half.

leodemoura:
> On a related note, the standard library doesn't benefit at all from this kind of parallelization at the moment

In Lean2, only very complex files benefited from parallelization. The speedups were very modest (at most 3x faster) when using a machine with 16 cores. The main problem is the synchronization overhead. Moreover, we only got a descent speedup when using `tcmalloc`. 

I made several experiments, one bottleneck seems to be the reference counters. They are `atomic<unsigned>`. We don't use mutexes to update them, but `atomic_fetch_add_explicit`
and `atomic_fetch_sub_explicit`. However, there is still a huge performance penalty.

One alternative design is to remove this synchronization from `expr::m_rc`, and copy expressions when moving from one thread to another. We would have to copy "on demand" since different thread share the same environment, and we don't want to copy the whole environment.

leodemoura:
> if you parse a file and require the proof of a theorem, elaborate it right on the same thread. But this makes the value of the thread-local variables even more unpredictable.

It is not clear to me what the problem is.

leodemoura:
> However it might be problematic for software verification, I remember some talk about huge verification conditions--can anybody elaborate?

The huge VCs come from a different application scenario (lean-as-a-backend) where lean is not being driven by a human, but by a different tool. In the lean-as-a-backend scenario we don't need incremental compilation. So, we just need to be able to execute Lean in "batch" mode as we do with SMT and FO provers. Here is an example of VC created by F*: https://gist.github.com/leodemoura/9e282a2676c8ca8cc8ec6ae49489fccf

That being said, I'm still concerned about memory consumption even when Lean is driven by a human. For example, more than once, I had to use `ninja -j 1` when compiling code on machines with small amount of memory (e.g., 4Gb). The issue is that g++ consumes a lot of memory sometimes. I suspect users will have similar experience when developing complex projects using Lean.

gebner:
> > It would be nice if error messages (from elaboration) were visible as soon as possible (or at least not delayed until the end of the file.)
> 
> Does "elaboration" here mean "elaborator - tactic execution"?

Yes, this is what I meant.

> If yes, we can perform a first pass where every tactic that produces a proof is skipped and replaced with a sorry. So, we can "quickly" check for parsing errors, typing errors, type class resolution problems, etc.

This would only solve part of the issue however, you'd still have to wait for the errors produced during elaboration.  If you have a file with 50 lemmas, each of which takes a second to prove, you do not want to wait half a minute to see the errors in the one you are currently editing.  Ideally you'd see them already after the one second it takes for elaboration. (You can work around this issue by writing `exit` after the lemma.  But then you still need to wait a few seconds for the exit to take effect...)  In any case, this is not too difficult if you elaborate the proofs in parallel, you just send back the errors that are ready, and then update them as new ones come in (implemented).

leodemoura:
> Ideally you'd see them already after the one second it takes for elaboration. (You can work around this issue by writing exit after the lemma. 

You are considering the scenario where we are editing a theorem in the middle of the file, and we add an `exit` after the theorem to avoid processing the rest of the file, right?

BTW, the error-on-demand is a server protocol issue, right? 
The current protocol sends a single big batch of error messages back to Emacs. 
You want to break it in small pieces as they become available. Am I correct?

gebner:
> > Modules are identified by their absolute file names (and the dependent modules in the olean files are absolute file names as well).
> 
> Why is this a problem?

Simply because the olean files now contain strings like `/home/gebner/lean/library/init/nat.lean`, which makes it impossible to move them somewhere else.

I guess we could use the absolute file names to identify modules during parallel compilation (by normalizing them via realpath), and put the import specification (i.e. pair<name, optional<unsigned>>) into the olean file.  But I'd really like to use the same identifier for both purposes, so I'd change module names to be names like `init.nat` everywhere.

gebner:
> BTW, the error-on-demand is a server protocol issue, right? 

One part is a server protocol issue, yes.  The other part is structuring the server in such a way that you can actually obtain the error messages incrementally.

> The current protocol sends a single big batch of error messages back to Emacs. 
> You want to break it in small pieces as they become available. Am I correct?

Actually, the prototype just transmits all available messages every time a request comes in.  The corresponding emacs code then just reschedules a recheck if the messages are incomplete:

``` elisp
       (when (eq incomplete t)
         (setq errors (cons
                       (flycheck-error-new-at 1 1 'warning
                                              "lean error checking still in progress"
                                              :checker checker :buffer buffer)
                       errors))
         (run-at-time "500 milliseconds" nil
                      (lambda () (flycheck-buffer))))
```

leodemoura:
> I guess we could use the absolute file names to identify modules during parallel compilation (by normalizing them via realpath), and put the import specification (i.e. pair>) into the olean file.

I'm not sure I understood the issue. In the current module.cpp, we use this approach. The .olean files contain module_name objects, but we use absolute file names to identify modules in  the import procedure: https://github.com/leanprover/lean/blob/lean3/src/library/module.cpp#L415-L418
This approach seems to work well.

gebner:
> > There is no support for caching previous elaboration results, all reverse dependencies are always invalidated on every change.
> 
> In Lean2, we had the following cache: https://github.com/leanprover/lean/blob/master/src/library/definition_cache.h
> It is essentially a mapping from pre-terms (expr before elaboration) to term (fully elaborated expression).
> We would also store a fingerprint for the configuration options and attributes. 

Ok, but this is not nearly enough to know when we need to reelaborate, right?  Elaboration depends also on the types of the constants in the environment, the attributes (which lemmas are marked with `[simp]`), and given that you can inspect the environment from tactics, actually everything in the environment.

> So, if we add a set_option or attribute command in the beginning of the file the whole cache would be invalided. We can add an option that ignores the fingerprint.

I think ignoring the changes is probably the wrong direction to go here.  It would be pretty confusing if you worked on a file, all warnings are gone, but the next morning when you reopen the file (and the caches are empty) half of the proofs would be wrong...

Looking at other projects, rust uses a pretty simple approach to handle invalidation in this kind.  Basically they'd store for every elaboration what parts of the environment were accessed, for example:
- type of `nat.succ`
- type of `eq`
- declarations annotated with the `simp` attribute
- ....

And then they only recompile if these accessed attributes have changed.  I don't think we can (easily) do this in Lean however, as I don't see any way to intercept these accesses to the environment.

gebner:
> I'm not sure I understood the issue. In the current module.cpp, we use this approach. The .olean files contain module_name objects, but we use absolute file names to identify modules in the import procedure: https://github.com/leanprover/lean/blob/lean3/src/library/module.cpp#L415-L418
> This approach seems to work well.

At some point during the development, using absolute file names everywhere really simplified the code.  But you're right, the current approach for import names together with absolute filenames for identification should work just fine.

leodemoura:
> Ok, but this is not nearly enough to know when we need to reelaborate, right? Elaboration depends also on the types of the constants in the environment, the attributes (which lemmas are marked with [simp]), and given that you can inspect the environment from tactics, actually everything in the environment.

The fingerprint takes care of attributes and options. 
We also stored a fingerprint for used dependencies. https://github.com/leanprover/lean/blob/master/src/library/definition_cache.h#L19
Here we only track the direct dependencies. So, we may still have problems. 
This approach worked well on Lean2.

> Looking at other projects, rust uses a pretty simple approach to handle invalidation in this kind. Basically they'd store for every elaboration what parts of the environment were accessed,

This is a nice solution, but as you said, it can't be easily implemented in Lean.
I have to think about it.

gebner:
> > if you parse a file and require the proof of a theorem, elaborate it right on the same thread. But this makes the value of the thread-local variables even more unpredictable.
> 
> It is not clear to me what the problem is.

The theorem can be elaborated with two different stack traces (slightly simplified):
1) if it is executed by a worker thread, let's say because we wanted the error messages in the server and put all messages on the queue:

```
elaborator::elaborate_with_type
single_definition_cmd_core::lambda (<- sets thread-local variables)
task_queue::spawn_worker
```

2) if it is executed as a dependency, let's say because of a reveal command:

```
elaborator::elaborate_with_type
single_definition_cmd_core::lambda (<- sets thread-local variables)
task_queue::join
reveal_cmd
parser::parse_command (<- sets thread-local variables)
parser::parse_commands (<- sets thread-local variables)
```

Depending on when the proof is revealed, it is executed with different thread-local variables.

leodemoura:
> I think ignoring the changes is probably the wrong direction to go here. It would be pretty confusing if you worked on a file, all warnings are gone, but the next morning when you reopen the file (and the caches are empty) half of the proofs would be wrong...

I agree. What about the following?
We use multiple passes that are more and more precise.
In the initial passes we always use the cache. The idea is to report problems as quickly as possible.
Sure, we can miss some errors because we used the cached value.
Then, if no error is found, the server tries again without using cache, and catches any missing error due to caching consistency problems (I'm assuming these are rare). 
Question: is it easy to interrupt the expensive pass if a new request arrives in the server?

In this kind of solution, we don't have to track dependencies.

leodemoura:
> The theorem can be elaborated with two different stack traces (slightly simplified):

Can't the task set these thread local variables? 
That is, we store in the task the values of these thread local variables (when the task is created), and restore them before we execute the task. Do you think this is too hackish?

gebner:
> We use multiple passes that are more and more precise.
> In the initial passes we always use the cache. The idea is to report problems as quickly as possible.
> Sure, we can miss some errors because we used the cached value.
> Then, if no error is found, the server tries again without using cache, and catches any missing error due to caching consistency problems (I'm assuming these are rare). 

So basically we'd prioritize proofs that are definitely out of date, versus those which definitely rely on outdated information.  This seems like a good strategy.

> Question: is it easy to interrupt the expensive pass if a new request arrives in the server?

The interruption would not be too difficult, and interrupting out-of-date requests is probably something we want to do anyway.  (The main obstacle here is that we want to reuse _some_ out-of-date requests when we reuse parser snapshots.)  However updating the messages is more difficult, right now we assume all results are immutable once they're produced.

One issue here is that I haven't yet figured out how to reuse messages at all--except for the parser snapshots.  This has the unfortunate side-effect that messages "flicker" a bit in emacs: if you edit a lemma, the messages below first vanish (as they are being elaborated), and then re-appear once they are elaborated.

> In this kind of solution, we don't have to track dependencies.

Well, we do track a very compressed version of the dependencies, a hash basically.

gebner:
> Can't the task set these thread local variables? 
> That is, we store in the task the values of these thread local variables (when the task is created), and restore them before we execute the task. Do you think this is too hackish?

I see two approaches here, neither of which seems more reliable:

1) Copy all thread-local variables (basically `memcpy`).  I'm not sure you can even do that in C++.  Besides, most of these thread-local variables are pointers that will be invalid when the task is executed.

2) Just copy a hand-picked list of thread-local variables (again, I'm sure we'll miss some).  Here, we also have the problem that we cannot copy some of them.  For example, `scope_pos_info_provider` just sets a _pointer_ to a `pos_info_provider`, which is an abstract class.  We could add a virtual clone method to `pos_info_provider`.  Then there is also the `abstract_type_context` in `scope_trace_env`.

leodemoura:
> 2) Just copy a hand-picked list of thread-local variables (again, I'm sure we'll miss some). Here, we also have the problem that we cannot copy some of them. For example, scope_pos_info_provider just sets a pointer to a pos_info_provider, which is an abstract class. We could add a virtual clone method to pos_info_provider. Then there is also the abstract_type_context in scope_trace_env.

For `pos_info_provider`, we can change it to a `std::shared_ptr`, and we just store this shared pointer in the task.

For `scope_trace_env`, we can skip it. We just create a new one when we start the task execution.

I acknowledge this is hackish since we are doing it case by case.

gebner:
> For pos_info_provider, we can change it to a std::shared_ptr, and we store just these shared pointer in the task.

The `pos_info_provider` is the parser usually, I'm not sure we want to keep the whole parser around.  But we could just store the `parser_pos_info_provider` instead.

> For scope_trace_env, we can skip it. We just create a new one when we start the task execution.

What happens if you don't give the type_context to `scope_trace_env`, do you just get worse trace messages, or none at all?  The only reason I had to add `scope_trace_env` was because otherwise the trace messages were missing completely.

I'm beginning to be convinced that these elaboration tasks just have to run on their own thread.  Setting up the thread-local variables correctly is difficult enough, but they shouldn't depend on random input (i.e. which thread gets there first).  If anything is missing, at least it will be missing consistently then and we can debug it.

leodemoura:
> But we could just store the parser_pos_info_provider instead.

Yes, this is what I had in mind.

leodemoura:
> What happens if you don't give the type_context to scope_trace_env

`tout()` is a dummy if it is not in the scope of 

```
    scope_trace_env(environment const & env, options const & opts, abstract_type_context & ctx);
```

OR

```
    scope_trace_env(environment const & env, abstract_type_context & ctx);
```

The pretty printer needs information (e.g., notation declarations) stored in the environment.
It also needs to infer types (`ctx` provides this functionality).
I'm assuming tasks are created only for elaborating declarations.
If this is true, we can create a simple `abstract_type_context` like the parser.
Assuming the task stores `env` and `opts`. We can setup the `scope_trace_env` by using:

```
        type_checker tc(env);
        scope_trace_env  scope(env, opts, tc);
```

Remark: the pretty printer cannot create its own type checker because it doesn't know how the type of locals and metavariables has been encoded. In the tactic framework, the types are stored in a `local_context`. The `type_checker` object doesn't know how to handle these `local_context`s since they are not defined in the kernel. This is why tactics create their own `scope_trace_env`. Example: 
https://github.com/leanprover/lean/blob/lean3/src/library/tactic/simplify.cpp#L64

gebner:
Ok, thanks for the explanation!  It's easy to set each of the scopes up, it's just difficult to figure out the complete list.  This is what I have so far:

``` cpp
scoped_expr_caching disable(false);  // FIXME: otherwise sigma.eq fails to elaborate
io_state ios = // ...
scope_global_ios scope1(ios);
type_context tc(env, opts, mctx, lctx);
scope_trace_env scope2(env, opts, tc);
auto pos_provider_ = pos_provider;
scope_pos_info_provider scope3(pos_provider_);
```

leodemoura:
> It's easy to set each of the scopes up, it's just difficult to figure out the complete list. 

I agree.

> This is what I have so far:

The list looks good. 
We have scopes at `decl_util.h` (e.g., `declaration_name_scope`). I'm assuming we don't need to set them in the task since they are set when we are elaborating a definition/theorem.
https://github.com/leanprover/lean/blob/lean3/src/frontends/lean/definition_cmds.cpp#L558-L559

leodemoura:
> I'm beginning to be convinced that these elaboration tasks just have to run on their own thread. 

It makes sense. My concern is performance. We have many thread local caches. If we create a new thread per task, these caches cannot be reused.
For example, @kha implemented a caching mechanism for user attributes. 
https://github.com/leanprover/lean/blob/lean3/src/library/tactic/user_attribute.cpp#L144-L197
This is super useful. For example, I use it here: https://github.com/leanprover/lean/blob/lean3/library/init/meta/simp_tactic.lean#L275
The`simp_lemmas` is constructed only once (and cached) IF the user does not add/remove "simp" attributes.

gebner:
Just to make this clear, I do not want to spawn a new thread per elaboration.  I just want to run them at the "top-level" of a thread, i.e. not inside the execution of another task.

> We have many thread local caches. If we create a new thread per task, these caches cannot be reused.

Ok, this is a bit of an unforeseen complication.  So even if we had dedicated elaboration-only threads, it would still be a bad idea to elaborate proofs from different modules on them?
That is, we should try not to elaborate proofs in the following order (in one thread): `module1/thm1`, `module2/thm4`, `module1/thm2`, `module2/thm3`?

The user-attribute-caching code seems to be really non-deterministic in this scenario, the cache would be regenerated at basically random environments.  If you forget to declare a dependency in the caching attribute or you have a collision in the fingerprints, you probably stand no chance of debugging it.

leodemoura:
> Ok, this is a bit of an unforeseen complication. So even if we had dedicated elaboration-only threads, it would still be a bad idea to elaborate proofs from different modules on them?
> That is, we should try not to elaborate proofs in the following order (in one thread): module1/thm1, module2/thm4, module1/thm2, module2/thm3?

Yes, this probably a bad order.

> If you forget to declare a dependency in the caching attribute or you have a collision in the fingerprints, you probably stand no chance of debugging it.

I'm not very concerned about these scenarios. The first one is a bug. The collision problem bugged me for a while. @kha and I discussed it in the summer. I think it is is extremely unlikely situation. @kha: any thoughts?

Kha:
> The collision problem bugged me for a while. @kha and I discussed it in the summer. I think it is is extremely unlikely situation. @kha: any thoughts?

I feel like we can live with that risk, especially on the meta level.

> > That is, we should try not to elaborate proofs in the following order (in one thread): module1/thm1, module2/thm4, module1/thm2, module2/thm3?
> 
> Yes, this probably a bad order.

Interesting, is rebuilding those caches that costly? I always assumed that they would be useful mostly with the repeated access patterns from elaboration or automation inside a single proof.

leodemoura:
> Interesting, is rebuilding those caches that costly? I always assumed that they would be useful mostly with the repeated access patterns from elaboration or automation inside a single proof.

Right now, they are not. However, it is conceivable we will have huge `simp_set`s in the future.
I'm concerned about a scenario where we have a `simp_set` with thousands of lemmas, and a file with hundreds of lemmas being proved by `simp`.

Kha:
> I'm concerned about a scenario where we have a simp_set with thousands of lemmas, and a file with hundreds of lemmas being proved by simp.

Right, that actually sounds quite realistic. So we should have per-module task queues and workers that  change queues either only after their current queue has been emptied, or not at all.

gebner:
So I have updated the code quite a bit.  Major changes from last week:
- Messages are now no longer passed around as task results.  Instead, there is a **single global mutable store for messages** called `versioned_msg_buf`.  This allows us to reuse old messages, and doesn't require `task_result<list<message>>` fields everywhere.  At the moment, `info_manager` instances are stored here as well.  The basic idea is the following; there is a counter for the current version, i.e. we start at 1, and when you change a file in emacs, we're then at 2, after the next change we are at 3.  The message buffer is then a map of _message buckets_.  Each of these buckets is identified by a name, has a version, and contains a list of messages.  When writing a message to a bucket, we only write it if the version matches.  The buckets are also hierarchical (using the same hierarchy as the names that identify them): this allows us to remove obsolete messages very easily.  This all works pretty much automatically and completely non-intrusively: I only needed to add 5 lines to the parser for this approach to work reliably.
- **Tasks have a (dynamically computed) list of dependencies**; execution is only attempted after all dependencies are finished.  This almost completely removes the need to wait for other tasks during execution.  As a result, we no longer need to worry much about spawning extra threads / nesting tasks.  You can still wait for the results of a tasks that you didn't declare a dependency on, and the task queue might spawn an extra worker to guarantee progress.
- **Tasks are now queued eagerly.**  Since tasks now have side-effects (they output messages), it does not make sense to execute them lazily.  Nevertheless, I think we can have the main advantage of lazy evaluation by prioritization and priority inversion, i.e. we do not have to wait for all previous proofs to finish elaborating in order to evaluate a reveal command.  This seems to work fairly well in practice.
- **Tasks are now defined by creating subclasses** instead of using lambdas.  This has several benefits. 1) We can add more metadata.  For example, each task now carries a description.  This description is printed on the command-line and in emacs to show the currently executing task.  2) We can introspect tasks at runtime.  For example the prioritization function takes a tasks, looks at which file it is from and what kind of task it is and then decides which priority it gets.  This also makes it possible to modify the scheduling so that tasks from the same module get executed on the same thread.  3) You need to be explicit about which objects are in the closure of the task.  With lambdas it is easy to accidentally include large data structures in the closure, increasing memory consumption unnecessarily.
- **Module imports are now again stored as basically `pair<name, optional<unsigned>>`** (as they are in the lean3 branch).  This makes dependency resolution in the `module_mgr` a bit ugly, but it works for now.

Notable open issues:
- There is a big difference in how theorems are imported from files that are 1) parsed from source and those that are 2) parsed from olean files: they are imported as 1) axioms and 2) theorems, respectively.  The problem is that when we write olean files to disk, we first replace the axioms by the proven theorems, and then we no longer know which theorems were delayed.  We can't delay all proofs during import either, because then the equation compiler complains.
- Implicit universe parameters in theorems do not work.  The problem is that `elaborator::finalize` renames the parameters in the type but you can't apply the same renaming afterwards to the elaborated proof term.
- Editor integration.  Quite a lot of the code is somewhat untested because the emacs interface starts a separate lean server for every file.  The atom interface starts one server for the whole editor, but then again many commands don't specify the file name they should operate on.  We need to rework the server interface a bit.

Short feature list:
- Parallel compilation of files and parallel elaboration of proofs across files.
- Emacs interface with incremental error reporting.
- `lean a.lean` will automatically recompile the dependencies of `a` (if the olean files are not up to date).
- `lean --make` does pretty much the same as `linja`

EDIT: mixed up the first two paragraphs.

gebner:
EDIT: to clarify, `linja` is the current `lean3` branch, `lean --make` is this PR.

Regarding performance:

```
[gebner:~/lean/library] lean3+* ± rm **/*.olean; time linja
[87/87] "/home/gebner/lean/bin/lean"  -o "/home/gebner/lean/library/data/bitvec.olean" /home/gebner/lean/library/data/bitvec.lean

real    0m7.308s
user    0m14.439s
sys     0m0.694s
[gebner:~/lean/library] lean3+* ± rm **/*.olean; time lean --make

real    0m6.304s
user    0m19.445s
sys     0m0.539s
[gebner:~/lean/library] lean3+* ± rm **/*.olean; time lean -j1 --make

real    0m11.195s
user    0m10.939s
sys     0m0.315s
[gebner:~/lean/library] lean3+* ± rm **/*.olean; time lean -j0 --make

real    0m10.738s
user    0m10.443s
sys     0m0.269s
```

(The `-j0` argument uses a single-threaded "task queue" implementation, where tasks are not executed on a seperate thread, but directly at the point where they are submitted.)

gebner:
All tests pass.  There a few new sorry warnings (because I used sorry as the results of proofs that failed to elaborate), and surprisingly also a correct new error about two lemmas with the same name.

I also had to add lot of NOLINT annotations as cpplint gets very confused by lambda types, for example it always thinks that `std::function<void (some_class *)>` is a C-style cast...

leodemoura:
@gebner Thanks for the update. I will go over it today. I will post feedback tomorrow morning.

leodemoura:
> Implicit universe parameters in theorems do not work. The problem is that elaborator::finalize renames the parameters in the type but you can't apply the same renaming afterwards to the elaborated proof term.

Could you post here a small example that doesn't work?

leodemoura:
> Editor integration. Quite a lot of the code is somewhat untested because the emacs interface starts a separate lean server for every file

Suppose we have two open files `a.lean` and `b.lean`, and they both import `c.lean`. Moreover `c.lean` is outdated, and needs to be recompiled. It seems the two different servers will try to compile `c.lean`, right?
Can `c.olean` get corrupted in the process because there are two different processes trying to write it?

leodemoura:
> I also had to add lot of NOLINT annotations as cpplint gets very confused by lambda types, for example it always thinks that std::function<void (some_class *)> is a C-style cast...

I have the same problem. I agree this is quite annoying. We should eventually fix the cpplint.py. Or, try to find a newer version that has support for C++11

leodemoura:
> There is a big difference in how theorems are imported from files that are 1) parsed from source and those that are 2) parsed from olean files: they are imported as 1) axioms and 2) theorems, respectively. The problem is that when we write olean files to disk, we first replace the axioms by the proven theorems, and then we no longer know which theorems were delayed. We can't delay all proofs during import either, because then the equation compiler complains.

What is your plan for this one?

leodemoura:
What about memory consumption?
I use following script to retrieve memory consumption information:

```
#!/usr/bin/env bash
# memusg -- Measure memory usage of processes
# Usage: memusg COMMAND [ARGS]...
#
# Author: Jaeho Shin <netj@sparcs.org>
# Created: 2010-08-16
set -um

# check input
[ $# -gt 0 ] || { sed -n '2,/^#$/ s/^# //p' <"$0"; exit 1; }

# TODO support more options: peak, footprint, sampling rate, etc.

pgid=`ps -o pgid= $$`
# make sure we're in a separate process group
if [ $pgid = $(ps -o pgid= $(ps -o ppid= $$)) ]; then
    cmd=
    set -- "$0" "$@"
    for a; do cmd+="'${a//"'"/"'\\''"}' "; done
    exec bash -i -c "$cmd"
fi

# detect operating system and prepare measurement
case `uname` in
    Darwin|*BSD) sizes() { /bin/ps -o rss= -g $1; } ;;
    Linux) sizes() { /bin/ps -o rss= -$1; } ;;
    *) echo "`uname`: unsupported operating system" >&2; exit 2 ;;
esac

# monitor the memory usage in the background.
(
peak=0
while sizes=`sizes $pgid`
do
    set -- $sizes
    sample=$((${@/#/+}))
    let peak="sample > peak ? sample : peak"
    sleep 0.1
done
echo "memusg: peak=$peak" >&2
) &
monpid=$!


# run the given command
exec "$@"
```

leodemoura:
In the lean server, what does it happen if we receive a new request from Emacs and we still have pending tasks?
Do we have to wait the pending tasks from a previous request? If not, do we eventually kill these pending tasks? Do we have a timeout for tasks?

gebner:
> > Implicit universe parameters in theorems do not work. The problem is that elaborator::finalize renames the parameters in the type but you can't apply the same renaming afterwards to the elaborated proof term.
> 
> Could you post here a small example that doesn't work?

``` lean
lemma foo : ∀B A : Type _, ∀a:A, a=a :=
let x : ∀A, ∀a:A, a=a := @rfl in λB, x
```

I encountered this problem in less contrived situations as well, however all of them seem to magically work now.

gebner:
> Suppose we have two open files a.lean and b.lean, and they both import c.lean. Moreover c.lean is outdated, and needs to be recompiled. It seems the two different servers will try to compile c.lean, right?
> Can c.olean get corrupted in the process because there are two different processes trying to write it?

No, for two reasons:
1. At the moment we don't write olean files in interactive mode.  Mainly because we don't keep track of what modules depend on a file changed in the editor (but not written to disk yet).
2. Writing and reading olean files uses `exclusive_file_lock` and `shared_file_lock`.

gebner:
> What about memory consumption?

This is a good point.  For the current library folder, it's 648M:

```
[gebner:~/lean-build-mgr/library] parbld* ± `which time` -f %M lean --make
648128
```

(This is the same value as the `memusg` script.)

For library_dev, it's about the same.  If you build both at once it's 1139M.

The increase is not that large for single files (i.e. if all dependencies are imported from olean files on disk).  For example `library_dev/logic/basic.lean` (a file with many lemmas) goes from 27M to 32M.

Memory usage seems to increase linearly with the number of loaded modules that are parsed from lean files.  I tried this on a few files in `library_dev/automation/super` (almost only meta defs), and the memory usage is about `15M + 20M*n` where `n` is the number of files.  We probably share almost no data between modules.

gebner:
The peak 648M memory consumption is without parser snapshots and without info_manager support.  However if you parse the whole standard library from source in the lean server (i.e. with parser snapshots, info_manager, including tactic states, for every file), then it's not much more, only 739M.

Kha:
> For the current library folder, it's 648M

That doesn't sound bad at all to me. Isabelle/jEdit is known to be a total memory hog (read: multiple GBs), and the problem is only compounded by having the PolyML GC and the JVM GC running at the same time. Crucially, Isabelle still doesn't allow one to unload a module (AFAIK), so I regularly had to restart the whole editor and make sure to load the right file in order to minimize dependencies.

gebner:
> > There is a big difference in how theorems are imported from files that are 1) parsed from source and those that are 2) parsed from olean files: they are imported as 1) axioms and 2) theorems, respectively. The problem is that when we write olean files to disk, we first replace the axioms by the proven theorems, and then we no longer know which theorems were delayed. We can't delay all proofs during import either, because then the equation compiler complains.
> 
> What is your plan for this one?

I can see two possible solutions:

1) An easy non-intrusive solution is to add an environment extension that just contains a `name_map<task_result<certified_declaration>>`.  The theorems that are still delayed at the end of a module would then be serialized as-is to the olean file without replacing them in the environment first.  We would also export which theorems from other modules were revealed.  This approach would also pull the reveal logic out of the parser, and make reveal an indepdendent operation on environments.

2) The completely crazy version.  We drop the requirement of explicit reveal commands, and instead change the value field for theorem declarations from `expr` to essentially `task_result<expr>`.  We then automatically wait for the delayed proofs during delta-reduction.

leodemoura:
> This is a good point. For the current library folder, it's 648M
> ...
> memory usage is about 15M + 20M*n where n is the number of files. 

Suppose the standard library gets really big (e.g., 500 files). It would be tricky to build it on a standard  notebook using `lean --make`. I guess we can address this problem (if it happens) by executing `lean --make` in each subdirectory (e.g., `init`, `data`, ...), right?

gebner:
> Suppose the standard library gets really big (e.g., 500 files). It would be tricky to build it on a standard notebook using lean --make. I guess we can address this problem (if it happens) by executing lean --make in each subdirectory (e.g., init, data, ...), right?

Yes, this is one possibility.

Another possibility would be to simulate the current approach.  So right now the `module_mgr` keeps track of a `module_info` structure for each module.  For modules loaded from olean, this just contains the list of direct dependencies and the contents of the olean file.  For modules loaded from lean, this basically contains the final environment, the map of delayed theorems and the olean version of the final environment where the delayed theorems are still axioms.
It would be an easy change to add an option where just immediately replace the version which contains the environment with the olean code that we would write to disk, and then discard environment, etc.

EDIT: If you run `lean --make` on the debugger directory, it's 69M for example:

```
[gebner:~/lean-build-mgr/library] parbld* ± time lean --make debugger
0.92user 0.03system 0:00.85elapsed 112%CPU (0avgtext+0avgdata 69504maxresident)k
48inputs+192outputs (0major+15540minor)pagefaults 0swaps
```

leodemoura:
> 2) The completely crazy version. We drop the requirement of explicit reveal commands, and instead change the value field for theorem declarations from expr to essentially task_result<expr>. We then automatically wait for the delayed proofs during delta-reduction.

I don't think this one is crazy. It looks pretty good. 
For theorems imported from .olean files, we don't need the `task_result<expr>` overhead.
So, we could just add a new kind of theorem declaration (e.g., `delayed_theorem`) that uses `task_result<expr>` instead of `expr`. That is, we can add `theorem`s and `delayed_theorem`s to an environment. Moreover, when we retrieve a theorem from the environment, we should not be able to distinguish whether we have a `theorem` or `delayed_theorem`.

Users may be concerned about the kernel correctness if we use `task_result<expr>` in the kernel.
We can address this concern by telling users that they can always re-typecheck the generated .olean files using trust level 0.

gebner:
> I don't think this one is crazy. It looks pretty good.

Ok, then I'll try to make it work.

> For theorems imported from .olean files, we don't need the task_result<expr> overhead.

There is little overhead in creating a `task_result<T>` object without a task.  It's basically just a smart pointer to a `task_result_cell<T>` structure that has a `T` field.  The size of that structure should be `sizeof(T) + 48` if I see it correctly.  Accessing the value just requires a atomic load (in addition to the pointer traversal).
For theorems the extra 56 bytes are probably tiny compared to the expressions for the type and the value.  (Furthermore, an empty `task_result<T>` object is just the null pointer.)

> Moreover, when we retrieve a theorem from the environment, we should not be able to distinguish whether we have a theorem or delayed_theorem.

I agree.  However I think we need to return the `task_result<expr>` version for all kinds of theorems, as you still need some basic information about the declaration for delayed theorems as well, in particular the type and the universe parameters.

So I'd propose to add an extra `task_result<expr> m_proof` field to the declaration_cell structure.  The value of a theorem is always the value of the value of this new field.  The `check` function then returns a `certified_declaration` object where the declaration's `m_proof` field points to a task that type-checks the provided proof.
In trust level 0, the type-checker could check the proof immediately when constructing a certified_declaration.

gebner:
BTW, I see that the refcount in declarations starts from 1.  The current convention is to start from 0, right?

Kha:
> For theorems imported from .olean files, we don't need the task_result<expr> overhead.

Leo, you previously talked about "deserialization on demand" experiments in Lean 2. Couldn't we use the same concept of an "expr future" to represent both delayed deserialization and delayed proving? It probably doesn't quite fit in with the current `task_result` API (we would need support for 'cold' results that haven't been started yet), but conceptually it feels like the same operation to me.
You also mentioned some problems with duplicated exprs when deserializing on demand. While I couldn't quite follow the explanation, would it be correct to assume that using caching futures would resolve that problem?

leodemoura:
> However I think we need to return the task_result<expr> version for all kinds of theorems, as you still need some basic information about the declaration for delayed theorems as well, in particular the type and the universe parameters.

I didn't get this one. For example, the type will always be an `expr`. Even if we have a delayed theorem, right?

leodemoura:
> So I'd propose to add an extra task_result<expr> m_proof field to the declaration_cell structure. The value of a theorem is always the value of the value of this new field. The check function then returns a certified_declaration object where the declaration's m_proof field points to a task that type-checks the provided proof.
> In trust level 0, the type-checker could check the proof immediately when constructing a certified_declaration.

I think this is a good plan.

leodemoura:
> BTW, I see that the refcount in declarations starts from 1. The current convention is to start from 0, right?

Yes.

gebner:
> I didn't get this one. For example, the type will always be an expr. Even if we have a delayed theorem, right?

Yes, that's correct.  I was referring to having two different kinds of declarations, delayed theorems and normal theorems; and then returning only normal theorems from the environment.  This does not work as we usually want to get the type without having to wait for the proof, which returning a normal theorem entails.  So we always have to return delayed theorems.  And then we might as well not have normal theorems at all, for simplicity's sake.

leodemoura:
> So we always have to return delayed theorems. And we might as well not have normal theorems at all, for simplicity's sake.

I agree. The overhead you have described above is indeed small.

leodemoura:
> Couldn't we use the same concept of an "expr future" to represent both delayed deserialization and delayed proving? 

Yes, in principle, we could.

leodemoura:
> While I couldn't quite follow the explanation, would it be correct to assume that using caching futures would resolve that problem?

@Kha When we serialize an expression such as `g (f a) (f a)`, we encode `f a` only once. The second `f a` is stored as a reference `#idx`. The same happens if `f a` occurs in multiple declarations. 
When we are deserializing, after deserializing the first `f a`, we store an entry in the following table.
https://github.com/leanprover/lean/blob/master/src/util/object_serializer.h#L55
Note that, we don't store the `idx` in the .olean file, we rely on the fact that we deserialize objects in the same order we serialize them. 

So, in the current .olean format, we _must_ deserialize declarations in the other they were serialized. 
This prevent us from deserializing on demand.

In the experiments that I have tried, I fixed this problem by serializing each theorem proof independently. That is, I started the serialization for each theorem proof with an empty table:
https://github.com/leanprover/lean/blob/master/src/util/object_serializer.h#L20
In this approach, if `f a` occurs in two different lemmas, then it will be serialized twice. 
The .olean file is slightly bigger, but we can deserialize a theorem proof with an empty table:
https://github.com/leanprover/lean/blob/master/src/util/object_serializer.h#L47
That is, it doesn't depend on previous deserialization tasks.

leodemoura:
> `lemma foo : ∀B A : Type _, ∀a:A, a=a :=
> let x : ∀A, ∀a:A, a=a := @rfl in λB, x`
> I encountered this problem in less contrived situations as well, however all of them seem to magically work now.

I understood the problem. I will fix it tomorrow.

leodemoura:
@gebner The commit above should fix the implicit universe issue in examples such as:

``` lean
lemma foo : ∀B A : Type _, ∀a:A, a=a :=
let x : ∀A, ∀a:A, a=a := @rfl in λB, x
```

gebner:
> @gebner The commit above should fix the implicit universe issue in examples such as:

Awesome!  I have added it to the branch and rebased.

The `task_result<expr>`-in-declaration approach works out really nicely.  However now every access to the environment (or more precisely, calling `decl.get_value()`) can potentially cause a wait on a delayed proof, which we want to avoid.  The main ones:
- Checking whether a lemma is refl.  I think we could just check whether the pre-expression is literally `rfl`, and then set a `rfl_lemma` attribute on that definition.  The `is_rfl_lemma` function would then just check for that attribute.  Do we generate rfl lemmas in other places than `definition_cmds.cpp` as well?
- Inlining `_main` lemmas.  Where are they created?
- Noncomputability checks where the type is not in Prop.  This only affects a few lemmas about equality, so I'm not too worried about this.

gebner:
> In the lean server, what does it happen if we receive a new request from Emacs and we still have pending tasks?
> Do we have to wait the pending tasks from a previous request? If not, do we eventually kill these pending tasks? Do we have a timeout for tasks?

We try to cancel some tasks (i.e. prevent them from being executed, not interrupting them), but that is restricted to parsing.  The tasks that are would be really important to cancel are proof elaborations from the current file that are no longer necessary.  Unfortunately, because of the parser snapshots, we cannot just cancel all tasks from the current file.

Ideally we would cancel not only the scheduled tasks, but also interrupt the currently executing ones.  We already have a nice global `get_g_interrupt` flag that is check in many places, so actually interrupting the running tasks will not be too difficult.  I am just trying to figure out how to best check what tasks need to be cancelled, and how to wire that up nicely.  We already have the information if the task's messages would still be valid, maybe I can just leverage that.

leodemoura:
> Ideally we would cancel not only the scheduled tasks, but also interrupt the currently executing ones. We already have a nice global get_g_interrupt flag that is check in many places, so actually interrupting the running tasks will not be too difficult. 

Yes. BTW, at `interrupt.h`, we define `interruptable_thread`. It has a `request_interrupt` method.

leodemoura:
> Checking whether a lemma is refl. I think we could just check whether the pre-expression is literally rfl, and then set a rfl_lemma attribute on that definition. The is_rfl_lemma function would then just check for that attribute.

Yes, if the pre-term is `rfl`, we can avoid the problem by compiling the theorem as a definition.

> Do we generate rfl lemmas in other places than definition_cmds.cpp as well?

In the equation compiler, we add them. 
https://github.com/leanprover/lean/blob/lean3/src/library/equations_compiler/util.cpp#L298

leodemoura:
> Inlining _main lemmas. Where are they created?

They are also created by the equation compiler. BTW, the equation compiler will also create `_match` auxiliary definitions for nested pattern matching.

https://github.com/leanprover/lean/blob/lean3/src/library/equations_compiler/util.cpp#L248-L285

gebner:
> > Checking whether a lemma is refl. I think we could just check whether the pre-expression is literally rfl, and then set a rfl_lemma attribute on that definition. The is_rfl_lemma function would then just check for that attribute.
> 
> Yes, if the pre-term is rfl, we can avoid the problem by compiling the theorem as a definition.

Hmm, are you suggesting that when we encounter a `lemma foo : a=b := rfl` we compile it as if it were a `def foo : a=b := rfl`, i.e. with `decl.is_theorem() == false`?
What is the effect of setting `decl.is_theorem() == false`?

leodemoura:
> Hmm, are you suggesting that when we encounter a lemma foo : a=b := rfl we compile it as if it were a def foo : a=b := rfl, i.e. with decl.is_theorem() == false?
> What is the effect of setting decl.is_theorem() == false?

We would still add it as a theorem, but we would elaborate it as we elaborate definitions.
That is, we would take this branch:
https://github.com/gebner/lean/blob/parbld/src/frontends/lean/definition_cmds.cpp#L763

gebner:
> We would still add it as a theorem

The problem then happens when we add this theorem to a simp set: we check whether it is rfl using `is_rfl_lemma`--and `is_rfl_lemma` has no way of knowing how the theorem was produced.

leodemoura:
> The problem then happens when we add this theorem to a simp set: we check whether it is rfl using is_rfl_lemma--and is_rfl_lemma has no way of knowing how the theorem was produced.

Can we check whether the value for a task is already available?
That is, we add a method `optional<expr> declaration::try_get_value(declaration const & d)`.
For definitions, it just return the value.
For theorems, it returns some value if the value is available. If the task has not been completed, it returns none.

By elaborating `rfl` lemmas eagerly, we make sure their values are always available.

Then, we use `try_get_value` instead of `get_value` to define `is_rfl_lemma`.

gebner:
> Can we check whether the value for a task is already available?
> That is, we add a optional<expr> declaration::try_get_value(declaration const & d).

We can.  Indeed we have it already, it's `decl.get_value_task().peek()`.
But I think it's a bad idea to make decisions based on whether a task is finished or not.  Consider the following lemma (with a hypothetical `sleep` function):

``` lean
lemma bar : 0 + 0 = nat.zero := by do
sleep 100,
proof ← to_expr `(@rfl nat nat.zero),
exact proof

-- ...a few hundred lemmas later...

attribute [simp] bar
-- is bar now a refl-lemma?
```

Whether `bar` is now a refl-lemma depends on how long it took to parse the lemmas in between.  Less than 100ms, it's not a refl-lemma, otherwise it is.

leodemoura:
> Whether bar is now a refl-lemma depends on how long it took to parse the lemmas in between. Less than 100ms, it's not a refl-lemma, otherwise it is.

Why would someone define a `rfl` lemma using tactics?
Are you concerned about some complex tactic that tries many things, and eventually returns a proof by reflexivity? In this scenario the user may experience the behavior you have described.
I'm not sure how problematic it would be.

If we want to be safe, we can add an internal attribute `[_refl_lemma]`. We set this attribute only when the pre-term is `rfl`. That is, if the pre-term is rfl, we elaborate the theorem eagerly, and then set this internal attribute. The `is_rfl_lemma` uses this attribute instead of checking `d.get_value()`.

gebner:
> Are you concerned about some complex tactic that tries many things, and eventually returns a proof by reflexivity? In this scenario the user may experience the behavior you have described.

If you put it like this, I think it will definitely happen.  People will definitely use `by simp` on everything.

> If we want to be safe, we can add an internal attribute [_refl_lemma]. We set this attribute only when the pre-term is rfl. That is, if the pre-term is rfl, we elaborate the theorem eagerly, and then set this internal attribute. The is_rfl_lemma uses this attribute instead of checking d.get_value().

This is my preferred solution.

leodemoura:
> This is my preferred solution.

I agree.

gebner:
There is now some ugly support to interrupt running tasks.  It uses the same infrastructure as the messages, when a task's messages would no longer be shown, it gets canceled.

Old proof elaboration tasks are only cancelled when the parser is finished (because that's when we figure out exactly which ones to reuse, which ones to redo, and which ones are no longer there).  After trying this out a bit, I think this is a bit of a problem.  Mainly because this increases 1) the latency before we get to work on the new proofs, and 2) long-running tasks can delay the parsing.  Maybe the idea to keep a source file position with each task and then just cancel everything after the changed position is not that bad at all.  The reason I have shied away from that approach is because it would make it difficult to reuse results from below the changed positions, i.e. if we want to cache the elaboration results then we would only want to cancel proof elaboration tasks whose input environment and pre-expression have not changed.

Also I think that I'll need to refactor the task queue code a bit.  In the beginning I thought it would be a good idea to have a generic task queue interface that does not know anything about files, positions, io_state, etc., and then build these abstractions on top.  This leads to somewhat baroque class hierarchy with separate `module_task<T>` and  `task<T>`, even though everything is a `module_task<T>` anyways.  However I think that this only introduces unnecessary complexity; all the tasks we consider are tied to a source file, they all want a reasonable set-up with a global io_state, etc.

BTW, I stumbled upon some interesting C++ trivia; what does the following code print?
```c++
struct foo {
  ~foo() { throw std::exception(); }
};
int main() {
  try {
    foo f;
    throw std::exception();
  } catch (std::exception) {
    std::cout << "exception caught" << std::endl;
  }
  return 0;
}
```
This actually happened, as we use exceptions to interrupt threads.  In particular, the `~elaborator` destructor finalizes its `info_manger` by instantiating the meta-variables, which uses the replace visitor, which checks for interrupts, which can throw an exception.  Now imagine the elaborator is deallocated because an exception was thrown...

leodemoura:
> In particular, the ~elaborator destructor finalizes its info_manger by instantiating the meta-variables, which uses the replace visitor, which checks for interrupts, which can throw an exception.

This is a problem. We can get in trouble by throwing an exception inside a destructor.
If another exception is already being propagated, the application may crash.

leodemoura:
@gebner I think we should make the finalization explicit. That is, we have a method `finalize_info` in the `elaborator`. We explicitly invoke this method when using the `elaborator`.
It is less convenient to use, but the code will be safer. What do you think?

gebner:
> @gebner I think we should make the finalization explicit. That is, we have a method finalize_info in the elaborator. We explicitly invoke this method when using the elaborator.
> It is less convenient to use, but the code will be safer. What do you think?

This was the first thing I tried.  However it is *really* inconvenient, as we want to get the info_manager in the case of an elaborator exception as well.  (You want to see the current goal and other infos *particularly* in the case something goes wrong.)  If you run the finalization in the destructor, you get this behavior for free.

For now, I have just wrapped the call to the info manager in a `try {} catch (...) {}` block.

leodemoura:
> For now, I have just wrapped the call to the info manager in a try {} catch (...) {} block.

I think this is fine.

gebner:
How do we proceed with this PR?  As far as I can see, most of the major issues have been addressed.  I've been using this branch for the last week without any issues, and of course all tests pass.  Does anybody see points that should be addressed?

There is probably some polishing to be done, there are still a few compiler warnings, obsolete arguments, and missing documentation.  I didn't do much of that so far, as rebasing is already quite a bit of work, and I didn't want to add extra code to rebase.

leodemoura:
@gebner I managed to merge `parbld` with `lean3`. I tested it using Ubuntu 12.04, 16.04, and MSYS2 with Windows 8. I will be pushing in a few minutes. I will test on OSX and Win 10 later.
I fixed some compilation problems and fixed all warnings.

leodemoura:
merged with `lean3`

gebner:
Awesome!  I'll see if I can disentangle the kernel-library dependency.

PR #1182, opened 2016-11-07
Bring back basic support for 'info at pos' and 'find definition'

Kha:
Shows the full name and type of an identifier or symbol when hovering over it, and on `M-.` instantaneously jumps to its definition.

The implementation is based on the old `info_manager`, but the server part and especially the client part are heavily simplified so that we can achieve a clean, client agnostic API. The new `info_manager` is kept around in parser snapshots and passed to the elaborator for information gathering. The elaborator part is simplistic and records the uninstantiated type (which I guess we will need anyway for `lean-show-id-keyword-info`). @leodemoura will surely know better than me where to record the instantiated type and other missing information.

The ['go to definition' emacs implementation](https://github.com/leanprover/lean/commit/22b64223c30cb9a8594ee42380aaaa3f3b95a398#diff-c0531ef3a8be3a2d2c3a4af948d55a9aR59) is just as simplistic right now. There actually is a [unified interface](https://www.gnu.org/software/emacs/manual/html_node/emacs/Xref.html) for doing these things, but of course it's emacs 25-only. Too bad.


bin/lean-gdb.py
```diff
@@ -1,3 +1,11 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (c) 2016 Microsoft Corporation. All rights reserved.
+# Released under Apache 2.0 license as described in the file LICENSE.
+#
+# Author: Sebastian Ullrich
```
gebner: I wasn't even aware this file existed.  This is awesome!  And it even works in CLion!  (I couldn't find any info on how to use it, so for the record, you just have to add `source ..../bin/lean-gdb.py` to your `~/.gdbinit` file.)

How difficult would it be to add pretty-printers for `declaration` and `expr`?

Kha: `gdb bin/lean` should load the file automatically because of the file name (or print a message that the script must be whitelisted). I haven't tested it in CLion yet.

> How difficult would it be to add pretty-printers for declaration and expr?

For `expr` and a few other types, we already have top-level `print` functions. Reimplementing those in Python sounds wasteful, so while it's not the nice way, we could just call into C++ from inside the GDB pretty printer for that. For `expr`, you'd usually want to use the Lean pretty printer, which also needs an `environment`. We actually had helper functions for that and I was confused why I couldn't find them until I noticed that you had [removed](https://github.com/leanprover/lean/commit/b05b514cc2425a08a77c4c41ff4cc0bb1aabaabf#diff-be1b53035565a98b2f6d4fa12841cd36L1764) them :joy: .

gebner: > gdb bin/lean should load the file automatically because of the file name (or print a message that the script must be whitelisted). I haven't tested it in CLion yet.

Ok, this only works if you run the lean binary in the bin directory, but not if you run the one in the cmake build directory.

> We actually had helper functions for that and I was confused why I couldn't find them until I noticed that you had removed them￼ .

Ooops, they were buried in a lot of commented-out code so I thought they were similarly unused.  I'll put them back.

> For expr, you'd usually want to use the Lean pretty printer,

Sometimes, yes.  But sometimes, I want to see the actual structure of an expression, and then you're completely lost, as the debugger will only show you the common fields for an expr since expr cells don't have RTTI info..

BTW, can you access the fields of the cell in a declaration from GDB?  I always get "incomplete type".

src/frontends/lean/json.cpp
```diff
@@ -4,8 +4,7 @@ Released under Apache 2.0 license as described in the file LICENSE.
 
 Author: Gabriel Ebner
 */
-#if defined(LEAN_SERVER)
```
gebner, line 7: Is there a reason you removed the LEAN_SERVER ifdef here?  I know it should rather be called LEAN_JSON because it disables all JSON features, but I think Leo added it to support compilers which do not work with the json.hpp library.

Kha: Oops, I didn't remember the flag was introduced for the JSON library. I guess I'll need a few more `#if`s in `info_manager` then, urgh.

src/frontends/lean/elaborator.cpp
```diff
@@ -2766,7 +2779,9 @@ static expr translate(environment const & env, local_context const & lctx, expr
 
 expr nested_elaborate(environment & env, options const & opts, metavar_context & mctx, local_context const & lctx,
                       expr const & e, bool relaxed) {
-    elaborator elab(env, opts, mctx, lctx);
+    // TODO(sullrich): pass through info_manager?
```
leodemoura: @kha I think the easiest solution is to use a thread local variable for the `info_manager`.

bin/lean-gdb.py
Kha: > Ok, this only works if you run the lean binary in the bin directory, but not if you run the one in the cmake build directory.

Ah, I've never run that binary before :) .

> Sometimes, yes. But sometimes, I want to see the actual structure of an expression, and then you're completely lost, as the debugger will only show you the common fields for an expr since expr cells don't have RTTI info..

Well, you can already do `call print(e)` :) . I just looked up the hack for calling C++ functions from gdb, it's... beautiful http://stackoverflow.com/a/22798055/161659

> BTW, can you access the fields of the cell in a declaration from GDB? I always get "incomplete type".

Hm, I'm not sure I've ever tried that.

src/frontends/lean/elaborator.cpp
Kha: I wasn't sure whether the `info_manager` would have to be part of the elaborator snapshots, could we get redundant/wrong information when using a global one?

bin/lean-gdb.py
gebner: > Well, you can already do call print(e)

That doesn't do anything for me:

```
(gdb) call print(val)
(gdb) call print(val)
```

I was thinking about casting the expr_cell to the appropriate subtype based on the kind field.  So if `m_ptr->kind == 2`, then cast to `expr_const *`.  Is this possible in gdb?

leodemoura:
@kha How will the new `info_manager` work with the parallel compiler being developed by @gebner?
It seems multiple threads will try to update the same `info_manager`, right?
If yes, then we need to add a mutex, right?

src/frontends/lean/elaborator.cpp
leodemoura: @kha I think the top-level elaborator should set the thread local variable. 
So, we make sure that nested elaborators use the same `info_manager` being used by the top-level one.

Kha:
It's not quite compatible with the current design yet, but @gebner suggested having a separate `info_manager` per delayed proof. This would make invalidation very simple. The server would then query the `info_manager` in the latest parser snapshot (as it does right now) as well as all elaboration `info_manager`s.

gebner:
> How will the new info_manager work with the parallel compiler being developed by @gebner?

The current idea is to have multiple info_managers, basically one for each file, and one for each proof.  Each query would then go through all of them.

However, I'm currently rethinking the parallel compilation approach a bit.  In particular when it comes to messages, I think it is better to keep a global mutable repository of messages that gets updated by each task instead of the always-correct separate message lists that are implemented right now.  I think this is the easiest approach to display old messages as long as possible until they are replaced by new one, e.g. when a proof is elaborated.  Here each module would have a map of buckets (basically one for parsing, one for each theorem elaboration, one for each print command, etc.), that could then be replaced individually.  The multiple info-manager approach could be used here as well, or we could go back to one global info-manager, which would need to handle invalidation and locking itself.

In any case, I don't think that's a big problem and easy to switch later on.

bin/lean-gdb.py
Kha: > That doesn't do anything for me:

Huh, that's weird.

```
(gdb) p fn
$1 = (const lean::expr &) @0x7fffffffac80: {m_ptr = 0x1a57598}
(gdb) call print(fn)
bool.tt
```

> I was thinking about casting the expr_cell to the appropriate subtype based on the kind field. So if m_ptr->kind == 2, then cast to expr_const *. Is this possible in gdb?

Yes, that works out of the box.

```
(gdb) p ((expr_const *)fn.m_ptr)->m_name
$6 = lean::name('bool'.'tt')
```

src/frontends/lean/json.cpp
leodemoura: Yes, the `LEAN_SERVER` ifdef was added to support compilers where compiling the server support did not work. I called the option `LEAN_SERVER` because it disables the server module.

bin/lean-gdb.py
gebner: > Huh, that's weird.

Oops, didn't notice the output goes to stdout, which is a different tab from gdb output in clion.  Works for me as well now.

> Yes, that works out of the box.

Ok, it's clear that I can cast it myself.  But how do you know it is a expr_const?

Kha: > But how do you know it is a expr_const?

Probably not at all in any automatic way. I really think calling into C++ is the simplest way to go here.

gebner: I think I'll do it via python.

leodemoura:
@kha I'm getting the following error (Emacs 24.5.1):

```
error in lean-server command handler: (void-function if-let)
```

leodemoura:
@kha I get the error when I try to use `M-.` (jump to definition).

Kha:
@leodemoura Ugh, I'm getting slightly annoyed that there's no versioning information at all in the emacs code. At least this time there was a simple replacement. Can you cherry-pick the last commit from this branch?

leodemoura:
> @leodemoura Ugh, I'm getting slightly annoyed that there's no versioning information at all in the emacs code. At least this time there was a simple replacement. Can you cherry-pick the last commit from this branch?

Thanks. It fixed the problem.

However, I found a new one. `M-.` on a function defined in the same file.
Example, if I press `M-.` on the second `g` at

```
def g : nat → nat := λ x, x

check g
```

I get the error:

```
error in lean-server command handler: (wrong-type-argument stringp nil)
```

leodemoura:
@kha If the function has been defined in the current file, the API `environment.decl_olean` returns `none`.
Here is an example:

``` lean
open tactic

def g : nat → nat :=
λ n, 0

meta def show_pos (n : name) : command :=
do env   ← get_env,
   pos   ← returnopt (env^.decl_pos_info n),
   olean ← returnopt (env^.decl_olean n) <|> return "current file",
   trace $ to_string n ++ " was defined at " ++ olean ++ " : " ++ to_string pos

run_command show_pos `add
run_command show_pos `g
```

Kha:
Ah, the rebase seems to have changed something there, I think. It should work now.

leodemoura:
It worked! Thanks.

PR #1185, opened 2016-11-08
feat(emacs/lean-info): push marker to go back after find-definition

gebner:
The new `xref` package that @Kha mentioned in #1182 allows you to go back after calling find-definitions.  Supporting this feature just requires us to call `xref-push-marker-stack` when going to a definition.  Then you can go back via `M-,` (default keybinding if you have it).

This feature also integrates seamlessly with the spacemacs `C-t` binding. Regarding spacemacs:  @Kha, you mentioned that you were interested in better spacemacs support for lean, maybe also a spacemacs layer.  Any ideas on how to get that started?  Do you already have some customization?  So far, I've only added support for find-definition as the `M-.` binding conflicts with evil-mode:
```
(spacemacs|define-jump-handlers lean-mode lean-find-definition)
```

### DIFF
--- a/src/emacs/lean-info.el
+++ b/src/emacs/lean-info.el
@@ -56,6 +56,7 @@
         (funcall cont record))))))
 
 (cl-defun lean-find-definition-cont (&key file line column)
+  (when (fboundp 'xref-push-marker-stack) (xref-push-marker-stack))
   (find-file file)
   (goto-char (point-min))
   (forward-line (1- line))


Kha:
> @Kha, you mentioned that you were interested in better spacemacs support for lean, maybe also a spacemacs layer. Any ideas on how to get that started?

I don't have any experience with writing layers yet, but we should probably keep the layer in-tree instead of pushing it into the spacemacs repo until lean3 is finished (if @leodemoura doesn't mind). We can adjust `dotspacemacs-configuration-layer-path` in the lean-emacs initialization to include it (or the other way around).

> Do you already have some customization? So far, I've only added support for find-definition as the M-. binding conflicts with evil-mode:

I haven't added any keybindings yet - in fact, I've called `lean-find-definition` only explicitly so far in order to test it :laughing: . I have configured indentation:

```
  (push 'lean-mode spacemacs-indent-sensitive-modes)
```

and removed `syntax-checking/post-init-popwin` to get flycheck out of popwin (I should probably wrap that in an advice).

leodemoura:
@gebner This is nice. It is only for Emacs >= 25, right?

leodemoura:
@kha 

> we should probably keep the layer in-tree instead of pushing it into the spacemacs repo until lean3 is finished (if @leodemoura doesn't mind). 

I think this would be nice. 

BTW, I should install spacemacs :)

gebner:
Yes, this is only for emacs 25.

Kha:
I went looking around for other go to definition implementations, this should do the same thing for emacs<25: https://github.com/abingham/emacs-ycmd/blob/master/ycmd.el#L1157

PR #1193, opened 2016-11-21
fix(library/init/logic): if_true and if_false need the instance as ar…

dselsam:
…gument

### DIFF
--- a/library/init/logic.lean
+++ b/library/init/logic.lean
@@ -817,11 +817,11 @@ lemma if_simp_congr {α : Type u} {b c : Prop} [dec_b : decidable b] {x y u v : 
 @if_ctx_simp_congr α b c dec_b x y u v h_c (λ h, h_t) (λ h, h_e)
 
 @[simp]
-def if_true {α : Type u} [H : decidable true] (t e : α) : (@ite true H α t e) = t :=
+def if_true {α : Type u} {H : decidable true} (t e : α) : (@ite true H α t e) = t :=
 if_pos trivial
 
 @[simp]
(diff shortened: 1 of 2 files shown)


Kha:
Would an implicit argument be even more general, at least in the context of rewriting? I encountered similar problems after importing `prop_decidable` in Lean 2 and switched to [implicit versions of `(d)if_pos/neg`](https://github.com/Kha/electrolysis/blob/master/thys/move.lean#L624). But perhaps that sort of `decidable` instances mixing is more of a fringe case.

dselsam:
I don't see a use case where the instance wouldn't unify and hence would need to be synthesized either way. I changed it anyway though as you suggest, in case there are some edge cases I am failing to consider.

Kha:
I was never quite sure why it happened even with `prop_decidable`'s low priority, but I do get unification errors like these with the original `if_pos` in Lean 2:
```
  -fail to unify equation source
    @ite
      ...
      (@decidable_and (@RangeFrom.start ℕ r ≤ @length T s)
         (@length T s ≤ @length T s)
         (decidable_le (@RangeFrom.start ℕ r) (@length T s))
         (decidable_le (@length T s) (@length T s)))
      ?M_4
      ?M_5
      ?M_6
  with subterm
    @ite
      ...
      (@decidable_and (@RangeFrom.start ℕ r ≤ @length T s)
         (@length T s ≤ @length T s)
         (classical.prop_decidable (@RangeFrom.start ℕ r ≤ @length T s))
         (classical.prop_decidable (@length T s ≤ @length T s)))
      ...
```

PR #1194, opened 2016-11-22
feat(library/tactic/simplify): better debug.simplify.try_rewrite tracing

dselsam:


src/library/tactic/simplify.cpp
```diff
@@ -478,7 +478,6 @@ simp_result simplify_core_fn::rewrite(expr const & e) {
 
     list<simp_lemma> const * srs = sr->find(e);
     if (!srs) {
-        lean_trace_d(name({"debug", "simplify", "try_rewrite"}), tout() << "no simp lemmas for: " << e << "\n";);
         return simp_result(e);
```
leodemoura, line 481: Why did you remove this one?

dselsam: It is a deluge of information, and I have only ever used it to see if a given lemma was even considered or not. The `try_rewrite` message still handles this, since if you search for the lemma name in the trace and do not find it, then it was never tried (after a trace-message is inserted for the `umeta` case, which I just added and will push shortly). If you like the old message, then I suggest we put it in an inner trace-class that can be disabled independently, i.e. `debug.simplify.try_rewrite.no_lemmas`.

dselsam: Note: there are still many other gaps in the debug-simplify-tracing scheme. Not every kind of congruence triggers a message, and no failures do when applying congruence. Making the tracing/debugging intuitive and robust is a priority for me, but I would like to discuss the overall simplifier/converter/rewriter ecosystem with you next week before fine-tuning much further.

PR #1202, opened 2016-11-24
updates to Lean 2 HoTT library

fpvandoorn:
This pull request moves some of the lemmas from the [Spectral repository](https://github.com/cmu-phil/Spectral) to the Lean 2 HoTT library.

It also fixes/cleans-up various things:

Changed definitions:
* `is_homomorphism` is defined for `has_mul` and not for `group`
* only reducible definitions are used in the instances to `has_add`/`has_zero`/`has_neg`
* slightly redefine `pcofiber` and `pwedge` to make using recursors for them nicer
* redefine Eilenberg-Maclane spaces (and prove their uniqueness)
* give a completely different (simpler) definition of `smash`

Renamings:
* rename theorems with `transport_eq` or `pathover_eq` in the name to use `eq_transport` and `eq_pathover`
* rename `comm_group` and `CommGroup` to `ab_group` and `AbGroup`
* `natural_square` and `natural_square_tr` have their name interchanged because the (old) `natural_square_tr` was used much more frequently

Other:
* merge files `algebra.hott`, `pointed_pushout` and `quotient_functor` into `group_theory`, `pushout` and `quotient` respectively
* have coercions from `Group` to `Set*`, `Set` and `Type*`
* increase the precedence of `◾` so that we can now write `p ◾ q ⬝ r ◾ s` for `(p ◾ q) ⬝ (r ◾ s)`
* interchange the two arguments of `whisker_right`. Now it's consistent with `pwhisker_right`, which has a better order, since the first argument could usually be inferred from context.


fpvandoorn:
Oops, one file imported a deleted file, and I forget to fix the tests. I force-pushed and now it should compile fine.

PR #1205, opened 2016-12-01
inform compilers about allocator

diakopter:
gperftools documentation recommends it..


diakopter:
ugh

gebner:
Was there something wrong with this patch?

diakopter:
Well travisci didn't seem to like the cmake syntax

gebner:
I see, it's probably using an old cmake version.

PR #1090, opened 2016-07-20
Lean 3 backend (take 2)

jroesch:
This set of commits implements a compiler from Lean's core expression type to C++. This is still very much a prototype and needs more testing and features.

I am closing the previous PR and opening a new one since I changed branches, and this second attempt is more tightly integrated with the Lean 3 redesign. 

I think there is still a little bit of work left before this code should be merged, I'm currently streamlining the CLI and code, and trying to enable passes to be written in Lean itself.  

Previous PR: https://github.com/leanprover/lean/pull/1055

PR #1210, opened 2016-12-02
Lean.native (take 3)

jroesch:
This set of commits implements a compiler from Lean's core expression type to C++ *written in Lean*. This is a prototype and needs more testing, optimizations, re-organization and more.

My second attempt was closed in the Lean 3 move, my final attempt has rewritten most of the compiler into Lean.

There are still a few more TODOs, but hopefully we can merge this Monday next week.

Previous PR2: #1090 
Previous PR1: #1055


src/util/process.cpp
```diff
@@ -32,7 +32,7 @@ void process::run() {
         buffer<char*> pargs;
         for (auto arg : m_args) {
             std::cout << arg << std::endl;
```
gebner: The whole loop body can be written as just `pargs.push_back(arg.c_str())`.  The pointer returned by `c_str` is valid as long as we own the string `arg` and do not pass it as a mutable reference somewhere.

gebner: Github doesn't let me comment a few lines down on `pargs.data()[pargs.size()] = NULL;`, so I'll do it here.  The only reason this works is because buffer starts out with a capacity of 16, otherwise this might already be a write to unallocated memory.  I'd use `pargs.push_back(nullptr)` here.

Also `pargs.data()[0]` seems redundant, doesn't `pargs[0]` work as well?

src/shell/lean.cpp
```diff
@@ -503,8 +503,9 @@ int main(int argc, char ** argv) {
             }
         }
 
+        // Options appear to be empty, pretty sure I'm making a mistake here.
         if (compile && !mods.empty()) {
-            auto opts = ios.get_options();
+            auto opts = mod_mgr.get_options(); // ios.get_options();
```
gebner: `mod_mgr.get_options()` just returns the initial options, not the options obtained after parsing a module (the method does not even know which module we're interested in).  At the moment we do not return the options from module_mgr, but you can just add a field to `module_info::parse_result`, right next to `m_env`.

Issue #1197, opened 2016-11-22
New lean server protocol


gebner:
With #1178 on its way, it becomes necessary to think about a revision of the server protocol.  The main big limitation with the current protocol is that it only handles a single file per server process.  Since adding support for multiple files is a significant change, it might make sense to roll up some other changes as well.

 * Support for multiple files.  The protocol somewhat supports this already, the immediate stumbling block is that the complete and info commands do not specify the filename.
   * It would be nice to have the list of currently visible files and maybe even line number ranges as well, in order to prioritize the visible stuff.
   * Unloading files?
 * Push messages.  Right now the only way to incrementally get error messages is by polling; we try and see which messages are already available, and if there are not complete yet, we try again 500ms later.  It would be nice if we could get those messages earlier.
   * At the moment, the server cannot send unsolicited messages; every response necessarily corresponds to a request.  One way to fix this is to introduce sequence numbers for every request: the server then includes this number in the corresponding response and the client can the figure out which response corresponds to which request (or was unsolicited).
 * What should the `LEAN_PATH` be?  With the one-server-per-file approach we do the following: we look for the closest parent directory that has a `.project` file and start the server there.
 * How do we want to implement the debugger interface?  Should we do it via the server interface?  (We do not have to do this right now.  I just want to make sure that we can easily integrate debugging later on.)

@Kha @jroesch Ideas, comments, wishes?

There was a bit of talk on slack about the [LSP](https://github.com/Microsoft/language-server-protocol), which is used by VS Code and the new Rust IDE support.  Does anybody have an opinion on whether/how we should adopt/copy it?
Personally I don't think it would buy us much to implement it directly.  Pretty much the only editor that supports it at the moment is VS Code, and we would still need to develop syntax highlighting and an input method to get VS Code support.  On the other hand, the LSP is focused pretty narrowly on syntax checking, and probably won't ever support debugging. For a potential VS Code plugin, I'd rather do a proxy that translates between lean and LSP, and also implements the VS-Code-specific debug protocol.

Kha:
> At the moment, the server cannot send unsolicited messages

That's mostly a restriction of tq.el, no? As long as the server makes sure never to send out outdated responses, I would assume that with something like `process-filter` the client could just handle and display that information, without actually relating it to a prior request.

> What should the LEAN_PATH be? With the one-server-per-file approach we do the following: we look for the closest parent directory that has a .project file and start the server there.

[...]

PR #1227, opened 2016-12-07
[RFC] New lean server protocol

gebner:
I have spent the day on drafting a prototype for the new lean server protocol
as discussed in issue #1197.  The implementation works, and you can try it out.
I would appreciate if someone with emacs 24 could see whether it works, I'm
almost certain I used some 25-only features.

To do (aka what's missing from #1197):
* *No indication whether lean is still running, or what it is working on.*
* Opening a file most likely recompiles all its reverse dependencies even if you don't change anything.
* No information about currently visible files is exchanged.
* Unloading single files is not supported.
* No debugger interface.

### Short overview of server protocol

It looks almost like before, with two main differences: 1) commands that return
responses such as info or complete now take an additional `seq_num` parameter,
which is a sequentially increasing number maintained by the client.  2)
Messages are no longer returned on check, but in separate unsolicited messages.

Example exchange:
```
18:37:08.976 -- server<= {"seq_num":0,"command":"sync","file_name":"/home/gebner/scratch7.lean","content":"example : true := ⟨⟩"}
18:37:09.025 -- server=> {"message":"file invalidated","response":"ok","seq_num":0}

18:37:15.954 -- server<= {"seq_num":1,"command":"info","file_name":"/home/gebner/scratch7.lean","line":1,"column":10}
18:37:15.965 -- server=> {"record":{"full-id":"true","source":{"column":10,"file":"/home/gebner/lean/library/init/core.lean","line":90},"type":"Prop"},"response":"ok","seq_num":1}
...
18:37:24.389 -- server<= {"seq_num":3,"command":"sync","file_name":"/home/gebner/scratch7.lean","content":"example : tre := ⟨⟩"}
18:37:24.407 -- server=> {"message":"file invalidated","response":"ok","seq_num":3}

18:37:24.622 -- server=> {"msg":{"caption":"","file_name":"/home/gebner/scratch7.lean","pos_col":10,"pos_line":1,"severity":"error","text":"unknown identifier 'tre'"},"response":"additional_message"}
```

Note that each command sent by the client carries a sequence number, which is
returned by the server with the corresponding response.  The last response does
not correspond to a request, it transports an error message.

Possible requests/commands (all of them have mandatory `command` and `seq_num` fields):
* `sync` (parameters `file_name`, `content`): informs the server about file contents
* `info` (parameters `file_name`, `line`, `column`): request information about the thing at the given position
* `complete` (parameters `file_name`, `line`, `column`): request auto-completion choices at the given position

Possible responses (the kind is indicated in the `response` field):
* `ok` (has `seq_num`): result of a successful request
* `error` (has `message`): error
  * can have an optional `seq_num` field if the error is related to a request
* `additional_message` (has `msg` field containing a single message)
* `all_messages` (has `msgs` field containing an array of messages)

In the spirit of "never transfer changes, always transfer the whole file", the
protocol does not support deleting messages in any fancy way.  If a message
should no longer be shown, you have to send an `all_messages` response
containing the now valid messages.  The `additional_message` response just adds
a single new message; but we could get rid of that.

Project handling is not very sophisticated either: we spawn one lean server per
project.  Invalidation across project boundaries is not supported.  If you
change something in the standard library and want to use it somewhere else,
please restart the lean server.

### Emacs interface

Since we have unsolicited responses from the server, we can no longer use
tq.el, but have to roll our own version with process filters.  No major
problems there.  We store a list of the currently valid messages client-side.
lean-debug-mode is now set globally instead of per-buffer.

On the flycheck side, we now use flycheck only as a dumb error message viewer
(see flycheck/flycheck#774).  We do the change detection ourselves.  200ms
after the last change, we send the contents to the server.  Whenever new
messages arrive, we let flycheck "check" the buffer.  Our flycheck checker only
returns the currently stored messages.  This is the same technique that
`intero-mode` uses as well.

### DIFF
--- a/src/library/module_mgr.cpp
+++ b/src/library/module_mgr.cpp
@@ -18,13 +18,14 @@ Author: Gabriel Ebner
 
 namespace lean {
 
-void module_mgr::mark_out_of_date(module_id const & id) {
+void module_mgr::mark_out_of_date(module_id const & id, buffer<module_id> & to_rebuild) {
     for (auto & mod : m_modules) {
         if (!mod.second || mod.second->m_out_of_date) continue;
         for (auto & dep : mod.second->m_deps) {
             if (dep.first == id) {
                 mod.second->m_out_of_date = true;
-                mark_out_of_date(mod.first);
+                to_rebuild.push_back(mod.first);
+                mark_out_of_date(mod.first, to_rebuild);
                 break;
             }
         }
@@ -291,13 +292,11 @@ void module_mgr::invalidate(module_id const & id) {
     m_current_period++;
 
     if (auto & mod = m_modules[id]) {
-        mod->m_out_of_date = true;
-        mark_out_of_date(id);
-
         buffer<module_id> to_rebuild;
-        for (auto & other : m_modules)
-            if (other.second->m_out_of_date)
-                to_rebuild.push_back(other.first);
+        mod->m_out_of_date = true;
+        to_rebuild.push_back(id);
+        mark_out_of_date(id, to_rebuild);
+
         for (auto & i : to_rebuild)
             try { build_module(i, true, {}); } catch (...) {}
     }
--- a/src/library/module_mgr.h
+++ b/src/library/module_mgr.h
@@ -82,7 +82,7 @@ class module_mgr {
     mutex m_mutex;
     std::unordered_map<module_id, std::shared_ptr<module_info>> m_modules;
 
-    void mark_out_of_date(module_id const & id);
+    void mark_out_of_date(module_id const & id, buffer<module_id> & to_rebuild);
     void build_module(module_id const & id, bool can_use_olean, name_set module_stack);
     std::vector<module_name> get_direct_imports(module_id const & id, std::string const & contents);
     void gather_transitive_imports(


gebner:
In addition to the missing stuff, maybe I should mention some of the features:
 * Supports multiple files in the same lean server.  Dependency invalidation should finally work.
 * When you open a lean file, it gets checked immediately.
 * Messages are displayed immediately when they are produced.
 * Completions are handled completely asynchronously as tasks in the task queue.

leodemoura:
It works on Emacs 24.

leodemoura:
I made the following changes to be able to compile the code on my machine:
https://github.com/leodemoura/lean/commit/58768940e33b4569ebf60c8d249a7b754c486d22

leodemoura:
I'm having problems with the `LEAN_PATH`.  Here is the repro:
Create files `a.lean` and `b.lean` in some directory, where `a.lean` contains
```
def x := 10
```
and `b.lean` contains
```
import a
eval x
```
Inside of Emacs, I get `invalid import 'a'\n\nfile 'a' not found in the LEAN_PATH`. I get the same error if I generate `a.olean`.
In the command line, `lean b.lean` works correctly, even if I don't have `a.olean`.

leodemoura:
Remark: the example above works if I create the files `a.lean` and `b.lean` in one of the folders of the standard library.

gebner:
> `lean b.lean`

I guess you're starting emacs from a different directory than the one where `b.lean` is, right?  For files that are not in projects (i.e. do not have a `.project` file somewhere above them), I am starting the lean server in the emacs working directory.

Instead of falling back to the working directory, we could fall back to the directory of the file.

leodemoura:
> I guess you're starting emacs from a different directory than the one where b.lean is, right?

I see.
Yes, it works if I start emacs at the `b.lean` directory. It also works when I add the `.project` file.

leodemoura:
> Instead of falling back to the working directory, we could fall back to the directory of the file.

I think this is a good idea.

gebner:
Some observations:
 * Emacs sometimes hangs and/or is extremely laggy when lean is running at 100% CPU (for example when `super` unsuccessfully tries to prove something)
 * A global error list would be nice.

leodemoura:
> Emacs sometimes hangs and/or is extremely laggy when lean is running at 100% CPU (for example when super unsuccessfully tries to prove something)

Is it clear why this happens? I did not manage to reproduce this problem yet.

Here is a minor problem. Suppose I am editing a proof `A`, and the proof `B` above `A` is broken, and `B` relies on expensive automation. Now, as I edit the file, the red marks on `B` keep blinking since the closest snapshot is above `B`, and whenever I change `A`, lean tries to reprocess `B`. 
If I add a command between `A` and `B` (e.g., `check 0`), then the "blinking" goes away. So, I'm wondering now whether we should save snapshot after `B` even if it is broken. What do you think?

leodemoura:
> Emacs sometimes hangs and/or is extremely laggy when lean is running at 100% CPU (for example when super unsuccessfully tries to prove something)

For me, Emacs is still responsive as an editor if I use `super` on a big proof. However, the green/red marks take a long time to update. Is that what you meant by "hangs and/or is extremely laggy"?

gebner:
> Is it clear why this happens? I did not manage to reproduce this problem yet.

Not at all.  I should've attached gdb then.  But I got it when just viewing the `automation/super/examples.lean` from before I fixed the breakage caused by the lean-native merge earlier today.

> Here is a minor problem. Suppose I am editing a proof A, and the proof B above A is broken, and B relies on expensive automation. Now, as I edit the file, the red marks on B keep blinking since the closest snapshot is above B, and whenever I change A, lean tries to reprocess B.

The fact that B is broken is only relevant in so far as that there's an error message.  You'd observe exactly the same phenomenon if B just had trace output.

I can't recall exactly why and can't find the discussion that led to it, but we discard the last snapshot as well because the command right before the snapshot might now be continued after the snapshot.  We introduced this behavior in https://github.com/leanprover/lean/commit/87ec1e0ac5a0d292262b7284c326d9891152eb4e

gebner:
> For me, Emacs is still responsive as an editor if I use super on a big proof. However, the green/red marks take a long time to update. Is that what you meant by "hangs and/or is extremely laggy"?

No, I meant laggy as in "emacs takes five seconds to process a keystroke" and "emacs no longer redraws the screen".  I'll try to reproduce it tomorrow.

leodemoura:
I have merged the commits sent so far.

Kha:
> I can't recall exactly why and can't find the discussion that led to it, but we discard the last snapshot as well because the command right before the snapshot might now be continued after the snapshot.

The underlying problem is that our syntax grammar does not have a mandatory terminator token for commands ending in terms. We do actually have the Coq `.` terminator, but that doesn't help if nobody is using it.

I believe the best job at placing snapshots we can do right now is to set one in _front_ of a command _after_ we've parse its command token -- in particular, there should be no snapshot at the end of the file. Only then do we know that the previous command is truly complete.

However, that would still invalidate the previous proof a few times while the user is typing out 'theorem` below it. We could do a basically perfect job by including a small tweak in our grammar: Do not allow empty lines in a top-level (non-parenthesized) function application, like in
```
lemma a :=
b

theore
```
While this would be somewhat unpleasing from a theoretic point by making our language whitespace-sensitive, I would argue that it should not create any problems in practice. Furthermore, it would allow context-sensitive autocompletion to know when it should only complete commands.

gebner:
> I believe the best job at placing snapshots we can do right now is to set one in front of a command after we've parse its command token

It's not immediately clear how that would would work, when we restore from a snapshot we move the scanner to the position of the snapshot.  If we then start to parse a command, it would fail because the first token has already been consumed.

A variant on this approach would be to just have a separate field that records the end-position of the first command token after a snapshot, i.e. the point at which we commit to use the snapshot and no changes after it make the snapshot invalid.

gebner:
> However, that would still invalidate the previous proof a few times while the user is typing out 'theorem` below it. 

We just need to train our users to type more quickly; you have 200ms between each keystroke or your CPU dies. :smile:

Kha:
> It's not immediately clear how that would would work, when we restore from a snapshot we move the scanner to the position of the snapshot. If we then start to parse a command, it would fail because the first token has already been consumed.

Sorry, the wording was a bit ambiguous. We can only act after parsing a command token, but then we create a snapshot directly in front of the token. That should be a simple change.

gebner:
> Sorry, the wording was a bit ambiguous. We can only act after parsing a command token, but then we create a snapshot directly in front of the token. That should be a simple change.

I don't think it matters when we create the snapshot.  The snapshot might be successfully created after parsing `theorem`, but the user can still change `theorem` to `thorem` and we need to discard the snapshot.  We need that information when we figure out which snapshots to throw away.

Kha:
> The snapshot might be successfully created after parsing theorem, but the user can still change theorem to thorem and we need to discard the snapshot.

Oh, you're completely right! We basically have to regard the command token to be part of both commands. Your suggestion of also recording the end position of the snapshot sounds fine then.

gebner:
@leodemoura Should I open a new PR for the unmerged commits?

leodemoura:
I will reopen

leodemoura:
> No, I meant laggy as in "emacs takes five seconds to process a keystroke" and "emacs no longer redraws the screen". I'll try to reproduce it tomorrow.

I managed to reproduce this behavior. It happened for me when Lean was consuming a lot of memory (e.g., > 16Gb). It was triggered by a temporary bug. 
I think this may also happen when executing tactics since we don't have a `check_memory` in the VM loop.
I will address this issue today.

gebner:
@jroesch There has been one addition: the server now sents unsolicited `current_tasks` messages, informing the client about the work the server is currently doing: that messages contains an `is_running` boolean parameter, an optional task `cur_task`, and an array of tasks `tasks`.  Each task has a `file_name`, `pos_col`, `pos_line`, and `desc`.

leodemoura:
@gebner Should we close this one?

gebner:
Yes.  I "smuggled in" the unmerged commits in another PR already.  We should make new PRs for further improvements.

PR #1232, opened 2016-12-08
fix(emacs/load-lean): require unicode-fonts package

gebner:


avigad:
@gebner @leodemoura The package `fill-column-indicator` is used to display a line e.g. at the 100th column. There is no reason to make this a dependency. Users can load it themselves if they want it. I am inclined to get rid of it everywhere (`lean-require`, `lean-settings`, `lean-mode`, ...). Any objection?

gebner:
Since `fill-column-indicator` is even disabled by default, I'm all for removing it.

leodemoura:
me too.

PR #1234, opened 2016-12-08
fix(util/thread): do not initialize thread_finalizers_manager in multiple threads

gebner:
So, I believe this is the reason the `thread_test` fails in the test-coverage and macOS builds.  Or maybe it's just one of the bugs, travis will tell.

The problem was the following: `thread_test` starts two threads, each of which allocates a memory pool, which registers a thread finalizers.  Registering a thread finalizer causes the global variable `g_aux` to be initialized if it is not set yet, which is unsafe if done from multiple threads in parallel.

### DIFF
--- a/src/tests/library/expr_lt.cpp
+++ b/src/tests/library/expr_lt.cpp
@@ -7,6 +7,7 @@ Author: Leonardo de Moura
 #include "util/test.h"
 #include "kernel/abstract.h"
 #include "library/expr_lt.h"
+#include "util/init_module.h"
 using namespace lean;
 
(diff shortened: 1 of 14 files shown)


gebner:
The windows build is severely broken now, everything segfaults, argh.

gebner:
Not only the windows build, apparently some of the tests are missing a call to `initialize_util_module`.

leodemoura:
@gebner I think it is not that bad. Only old tests that do not explicitly invoke initialization are breaking.

gebner:
Ok, next try.  Hopefully everything works now.  At least all the tests run locally here.

PR #1242, opened 2016-12-10
fix(emacs/lean-server): make sure all timers are wrapped in save-match-data

gebner:
@leodemoura Does this fix the issue with `M-/`?

### DIFF
--- a/src/emacs/lean-server.el
+++ b/src/emacs/lean-server.el
@@ -194,16 +194,16 @@
 
 (defun lean-server-sync ()
   "Synchronizes the current buffer state with lean server"
-  (save-match-data
-    (lean-server-send-command
-     'sync (list :file_name (buffer-file-name)
-                 :content (buffer-string)))))
+  (lean-server-send-command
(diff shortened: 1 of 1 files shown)


leodemoura:
@gebner It did not fix it. I am still experiencing the same bug with `M-/` and `query-replace`.

gebner:
Okay, that's bad.  I can't reproduce it under either linux or windows (both emacs 25.1).

leodemoura:
Could you install Emacs 24.5.1?

BTW, Ubuntu 16.04 uses Emacs 24.5.1. I'm experiencing the bug with the default Ubuntu 16.04 installation.

gebner:
Can reproduce with emacs 24.5, I'll try to figure out what goes wrong.

gebner:
@leodemoura I can now call `M-/` in emacs 24 without jumping to the top of the file.

leodemoura:
@gebner Thanks! It works for me too.

PR #1243, opened 2016-12-11
[RFC] Importing modules as a direct operation on environments

gebner:
Right now we import modules by replaying the serialized representation onto the environment.  Hence processing the imports for one files takes linear time in the number/size of the dependencies.  In particular, if you have n files, `a_1.lean` to `a_n.lean`, where each file imports the previous one, then compiling them all has quadratic runtime.

This PR is ultimately an attempt to reduce this cost (the runtime of the example above is only linear with this PR).  Concretely, we introduce a binary operation `env1.union_with(env2)` on environments.  This operation takes the union of the declarations in both environments (soundness of this inference requires just weakening as far as I can see, and this is true for our environments).  We extend this operation to all environment extensions as well.  However there it is much less well-behaved, e.g. it is non-commutative as we want to give precedence to later imports sometimes.

Importing a module is now defined in terms of the environment that module produces.  I.e. the lean code `import foo` now denotes the command `env = env.union_with(foo)`.  We implement one nice optimization: if `env` is an ancestor of `foo`, then we return `env.union_with(foo) = foo`.  Local attributes, etc., are hidden from other modules via scoping: we open a scope at the begin of a module (right after the imports), and then close it at the end.

The good news: I have gotten it to a point where every single test passes (with the exception of the always problematic `thread_test`), and where there are promising runtime and (unexpectedly) memory improvements.  Compiling the whole standard library from scratch now takes the following time and memory:
```
master:  33.03user 0.72system 0:13.48elapsed 250%CPU 1054M
This PR: 18.42user 0.19system 0:06.96elapsed 267%CPU  169M
```
So roughly twice as fast, and 6x less (peak) memory usage.

The bad news: compiling a single file takes much longer.  In order not to introduce any inconsistencies, we use the same importing approach for olean files as we do for lean files: we construct an environment for every imported module and then take the union.  Compiling just `standard.lean`:
```
master:  0.29user 0.01system 0:00.31elapsed  99%CPU 44M
This PR: 1.37user 0.06system 0:00.68elapsed 211%CPU 81M
```

Alternatives:

1. We could try to implement the ancestor-optimization from this PR with the deserialization-import: if a file uses `import foo bar baz`, then we cache the environment produced after importing `init` and `foo`, in the hope that other files can reuse that environment.

2. Right now the only way of obtaining a list of the modifications made by a module is by producing a byte-level serialization of that list.  We could decouple these two aspects and introduce intermediary data structures for the modifications.  Instead of a list of writers, an environment would then carry a list of modifications.  These modifications could either be serialized, or directly replayed onto another environment.

### DIFF
--- a/library/data/bitvec.lean
+++ b/library/data/bitvec.lean
@@ -22,7 +22,7 @@ begin unfold bitvec, apply_instance end
 def zero {n : ℕ} : bitvec n := tuple.repeat ff
 
 -- Create a bitvector with the constant one.
-def one {n : ℕ} : bitvec (succ n) := tuple.append (tuple.repeat ff : bitvec n) [ tt ]
+def one {n : ℕ} : bitvec (succ n) := tuple.append (tuple.repeat ff : bitvec n) (tt :: nil)
 
 section shift
 
--- a/library/data/bitvec.lean
+++ b/library/data/bitvec.lean
@@ -22,7 +22,7 @@ begin unfold bitvec, apply_instance end
 def zero {n : ℕ} : bitvec n := tuple.repeat ff
 
 -- Create a bitvector with the constant one.
-def one {n : ℕ} : bitvec (succ n) := tuple.append (tuple.repeat ff : bitvec n) (tt :: nil)
+def one {n : ℕ} : bitvec (succ n) := tuple.append (tuple.repeat ff : bitvec n) [ tt ]
 
 section shift
 
--- a/src/api/module.cpp
+++ b/src/api/module.cpp
@@ -5,6 +5,7 @@ Released under Apache 2.0 license as described in the file LICENSE.
 Author: Leonardo de Moura
 */
 #include <string>
+#include <vector>
 #include "util/lean_path.h"
 #include "library/module.h"
 #include "library/util.h"
--- a/src/api/module.cpp
+++ b/src/api/module.cpp
@@ -21,9 +21,11 @@ lean_bool lean_env_import(lean_env env, lean_ios ios, lean_list_name modules, le
     check_nonnull(ios);
     check_nonnull(modules);
     auto new_env = to_env_ref(env);
+    std::vector<module_name> imports;
     for (name const & n : to_list_name_ref(modules)) {
-        new_env = import_module(new_env, "", {n, optional<unsigned>()}, mk_olean_loader());
+        imports.push_back({n, optional<unsigned>()});
     }
+    new_env = import_modules(new_env, "", imports, mk_olean_loader(new_env));
     *r = of_env(new environment(new_env));
     LEAN_CATCH;
 }
--- a/src/frontends/lean/parser.cpp
+++ b/src/frontends/lean/parser.cpp
@@ -2190,6 +2190,7 @@ void parser::process_imports() {
         (mk_message(m_last_cmd_pos, WARNING) << "imported file uses 'sorry'").report();
 #endif
     }
+    m_env = push_scope(m_env, ios(), scope_kind::Namespace);
     m_imports_parsed = true;
 }
 
@@ -2258,7 +2259,7 @@ bool parser::parse_commands() {
         }
         scope_message_context scope_msg_ctx("end");
         save_snapshot(scope_parser_msgs);
-        m_env = clear_aliases(m_env);
+        m_env = pop_scope(m_env, ios());
         if (has_open_scopes(m_env)) {
             m_found_errors = true;
             if (!m_use_exceptions && m_show_errors)
--- a/src/frontends/lean/parser.cpp
+++ b/src/frontends/lean/parser.cpp
@@ -2166,18 +2166,17 @@ void parser::process_imports() {
     unsigned fingerprint = 0;
     auto imports = parse_imports(fingerprint);
 
-    for (auto & n : imports) {
-        try {
-            m_env = import_module(m_env, m_file_name, n, m_import_fn);
-        } catch (exception & ex) {
-            m_found_errors = true;
-            parser_exception error((sstream() << "invalid import '" << n.m_name << "'").str(),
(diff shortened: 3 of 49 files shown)


jroesch:
I think the performance improvements are a big win for things like the standard library, it would be great if its possible to reduce the overhead for single files. I don't know enough about of this code to know which approach is the right one.

leodemoura:
We need to address the issue you have described. I think option 2 is the way to go.
Since it will be changing the format of the .olean files. I think we can get another speedup by deserializing lemmas/definitions on demand. That is, the format of .olean would be composed of three blocks:

1- Declarations: name, type, flags, and position where value is located.
2- Modifications (as you described in option 2)
3- Values of theorems and definitions

Parts 1 and 2 should be much smaller than part 3. When load a .olean file, we deserialize parts 1 and 2, and store 3 as an array V of bytes. The theorem and definition objects contain the position where their serialized value is located in V. We access a small subset of the standard library in a Lean session. So, we should save a considerable amount of memory. We can keep of these V arrays per environment, and we just keep appending data to it as we load .olean files. If one day the library becomes gigantic, we can even store V in a temporary file. 

Another way to speedup the process is to merge many .olean files in a single one. For example, there is no reason to store the files in the init folder as separate .olean files. We should store them as single one.
The init folder has approx 100 .lean files. This will speedup the loading process for files that are not in the `init` folder. 

My main concern is scalability as the library grows. The lean3 stdlib is small in comparison with the one in lean2. 

BTW, I'm also super concerned that the `kernel` folder still depends on the `library` folder, and `library` depends on `frontends\lean`. We say that the Lean kernel is small. An easy way to confirm that is to check that kernel only depends on the `util` folder. It also feels weird the kernel indirectly depends on the `parser`. These dependencies were introduced by the `parbld` PR. We need to fix this.

gebner:
> We need to address the issue you have described. I think option 2 is the way to go.

Okay, it's probably safer as it is a smaller change to our current approach.  Let's see how the performance compares.

I did not mention one related change that was necessary for the union approach, but might be crucial to the memory savings with any of the approaches: right now, VM functions and cases have environment-specific IDs--hence we cannot reuse the bytecode across environments.  In this PR, I have changed the indices to be global.  The main downside that I see is that increases the size of the lookup tables in the VM state.  (If I were to haphazard a guess, I'd imagine not by more than 2x on average.)  My first plan was to scrap the lookup tables and index_maps alltogether, and replace them by direct pointers to the declarations--but this did not work because we modify the bytecode of existing VM functions when we compile mutually recursive functions.  @leodemoura What are your thoughts on this?

> Since it will be changing the format of the .olean files. I think we can get another speedup by deserializing lemmas/definitions on demand. That is, the format of .olean would be composed of three blocks:

I don't think that introducing an intermediary data structure for environment modifications would necessarily change the olean format at all.  Essentially the only change is the following:
```c++
// before:
typedef pair<std::string, std::function<void(environment const &, serializer &)>> writer;
// after:
class modification {
public:
  virtual environment perform(environment const &) = 0;
  virtual void serialize(serializer &) = 0;
};
```

We can implement on-demand deserialization separately.  To be honest, I'm not sure how much we would save by not deserializing expressions.  Looking at the profiler output(*), when compiling the whole standard library right now we spend about 35% of the time doing imports, and only 8% deserializing expressions.  Most of the time seems to be spent updating the extensions.  Besides, if we deserialize expressions on-demand, we need to synchronize this across multiple threads and maybe even wait on another thread for deserialization--the extra overhead might be more than any improvements we make.

(*) Take this with a grain of salt.  The perf-tools profiler that I'm using only samples the top 32k of the stack, and 32% of the time we're using more than 32k.  But I think the numbers should be in the right ballpark.

> Another way to speedup the process is to merge many .olean files in a single one.

Right now we're pretty far from being IO limited, so I'm not sure how much that would help.  For comparison, python also stores one `pyc` file for every `py` file--and I haven't heard too many complaints about python startup time.  But we can keep this in mind, it's a fairly straightforward change.

> My main concern is scalability as the library grows. The lean3 stdlib is small in comparison with the one in lean2.

AFAICT, we used the same importing approach in lean2.  Were there problems with startup speed or scalability?

> I'm also super concerned that the kernel folder still depends on the library folder, and library depends on frontends\lean.

I haven't forgotten about it, but I'll prioritize it now. The main issue was that tasks use the message data structures, and the message data structures references the info_manager, which should most likely remain in frontends/lean.  I'll probably move task queues and messages into util, and figure out something for the info manager.

leodemoura:
> @leodemoura What are your thoughts on this?

It is hard to predict the performance impact. It introduces a new synchronization point at `get_vm_index`. This function is mainly used when we are importing files and generating bytecode/native-code.
We don't compile definitions in parallel, only theorems, and we do not generate code for theorems. 
So, I think the impact may be minimal.

> We can implement on-demand deserialization separately. To be honest, I'm not sure how much we would save by not deserializing expressions. Looking at the profiler output(*), when compiling the whole standard library right now we spend about 35% of the time doing imports, and only 8% deserializing expressions. 

The goal is to save memory. I have made experiments with on demand deserialization, and the gains were substantial. I think I mentioned that on slack. By having all serialized values in a separate datastructure and deserialization on demand, we will be able to use different approaches for saving memory. Examples: store parts of the data-structure in a temporary file, use a simple compression procedure for individual values, etc.

> Right now we're pretty far from being IO limited, so I'm not sure how much that would help.

By merging .olean files, we minimize the number of union operations. 

> AFAICT, we used the same importing approach in lean2. Were there problems with startup speed or scalability?

My comment is based on the single file numbers you reported. The `lean2` numbers were faster for processing a single file. 
That being said, I was concerned about memory consumption and startup time in `lean2` too.

> I'll probably move task queues and messages into util, and figure out something for the info manager.

Can't we define an abstract `info_manager`?
We had a similar dependency problem with the `attribute_manager`, it needed to use very basic methods from the parser. We broke the dependency by adding an abstract parser class.

leodemoura:
We merged #1259

PR #1244, opened 2016-12-11
feat(tactic/exact_tactic): exact_core that takes transparency

dselsam:
I often want to use `exact` when I expect (near) syntactic equality. With big terms, one tiny name mix-up can cause the old `exact` tactic to take arbitrarily long. 

I left `exact` as `exact_core semireducible`, and added `rexact` for `exact_core reducible` though this doesn't seem to have any precedent in the library so far. One possibility for naming conventions: we could have full names `srewrite / sapply / sexact` for `semireducible`, `rrewrite rapply rexact` for `reducible`, and `nrewrite / napply / nexact` for `transparency.none`, with `rewrite := rrewrite, apply := sapply, exact := sexact` for the expected default values.

### DIFF
--- a/library/init/meta/tactic.lean
+++ b/library/init/meta/tactic.lean
@@ -286,7 +286,7 @@ meta constant mk_mapp_core  : transparency → name → list (option expr) → t
    Otherwise, try to find a local constant that has type of the form (t = t') or (t' = t).
    The tactic fails if the given expression is not a local constant. -/
 meta constant subst         : expr → tactic unit
-meta constant exact         : expr → tactic unit
+meta constant exact_core    : transparency → expr → tactic unit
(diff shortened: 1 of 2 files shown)


leodemoura:
> I often want to use exact when I expect (near) syntactic equality. With big terms, one tiny name mix-up can cause the old exact tactic to take arbitrarily long.

Do you understand why it doesn't fail quickly?
Is this a performance bug?

dselsam:
> Is this a performance bug?

That or an infinite loop.

> Do you understand why it doesn't fail quickly?

No. I saved the example and will investigate later this week.

dselsam:
Note: even though the underlying performance problem has been addressed, I still think it is desirable to support `exact_core`, especially given how simple it is to do so.

PR #1248, opened 2016-12-13
fix(emacs): do not query user before exiting emacs >= 25

Kha:
This is what we get for using cutting-edge features


Kha:
See also https://github.com/emacs-mirror/emacs/blob/master/src/process.c#L1696-L1700

Issue #554, opened 2015-04-21
[question] differences between constants and parameters


fpvandoorn:
When I write

``` lean
section
parameters {A : Type} (a : A) (...)
  (* stuff *)
end
```

My **intuition** of `A` and `a` is that as long as I stay inside the section, they are treated as constants i.e. treated exactly the same as

``` lean
constants {A : Type} (a : A) (...)
  (* stuff *)
```

Only when we end the `section`, magic happens, and the parameters are abstracted inside the definitions.

However, this intuition is false. In Lean they _are_ treated differently, as [multiple](https://github.com/leanprover/lean/issues/529#issuecomment-89869240) [issues](https://github.com/leanprover/lean/issues/530) indicate.

There are probably more differences. Can you give a (partial) list of differences between `parameters` (as long as you remain in the section) and `constants`?

leodemoura:
Let me first explain how the parameters are implemented.
When we declare a definition `foo` in a section:
1- Collect all parameters used by `foo`. This includes the explicit and implicit (by transitivity) ones.
2- Update the type (and value) of `foo` will all new parameters. 
3- Elaborate
4- Send elaborated `foo` to the kernel
5- Create an alias `foo` ==> `foo As` where `As` are the parameters used by `foo`

So, the declaration is abstracted right away, and we use the alias to hide this fact from the user.
This simulation "leaks" sometimes (e.g., the pretty printer issue you found).

The more important question is: why do we use this approach?
In Lean 0.1, we used an approach similar to the one used in Coq. Parameters were added like axioms/constants in the environment, then in the end of the section all declarations are abstracted with respect to these axioms/constants. This approach is simpler, but it has problems.
In March of 2014, Georges pointed out several of them (at the time, we did not have google groups for Lean, and we discussed this issue in an email between Jeremy, Cody and me). 

1- The type of a declaration changes over time. This affects the ability of compiling theorems in parallel. The "end section" is like a synchronization point.

2- We have to re-type-check the section declarations in the "end section" (like in Lean 0.1), or the abstraction mechanism must be part of the trusted code base (like in Coq). Georges pointed out that he found bugs in this part of Coq.

3- We need to implement a mechanism for preserving meta-information (attributes, notation, etc). 
In Lean 0.1, suppose we have

``` lean
...
-- Environment E1
section 
...
-- Environment E2
end
-- Environment E3
...
```

[...]

PR #1251, opened 2016-12-14
fix(frontends/lean/parser): support open inside sections with parameters

gebner:
At the moment the following works:
```lean
section
parameter α : Type
inductive foo : Type | a : α → foo | b
check (foo.b : foo)
end
```

But this doesn't:
```lean
open foo
check (b : foo)
```

It feels wrong to interpret the environment's aliases using the parser's local declaration map as done in this PR, but it is to the easiest solution I could think of here.


gebner:
Please wait with merging for a second, I just noticed that this fix doesn't work if we're inside a namespace.

gebner:
Okay, it works now, but I have no idea whether that's the right way to solve it.

gebner:
A related issue is that parameters are not exposed to the tactics (at least as far as I can tell), not even `to_expr` respects them:
```lean
run_command (do
  e ← to_expr `(b),
  t ← infer_type e,
  trace t)

-- prints: Π (α : Type), foo α
```

leodemoura:
I'm working on this issue.
To fix the `tactic` related issue, I will refactor the `aliases` module. 
We need to store the parser "local_refs" (such as `b` ==> `foo.b  α`) in the environment. Otherwise, the tactic framework will not see them since it does not have access to the `parser`.

leodemoura:
In the commit above, I addressed the original issue by changing how the aliases are created. 
The code will be cleaner after the `aliases` module refactoring since it will also store the entry `b` ==> `foo.b α`

Kha:
Leo, I believe we discussed implementing a more "declarative" version of `open` at some point where instead of replaying commands onto the environment, we derive a (cached) name map from all active `open` commands when needed. I proposed this for fixing the old "have to re-open a namespace after changing it" issue, but I assume it could also simplify this issue?

leodemoura:
The commit above solves the issues above, and cleanup the solution.

leodemoura:
@kha Yes, I remember. We should implement it in the future.
However, the issue raised here is a little bit orthogonal to it. 
In Lean, parameters are simulated. See discussion at https://github.com/leanprover/lean/issues/554

leodemoura:
@gebner Note that the example 
```lean
run_command (do
  e ← to_expr `(b),
  t ← infer_type e,
  trace t)
```
still doesn't work after the commit above. The issue is that it indirectly uses `α` which is not in the local context. However, the following similar example works.
```lean
include α
example : true :=
by do
  e ← to_expr `(b),
  t ← infer_type e,
  trace t,
 triv
```
because the `include` command forces `α` to be in the local context of the `example`.

leodemoura:
@gebner @kha At #554, I tried to summarize the advantages and disadvantages on the approach used to implement parameters in Lean vs the one used in Coq.

leodemoura:
I didn't push the commit above to the main repo yet because it broke a few tests.

gebner:
This solution is much nicer than my quick fix.

> The issue is that it indirectly uses α which is not in the local context.

What is a direct use?  Only parameters that are referenced in the type?

In the following we now have `α` in the local context, but something still goes wrong:
```lean
def ex : foo := begin trace_state, exact b end
```
```
13:23: α : Type
⊢ foo

13:36: invalid type ascription, expression has type
  _root_.foo α_1
but is expected to have type
  foo
state:
α : Type
⊢ foo
```

leodemoura:
> What is a direct use? Only parameters that are referenced in the type?

Yes, for theorems, only parameters referenced in the type are automatically included because we need to compute the precise type of the theorem before we elaborate it.

For definitions, we also include the value, but we ignore tactics. We ignore tactics because the names occurring there are only resolved at tactic execution time.

BTW, I willing to revise these rules. I'm not super happy with the way Lean handles parameters.

leodemoura:
I will fix the issue at 
```lean
def ex : foo := begin trace_state, exact b end
```

leodemoura:
@gebner @kha I think we should revise how parameters are handled in Lean after the POPL tutorial. It is too confusing (for us and for users). The Coq approach is simpler to understand, the only disadvantage is that the `end` section command would be a synchronization point. What do you think?

gebner:
> The Coq approach is simpler to understand, the only disadvantage is that the end section command would be a synchronization point. What do you think?

So we would first add the parameters as constants, parse the section normally, and then add abstractions to the declarations and replay them onto the environment from before the section?

Whether this requires waiting for the proofs depends on how we define what parameters are used by a lemma.  In my opinion, this should not depend on the proof.  And then it doesn't become a synchronization point either.

One issue that you mentioned in #554 is that almost all other environment changes get lost during the replay.  I think we could solve that problem using the modification lists I proposed in #1243: we would just need to add an `abstract_parameters` method to the modification objects.  Then we can reset the modification list at the begin of a section, and replay the abstracted changes at the end.

leodemoura:
> So we would first add the parameters as constants, parse the section normally, and then add abstractions to the declarations and replay them onto the environment from before the section?

Yes.

> Whether this requires waiting for the proofs depends on how we define what parameters are used by a lemma. In my opinion, this should not depend on the proof. And then it doesn't become a synchronization point either.

Yes, we can enforce this. 

> One issue that you mentioned in #554 is that almost all other environment changes get lost during the replay. I think we could solve that problem using the modification lists I proposed in #1243: we would just need to add an abstract_parameters method to the modification objects.

Yes, I agree. Moreover, when #554 was written we didn't have the attribute manager. It also helps.

I think this will be a good change. The current approach is too fragile, and allows the user to do weird stuff (e.g., use the `rename` tactic to rename a parameter). For example, the following example doesn't work because `b` is mapping to `foo.b α`, but `α` is not in the local context anymore after the `rename`.
```
def ex : foo :=
begin
  trace_state,
  rename α β,
  exact b
end
```
In the new approach, the user will not be able to write this.

leodemoura:
The commits above address the issues raised in this PR. 
The solution is not perfect, and we may change how parameters are implemented in Lean in the future as discussed above.

PR #1254, opened 2016-12-16
super, a superposition prover

gebner:
As discussed on slack, it's probably a good idea to include super in the main repo.  Because at least for now, that's the only way other people can actually try it out.

Notable features:
 * First-order complete  superposition prover with splitting and extensible inferences.  (`super` tactic)
 * Toy SAT solver.  (`cdcl` tactic)
 * Proof trimming to reduce size of produced proofs.
 * Makes an effort to produce constructive proofs.  (Well, this was more relevant when we still had HoTT support.)

If you have any questions, please ask.  I'm just terribly busy ATM or I'd write a longer summary.

I have tried to put most of the generally useful utility functions into the right modules, though there are still a few ones left in `tools.super`.
 * Since super heavily uses monad transformers, I have added a module for monad transformers in `data.monad.transformers`.  This also includes an explicit type-parameter preserving coercion `♯` between monads.  Unfortunately I couldn't get typeclass inference to work here, and had to declare all of the `monad_lift tactic prover_state`, etc., instances manually.
 * There is merge sort, but I didn't put it into the list module because it depends on `has_ordering` instead of `decidable_linear_order`.  Unfortunately there is almost no type that has both typeclasses, and I need it for `expr` which only has `has_ordering`.

### DIFF
--- a/library/data/monad/transformers.lean
+++ b/library/data/monad/transformers.lean
@@ -1,0 +1,40 @@
+namespace monad
+
+class monad_transformer (transformer : ∀m [monad m], Type → Type) :=
+(is_monad : ∀m [monad m], monad (transformer m))
+(monad_lift : ∀m [monad m] A, m A → transformer m A)
+
+instance transformed_monad (m t) [monad_transformer t] [monad m] : monad (t m) :=
+monad_transformer.is_monad t m
+
+class has_monad_lift (m n : Type → Type) :=
+(monad_lift : ∀A, m A → n A)
+
+instance monad_transformer_lift (t m) [monad_transformer t] [monad m] : has_monad_lift m (t m) :=
+⟨monad_transformer.monad_lift t m⟩
+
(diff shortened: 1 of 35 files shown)


gebner:
Oh, and two more things:
* There are some example files, some files include inline examples.  We should probably move them out of the library.
* Memory usage is skyrocketing--we're now at almost 2G for the whole standard library.

leodemoura:
> Memory usage is skyrocketing--we're now at almost 2G for the whole standard library.

I reduced memory consumption to 0.9G by enabling expr caching during deserialization.
To make it work, I also had to stop flushing the cache whenever we enable/disable caching.

leodemoura:
BTW, with `-j 0` the memory consumption drops to 0.7G since the expr caching is per thread.

leodemoura:
> There are some example files, some files include inline examples. We should probably move them out of the library.

I moved them to `tests/lean/run`

PR #1255, opened 2016-12-18
Refactoring bitvec

Kha:
I set out to define all signed `bitvec` operations on any `bitvec n`, as discussed in a previous comment. On top of that I did the following changes, mostly taken from [my lean2 fork](https://github.com/Kha/electrolysis/blob/master/thys/bitvec.lean) of the theory.

* (try to) apply style & conventions used in the rest of the standard library
* make parameters explicit that can only be inferred from the return type
* Add two-way conversion functions to/from `nat` and `int`. I've removed the generic unsigned conversion function, assuming that it is easier to just go through `nat`, especially in proofs.

I'm happy to discuss any of these changes. I can also move everything to `library_dev` if that makes more sense, or outright postpone the PR. This was mostly done to get a feeling for working in Lean 3.

/cc @joehendrix

### DIFF
--- a/library/data/bitvec.lean
+++ b/library/data/bitvec.lean
@@ -1,7 +1,7 @@
 /-
 Copyright (c) 2015 Joe Hendrix. All rights reserved.
 Released under Apache 2.0 license as described in the file LICENSE.
-Author: Joe Hendrix
+Authors: Joe Hendrix, Sebastian Ullrich
 
 Basic operations on bitvectors.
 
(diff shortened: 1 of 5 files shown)


library/data/bitvec.lean
```diff
@@ -74,92 +77,96 @@ end bitwise
...
 
-  -- Convert a bitvector to another number.
-  def from_bitvec [has_add α] [has_zero α] [has_one α] {n : nat} (v : bitvec n) : α :=
-  let f : α → bool → α := λr b, cond b (r + r + 1) (r + r) in
-  list.foldl f (0 : α) (to_list v)
-end from_bitvec
+  protected def ofNat : Π (n : ℕ), nat → bitvec n
+  | 0        x := nil
+  | (succ n) x := ofNat n (x / 2) ++ₜ [to_bool (x % 2 = 1)]
+
```
leodemoura, line 151: naming conventions :)

Kha: I was just about to comment on this :) . I deliberately decided to go against the convention in my own project for these definitions, because having functions containing `of_` is just confusing when proving things about them. With that being said, now is probably not the time to reopen that discussion, so I'm changing it right now.

PR #1257, opened 2016-12-18
Various interaction enhancements

Kha:


src/library/tactic/tactic_state.cpp
```diff
@@ -618,7 +618,10 @@ vm_obj tactic_add_decl(vm_obj const & _d, vm_obj const & _s) {
 
 vm_obj tactic_opened_namespaces(vm_obj const & s) {
```
Kha, line 619: Perhaps `open_namespaces` would be a more appropriate name.

leodemoura: I agree. I will fix it.

Issue #1262, opened 2016-12-22
error: file 'data/nat' not found in the LEAN_PATH & segfault


Hodapp87:
I followed [CMake & Make](https://github.com/leanprover/lean/blob/master/doc/make/cmake_make.md) build instructions verbatim for latest master on two different systems (one Arch, one Debian), followed by `make install`. I am trying to run just the second example in the [Short tutorial](https://github.com/leanprover/lean/blob/master/doc/lean/tutorial.org), that is, I've put the following into `test.lean`:

```lean
import data.nat
open nat
-- defines the double function
definition double (x : nat) := x + x
```

and then tried to run with `lean test.lean`. This simply gives me:

```
/home/hodapp/source/lean/test.lean:1:0: error: file 'data/nat' not found in the LEAN_PATH
Segmentation fault
```

Setting `LEAN_PATH` to `/usr/local/lib/lean/library` (what is it even supposed to be?) doesn't change anything. Running `leanemacs` and merely putting an `import` produces a similar error both in the editor and if I attempt to compile.

Kha:
The short tutorial has not been updated in quite a while, we should probably remove it for now (at least from the root readme). Please see the longer [Theorem Proving in Lean](https://leanprover.github.io/theorem_proving_in_lean/index.html) instead, which should be up-to-date with the `master` branch.

leodemoura:
> The short tutorial has not been updated in quite a while, we should probably remove it for now (at least from the root readme).

@kha I agree.

PR #1263, opened 2016-12-22
Avoid segfault on missing import file

Kha:
As reported in #1262

Note that a missing import will also prevent all other imports.

### DIFF
--- a/src/library/module_mgr.cpp
+++ b/src/library/module_mgr.cpp
@@ -261,19 +261,17 @@ void module_mgr::build_module(module_id const & id, bool can_use_olean, name_set
             mod->m_source = module_src::LEAN;
             mod->m_trans_mtime = mod->m_mtime = mtime;
             for (auto & d : imports) {
-                module_id d_id;
-                std::shared_ptr<module_info const> d_mod;
                 try {
-                    d_id = resolve(id, d);
+                    module_id d_id = resolve(id, d);
                     build_module(d_id, true, module_stack);
-                    d_mod = m_modules[d_id];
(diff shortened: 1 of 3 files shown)


leodemoura:
The new test is failing for me in Release and Debug mode.
It is producing the following output:
```
% lean -j 0 missing_import.lean 
/home/leo/projects/lean/tests/lean/missing_import.lean:1:0: error: file 'does/not/exist' not found in the LEAN_PATH
/home/leo/projects/lean/tests/lean/missing_import.lean:1:0: error: file 'does/not/exist' not found in the LEAN_PATH
<unknown>:1:1: error: file 'does/not/exist' not found in the LEAN_PATH
```
The test scripts produces:
```
% ./test_single.sh ../../bin/lean missing_import.lean
-- testing missing_import.lean
2,3c2
< missing_import.lean:1:0: error: file 'does/not/exist' not found in the LEAN_PATH
< <unknown>:1:1: error: file 'does/not/exist' not found in the LEAN_PATH
---
> missing_import.lean:1:0: error: invalid import
ERROR: file missing_import.lean.produced.out does not match missing_import.lean.expected.out
```

Kha:
Sorry, it seems I forgot a rebuild.

gebner:
The intention behind the empty `m_mod_info` field is to model dependencies that do not exist (yet), such as the `does.not.exist` import in the example.

Keeping around information about imports that cannot be resolved is important because we can then reparse that file as soon as the dependency is created. (Well, this does not work yet.  But it's something we should do, and that is on my todo list).

Kha:
Superseded by #1264

PR #1264, opened 2016-12-23
fix(module_mgr): guard all accesses to m_mod_info

### LINKED ISSUE #1262: error: file 'data/nat' not found in the LEAN_PATH & segfault

gebner:
Fixes #1262.  This is a version of PR #1263 that keeps around information about imports that do not exist yet.

### DIFF
--- a/src/frontends/lean/parser.cpp
+++ b/src/frontends/lean/parser.cpp
@@ -2166,16 +2166,23 @@ void parser::process_imports() {
     unsigned fingerprint = 0;
     auto imports = parse_imports(fingerprint);
 
-    try {
-        m_env = import_modules(m_env, m_file_name, imports, m_import_fn);
-    } catch (exception & ex) {
+    buffer<import_error> import_errors;
+    m_env = import_modules(m_env, m_file_name, imports, m_import_fn, import_errors);
+
+    if (!import_errors.empty()) {
         m_found_errors = true;
-        parser_exception error((sstream() << "invalid import").str(),
-                               m_file_name.c_str(), m_last_cmd_pos.first, m_last_cmd_pos.second);
-        if (!m_use_exceptions && m_show_errors)
-            report_message(error);
-        if (m_use_exceptions)
-            throw error;
+        for (auto & e : import_errors) {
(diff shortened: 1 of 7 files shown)


gebner:
The parser now also continues after failed imports.  Thanks @Kha for noticing this, I've also used the testcase from your PR.

gebner:
Now the `missing_import` test fails on windows (appveyor).  But I can't reproduce it locally as the test_single.sh scripts completely fails for me:
```
gebner@NILOTICA MINGW64 ~/lean/tests/lean
$ ./test_single.sh ~/lean/bin/lean missing_import.lean
-- testing missing_import.lean
1,2c1,2
< D:\msys64\home\gebner\lean\tests\lean\missing_import.lean:1:0: error: file 'does/not/exist' not found in the LEAN_PATH
< D:\msys64\home\gebner\lean\tests\lean\missing_import.lean:1:0: error: invalid import: does.not.exist
---
> missing_import.lean:1:0: error: file 'does/not/exist' not found in the LEAN_PATH
> missing_import.lean:1:0: error: invalid import: does.not.exist
ERROR: file missing_import.lean.produced.out does not match missing_import.lean.expected.out
```

PR #1265, opened 2016-12-23
fix(library/app_builder.cpp): need custom congr_fun and congr so we c…

dselsam:
…an use relaxed_whnf

### DIFF
--- a/src/library/app_builder.cpp
+++ b/src/library/app_builder.cpp
@@ -754,6 +754,53 @@ public:
         level lvl_2  = get_level(B);
         return ::lean::mk_app({mk_constant(get_congr_arg_name(), {lvl_1, lvl_2}), A, B, lhs, rhs, f, H});
     }
+
+    expr mk_congr_fun(expr const & H, expr const & a) {
+        expr eq = m_ctx.relaxed_whnf(m_ctx.infer(H));
+        expr pi, lhs, rhs;
+        if (!is_eq(eq, pi, lhs, rhs)) {
+            lean_app_builder_trace(tout() << "failed to build congr_fun, equality expected:\n" << eq << "\n";);
+            throw app_builder_exception();
+        }
+        pi = m_ctx.relaxed_whnf(pi);
+        if (!is_pi(pi)) {
+            lean_app_builder_trace(tout() << "failed to build congr_fun, function expected:\n" << pi << "\n";);
+            throw app_builder_exception();
+        }
+        expr A = binding_domain(pi);
+        expr B = mk_lambda(binding_name(pi), binding_domain(pi), binding_body(pi), binding_info(pi));
+
+        level lvl_1  = get_level(A);
+        level lvl_2  = get_level(::lean::mk_app(B, a));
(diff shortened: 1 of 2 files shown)


src/library/app_builder.cpp
```diff
@@ -754,6 +754,54 @@ class app_builder {
         level lvl_2  = get_level(B);
         return ::lean::mk_app({mk_constant(get_congr_arg_name(), {lvl_1, lvl_2}), A, B, lhs, rhs, f, H});
     }
+
+    expr mk_congr_fun(expr const & H, expr const & a) {
+        expr eq = m_ctx.relaxed_whnf(m_ctx.infer(H));
+        expr arg = m_ctx.relaxed_whnf(m_ctx.infer(a));
+        expr pi, lhs, rhs;
```
leodemoura, line 760: `arg` is not use. Moreover, this line may produce timeouts (Example: `arg` is `10000 + 10000`).

src/library/app_builder.cpp
```diff
@@ -754,6 +754,54 @@ class app_builder {
...
+        expr eq1 = m_ctx.relaxed_whnf(m_ctx.infer(H1));
+        expr eq2 = m_ctx.relaxed_whnf(m_ctx.infer(H2));
+        expr pi, lhs1, rhs1;
+        if (!is_eq(eq1, pi, lhs1, rhs1)) {
+            lean_app_builder_trace(tout() << "failed to build congr, equality expected:\n" << eq1 << "\n";);
+            throw app_builder_exception();
+        }
+        expr ignore, lhs2, rhs2;
+        if (!is_eq(eq2, ignore, lhs2, rhs2)) {
+            lean_app_builder_trace(tout() << "failed to build congr, equality expected:\n" << eq2 << "\n";);
```
leodemoura, line 789: We have an `is_eq` that extracts only the `lhs` and `rhs`.

leodemoura:
Is there a better solution for this kind issue?
Similar issues will happen with other procedures in the `app_builder` module (e.g., `mk_funext`).

leodemoura:
We mainly use `app_builder` in contexts where it is expected to work (e.g., for building proof terms).
However, the `app_builder` is currently using the reducibility settings from the `type_context` object provided by the caller. This is the source of the problem.

Here is a plan: 
- Add an extra argument `optional<transparency_mode>` in the `app_builder` functions. This argument is none by default.
- If the new argument is not specified, the app_builder overrides the transparency mode to `Semireducible`.

dselsam:
In your proposal, the simplifier would need to call the app-builder methods with `transparency.all`, or else we could trigger the same issue (in principle, if we did not write custom versions for everything) by setting `alpha` to be irreducible.

leodemoura:
> In your proposal, the simplifier would need to call the app-builder methods with transparency.all, or else we could trigger the same issue (in principle, if we did not write custom versions for everything) by setting alpha to be irreducible.

If you marked `alpha` as `irreducible`, then it is morally not a `Pi`. 
BTW, the type in your example doesn't even type check if you mark `alpha` as irreducible.

dselsam:
Okay. I have never used `irreducible` and I don't remember which modules do what with it when. I agree that building proofs with `semi` by default is the way to go.

dselsam:
Any reason not to merge the original commit? The more efficient version was already a TODO item in the code.

PR #1269, opened 2016-12-25
inform compiler about allocator

diakopter:
fixes (allegedly spurious) build errors on macosx with gperftools


diakopter:
well, it doesn't fix the errors, but the flags are otherwise recommended by gperftools

diakopter:
sigh; accidentally removed the commit

PR #1270, opened 2016-12-26
More accurate `info` command using server-side parsing

Kha:
Currently, the client editor is responsible for finding the correct source location for `completion` and `info` commands. In practice, this means that lean-emacs had to reimplement parts of the Lean lexer. This commit shifts the burden to the server side by transmitting the actual cursor location and letting the server discover the context by use of the actual Lean parser, starting with the `info` command.

In addition to making the API much easier to use, this commit also uses the `parser` integration to make `info` output work on any charactor of any token of a notation.

Some remaining issues:
* `info` triggers a "still running" message, even though it is a "tiny" task. I believe this is due to the `cur_task` response field, but I'm not quite sure what that field actually signifies
* Messages from the reparsing are not ignored yet, resulting in duplicate messages


gebner:
> info triggers a "still running" message, even though it is a "tiny" task. I believe this is due to the cur_task response field, but I'm not quite sure what that field actually signifies

The original intention behind the tiny flag was to hide tasks from the progress output on the command line, in order to reduce the amount of printing that we do.  Since then, I have changed the command line output to just sample the current task every 100ms, so we don't need the tiny flag there anymore.  As you have seen however, it is used by the server to decide which tasks to include in the messages (again in order to reduce the amount of traffic, we also only return tasks for "open" files).

I personally don't see a big problem in showing lean as working while we're doing auto-completion at the moment.  We might want to rework the progress/parsing-status display anyhow, to better show which parts of a file are already completely checked (which is something that I kind of miss from Isabelle & co.).

> Messages from the reparsing are not ignored yet, resulting in duplicate messages

This should do the trick:
```c++
null_message_buffer null_msg_buf;
scoped_message_buffer scope(&null_msg_buf);
```

gebner:
I just noticed one more thing: this will call the elaborator on all pre-expressions since the last snapshot, right?

If that's the case we need to cancel old tasks (+ elaboration tasks, etc.) when we get a new completion request, otherwise we'll drown in now-useless parsing tasks.  Another (additional) option is to disable elaboration (or just tactics execution) during auto-completion.

Kha:
> The original intention behind the tiny flag was to hide tasks from the progress output on the command line, in order to reduce the amount of printing that we do. Since then, I have changed the command line output to just sample the current task every 100ms, so we don't need the tiny flag there anymore.

Right, I forgot about the cmdline output. Do we need `cur_task` for the editor clients, though?

> I personally don't see a big problem in showing lean as working while we're doing auto-completion at the moment.

It is quite annoying with `info`, which is triggered on any movement.



> This should do the trick:

Nice, thanks.

> I just noticed one more thing: this will call the elaborator on all pre-expressions since the last snapshot, right?

Yes. I've pushed a preliminary fix that deactivates definition elaboration during `info`. There is no noticable lag now navigating `super_examples.lean` - at least as soon as the file is fully elaborated. Before that, you notice the `info_task` waiting to be scheduled.

gebner:
> Right, I forgot about the cmdline output. Do we need cur_task for the editor clients, though?

At the moment, emacs shows both the tasks that still need to be done in the current file (to give a sense of what's missing), and also (one of) the currently running tasks (so that you know what's happening when lean is recompiling the whole standard library).  The `cur_task` field is used for the second one.

BTW, I think you can drop the "HACK" now.

> It is quite annoying with info, which is triggered on any movement.

I see.  As a quick fix, we could simply ignore all tiny tasks in the server--then you shouldn't notice anything in emacs.

> Yes. I've pushed a preliminary fix that deactivates definition elaboration during info. There is no noticable lag now navigating super_examples.lean - at least as soon as the file is fully elaborated. Before that, you notice the info_task waiting to be scheduled.

Excellent!  Nevertheless we should still cancel the old info task, just in case.  (For example, if we are doing auto-completion in/after a print command we might end up waiting for a proof to elaborate.)  You can just keep the generic_task_result that you get when submitting, and then call `.cancel()` on it.  (You can even call this method if it's null.)

Kha:
> BTW, I think you can drop the "HACK" now.

Right, about that one. The problem is [this assertion](https://github.com/leanprover/lean/blob/f719991c719a011e2ac796bf7e4b9ec084b61fe9/src/library/message_buffer.cpp#L75).


> You can just keep the generic_task_result that you get when submitting, and then call .cancel() on it. (You can even call this method if it's null.)

Nice, that was easy enough! I'm still not too familiar with the parallelism additions to the server.

Kha:
The current goal is now reported at any position inside a tactics proof (it's the goal state prior to executing the current tactic step). Any other requests for making `info` smarter :) ? If not, I will move on to autocompletion.

gebner:
> > BTW, I think you can drop the "HACK" now.

> Right, about that one. The problem is this assertion.

Oh, I see.  Then let's keep the hack for now.  I'd like to refactor the task infrastructure a bit anyhow, now that we know a bit better what we actually want from it (somewhere between POPL and ITP).

PR #1276, opened 2017-01-02
fix(emacs): bug in suggested Emacs init

kim-em:
The original code set the variable `lean-required-packages`, but then expected `lean-mode-required-packages`.

I've also added a hint about where to find your emacs init file; this was something I had to learn to follow these instructions.

A question for follow-up: the requirements sections states that `fill-column-indicator` is a dependency, but the suggested init file does not attempt to install this. Should this be fixed?

### DIFF
--- a/src/emacs/README.md
+++ b/src/emacs/README.md
@@ -65,13 +65,13 @@ Installation
 ============
 
 You can also include lean-mode permanently in your emacs init file.  In this
-case, just put the following code in your Emacs init file:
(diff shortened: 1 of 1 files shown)


gebner:
> A question for follow-up: the requirements sections states that fill-column-indicator is a dependency, but the suggested init file does not attempt to install this. Should this be fixed?

Thanks for noticing this discrepancy!  We had an optional dependency on `fill-column-indicator` for some time, but a few weeks ago we removed it completely.  It looks like we forgot to remove it here.

PR #1281, opened 2017-01-04
Fix potential deadlocks with lazy values

gebner:
See discussion at https://github.com/leanprover/lean/commit/73bf232feae84bcdb493be21b45b4e9d1d81bcba


gebner:
@leodemoura Do we need to generate VM code for lemmas?  When adding lemmas from tactics, we unconditionally invoke the VM compiler (which causes lean to wait for the proofs).

leodemoura:
> @leodemoura Do we need to generate VM code for lemmas? 

No, we don't.

> When adding lemmas from tactics, we unconditionally invoke the VM compiler (which causes lean to wait for the proofs).

Are you talking about `tactic.add_decl`? 
This one is always generating VM code, even if the new declaration is a lemma.
This is bad, and should be fixed.

gebner:
> Are you talking about tactic.add_decl?

Yes, exactly this one.  I have now put the check for theorem/noncomputable/built-in in the `vm_compile` function itself, then we don't need to duplicate them everywhere.

> This one is always generating VM code, even if the new declaration is a lemma.

I suspected as much, but wasn't quite sure what we did with the following:
```lean
lemma my_dearest_number : ℕ := 0
```

leodemoura:
> I suspected as much, but wasn't quite sure what we did with the following:

One option is to enforce in the front-end that the type of a `lemma` is a proposition.
It solves this problem, and it is also a nice sanity checking for users.
What do you think?

gebner:
> One option is to enforce in the front-end that the type of a lemma is a proposition.

I'm not sure that will work.  Sometimes we want to prove statements that are other types than propositions, but still do not care about the proofs.  Looking through the standard library, some lemmas about equality are like that: `heq.elim` for example.

I think that the distinction between the theorem and definition commands is more useful to mark the values as delayed/irrelevant and immediate/compiled/relevant, resp.

leodemoura:
> I think that the distinction between the theorem and definition commands is more useful to mark the values as delayed/irrelevant and immediate/compiled/relevant, resp.

We can keep the code as is. In the worst case, the user will get an error `failed to generate code for 'boo' because there is no code for 'my_dearest_number'` if we use `my_dearest_number` above to define `boo`.

Another option is to refine the check at 

https://github.com/gebner/lean/blob/78dcdc508a400300b301b2eb85c79a61159c22d9/src/library/compiler/vm_compiler.cpp#L403

by replacing `d.is_theorem()` with `d.is_theorem() && is_prop(env, d)`,  where `is_prop` is a new function that returns true iff the type of `d` in `env` is a proposition.

PR #1282, opened 2017-01-05
Check whether proofs are correct

gebner:
If you have a lemma with a wrong proof, you will get tons of error messages, but the lean command-line tool will still exit with 0.

I also added two tests that check for the expected non-zero exit value.


gebner:
We even missed a broken test: the converter test.

leodemoura:
@gebner After this PR, I noticed a strange behavior in the debug build.
The command `lean --make` will compile a few files after a successful `lean --make` invocation.
Only after 3 `lean --make` invocations, it behaves as expected. I can consistently reproduce this behavior.

BTW, I cannot reproduce the problem with the release build.

gebner:
@leodemoura Thanks for noticing this again!  Should be fixed with #1286.

leodemoura:
It fixed the problem. Thanks!

PR #1287, opened 2017-01-06
Complete all the things

Kha:


leodemoura:
This is nice!

The "complete imports" doesn't seen to work for me.

leodemoura:
Here is a potential improvement: do not include internal declarations in the autocompletion.
In Lean, we say a declaration is internal when its name contains a fragment starting with `_`. Examples: `f._match_1`, `_nest2.ir.expr`, etc.
The frontend doesn't allow users to create this kind of declaration.

Kha:
Right, seems I broke things at some point. I'll look into it.

I also noticed that private definitions are listed. Any other definition kinds we should ignore on top of that?

leodemoura:
> I also noticed that private definitions are listed. Any other definition kinds we should ignore on top of that?

I noticed that internal definitions such as `foo.equations.eqn_1` are being listed too. 
I will fix this by renaming them to `foo._equations._eqn_1`, and make it clear they are internal.

There is also the issue regarding projections that you raised a long time ago. 
We don't get auto completion for them. We filter them to avoid a tsunami of cases from the algebraic hierarchy. I think we can minimize the problem by filtering only projections from type classes.
https://github.com/leodemoura/lean/commit/ca9e24ba7d95d82b17613e7484ec11fff0ec36e8
What do you think?

Kha:
> I think we can minimize the problem by filtering only projections from type classes.

That sounds reasonable. Alternatively, would the problem be solved by ignoring inherited projections? I guess we would have to add some more bookkeeping for that.

leodemoura:
> That sounds reasonable. Alternatively, would the problem be solved by ignoring inherited projections? I guess we would have to add some more bookkeeping for that.

Yes, it would avoid the tsunami of completions.
However, users might still want to use the inherited one for performance reasons.
For example, suppose we have 
```lean
structure A :=
(x : nat) ... -- many more fields

structure B extends A :=
(z : nat) ...
```

Given (b : B), to use `A.x b`, we need to coerce `b` into `A`. So, `A.x b` would be elaborated into `A.x (B_to_A b)`, where `B_to_A b` expands into `A.mk (B.x b) ...`.

> I guess we would have to add some more bookkeeping for that.

Yes, we would have to store this information somewhere.

Kha:
I see. Ignoring class projections sounds fine then.

PR #1288, opened 2017-01-06
fix(emacs/lean-server): disable undo in process buffer

gebner:
@leodemoura I believe this should fix the undo message you're seeing.  We don't
need undo in the process buffer anyway; this might save a bit of memory as well.

### DIFF
--- a/src/emacs/lean-server.el
+++ b/src/emacs/lean-server.el
@@ -93,7 +93,7 @@
          (proc (if (fboundp 'make-process)
                    (make-process ;; emacs >= 25 lets us redirect stderr
                     :name "lean-server"
-                    :buffer (format "*lean-server (%s)*" project-dir)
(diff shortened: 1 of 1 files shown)


Kha:
We should probably adjust the process buffer names:

> Buffers that are ephemeral and generally uninteresting to the user have names starting with a space, so that the list-buffers and buffer-menu commands don’t mention them (but if such a buffer visits a file, it is mentioned). A name starting with space also initially disables recording undo information; see Undo.

https://www.gnu.org/software/emacs/manual/html_node/elisp/Buffer-Names.html

gebner:
I'm not sure I like hiding the process buffers.  Previously it was quite clear what lean processes were running, this information is completely inaccessible now.

Kha:
Right. The buffer list may not be the best location to display that
information, but it's better than nothing.

PR #1300, opened 2017-01-11
info: provide more metadata

Kha:
* docs: attributes, options, identifiers
* location: attributes, imports


leodemoura:
This is really cool :)

PR #1307, opened 2017-01-12
fix(emacs/lean-flycheck): only update next-error-mode for current buffer

### LINKED ISSUE #1306: [emacs] next-error-mode doesn't work if multiple buffers are open
We add a flycheck hook to run `lean-next-error-copy` on flycheck changes.  However, with multiple buffers open, each of the open buffers gets flycheck updates when the server sends new messages.  Hence the next-error-mode flickers between error messages from several buffers.

gebner:
Fixes #1306.


Kha:
:+1:

PR #1309, opened 2017-01-13
Move lean.js to server mode

Kha:


gebner:
Awesome!

For reference, here is the corresponding PR for mkleanbook: leanprover/mkleanbook#2.

gebner:
I've been excitedly playing with it for last few minutes, it's really nice.  BTW, at the moment traces go to stderr (i.e., console.log) and are not captured as messages.  To fix that just set the `trace.as_messages` option:
```c++
            opts = opts.update(name {"trace", "as_messages"}, true);
```

Kha:
Oh I see, that initialization was kind of hidden :) . Fixed and deployed,
thanks.

PR #1316, opened 2017-01-17
fix(library/vm/vm): fix segfault in call_stack_fn

gebner:


gebner:
@leodemoura This produces the ugly stack trace that we talked about yesterday, it seems that we do not inline `tactic_bind` (even though it is `@[inline]`) because it is not fully applied:
```lean
open tactic
run_command (do
  x ← return 5,
  x ← return 5,
  x ← return 5,
  x ← return 5,
  x ← return 5,
  x ← return 5,
  x ← return 5,
  x ← return 5,
  x ← return 5,
  x ← return 5,
  trace_call_stack)
```

PR #1317, opened 2017-01-17
[RFC] Highlight pending tasks in emacs

gebner:
This highlights the running tasks with a different background color instead of including them in the flycheck error list.  There is now also an hourglass/checkmark indicator next to the Lean mode to indicate server status.  For example, when a `lemma` is being elaborated you have a nice orange background from `lemma` to the end of the proof.

How do you like the color?  It didn't look bad at first with the solarized theme, but it might get annoying after time, or really clash with other backgrounds.

### DIFF
--- a/src/emacs/lean-flycheck.el
+++ b/src/emacs/lean-flycheck.el
@@ -17,34 +17,6 @@
    (flycheck-mode (flycheck-mode -1))
    (t             (flycheck-mode  1))))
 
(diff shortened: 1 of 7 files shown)


leodemoura:
> How do you like the color? It didn't look bad at first with the solarized theme, but it might get annoying after time, or really clash with other backgrounds.

I like the color, but I use a light solarized theme.
What about an Emacs configuration option for setting the color?

PR #1319, opened 2017-01-17
Improve goal display

Kha:
foo

`C-c C-g` and `C-c C-n` now each toggle the visibility of a separate, auto-updating info buffer (just like `C-c ! l`). The buffers stop updating when they are invisible; there is no more need for a special minor mode like `lean-next-error-mode`. I'd welcome feedback about how you use the buffers and whether we should have a nice default layout like the one shown above.
Also, `,` in begin-end blocks will now show the following tactic state.


leodemoura:
This is awesome!

PR #1328, opened 2017-01-21
feat(library/vm/vm_aux): add vm code for sorry

gebner:
When we are going to have error recovery in the parser where we replace parts with sorry, it will be nice to use these half-broken functions in VM code.

### DIFF
--- a/library/init/util.lean
+++ b/library/init/util.lean
@@ -17,3 +17,7 @@ f ()
 /- This function has a native implementation that shows the VM call stack. -/
 def trace_call_stack {α : Type} (f : unit → α) : α :=
 f ()
+
+meta constant {u} undefined_core {α : Type u} (message : string) : α
+
+meta def {u} undefined {α : Type u} : α := undefined_core "undefined"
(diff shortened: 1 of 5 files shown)


gebner:
I don't think this decreases soundness, we can implement segfaults using `sorry` even without this change:
```lean
lemma they_are_isomorphic : (ℕ × ℕ) = ℕ := sorry

def repack (x : ℕ) : ℕ × ℕ :=
begin rw they_are_isomorphic, apply x end

def proj_nat := prod.fst ∘ repack

vm_eval proj_nat 4
```

gebner:
You can even do it more sneakily without `sorry`:
```lean
meta def they_are_isomorphic : unit → (ℕ × ℕ) = ℕ
| () := they_are_isomorphic ()

meta def repack (x : ℕ) : ℕ × ℕ :=
begin rw (they_are_isomorphic ()), apply x end

meta def proj_nat := prod.fst ∘ repack

vm_eval proj_nat 4
```

jroesch:
It would be cool if we could implement a deferred type error mechanism in this style, maybe something with the idealized signature: 
`sorry_with_message {T : Type} : string -> T`?

gebner:
That's what the `undefined_core` is for.  The `error` name that is used in Haskell seemed to precious to be used for a meta constant.

gebner:
BTW, better names for `undefined` and `undefined_core` are definitely welcome.

PR #1330, opened 2017-01-23
Re-use lean kernel as small type-checker

gebner:
In addition to the Haskell reference type checker, this PR adds a type-checker based on the Lean kernel.  None of the library/frontend/server code is included, considerably reducing the TCB.  Merging this PR makes travis do a full type-check of the whole library on every build.

- [x] I didn't really get the existing text-based output format: it has multiple modes (all/default), outputs module imports (not sure how they are resolved), and supports mutually inductive datatypes.  As a quick mockup, I just reused the binary serialization.  @dselsam Can we remove imports and mutual inductive datatypes from the text export, and always export all definitions?   I did not want to change anything there, so that we don't break the Haskell type-checker.
- [x] The bare-bones kernel still depends on sexpr, and transitively on gmp.
- [x] The pretty-printer heavily depends on the library, so I think we need to implement a separate one to show axioms, definitions and theorem statements.
- [x] The type checker doesn't work on emscripten at the moment.


Kha:
Time to revive my Rust type-checker project, I guess :smile: . I stopped at some (very early) point because I was unsure of the status of both the textual export format and the Haskell implementation.

gebner:
@Kha Did you do any of the optimizations where you cache extra information for each expression, e.g. the upper bound of the de Bruijn indices?  I tried the following, but it doesn't look very beautiful:

```rust
#[derive(Clone,Debug)]
pub struct Expr {
    var_bnd: u32,
    ln_bnd: u32,
    core: Arc<ExprCore>,
}

#[derive(Debug)]
pub enum ExprCore {
    Var(u32),
    Sort(Level),
    Const(Arc<Name>, Vec<Level>),
    LocalConst(LocalName, Binding),
    App(Expr, Expr),
    Lam(Binding, Expr),
    Pi(Binding, Expr),
    Let(Binding, Expr, Expr),
}
```

gebner:
Ok, I just tried the Haskell type-checker and it doesn't work with the current format anyhow.  I'll refactor the text format, and use it in the `leanchecker` executable.  Then we have one single export for all the type checkers (may there be as many as in metamath).

Kha:
> Did you do any of the optimizations where you cache extra information for each expression, e.g. the upper bound of the de Bruijn indices?

Yeah (save for the part where I actually ever executed the code): https://github.com/Kha/leantc-rs/blob/master/src/expr.rs#L8-L30

At least from the public API, using a newtype around the Arc/Rc that derefs into the enum didn't seem too bad.

gebner:
Putting the Rc around the cached cell and wrapping it with a newtype is nice.

gebner:
So I've refactored the text output and made leanchecker parse text files.
 * The `|` separators are gone from definitions and axioms.
 * The inductive declaration is now just a single line.

leodemoura:
> The bare-bones kernel still depends on sexpr, and transitively on gmp.

We could try to remove the `MPZ` and `MPQ` kinds from `sexpr`.

leodemoura:
> Merging this PR makes travis do a full type-check of the whole library on every build.

This is great.

leodemoura:
@Kha 
> Time to revive my Rust type-checker project, I guess :smile: .

This would be super cool.

dselsam:
@gebner 
> I'll refactor the text format, and use it in the leanchecker executable. Then we have one single export for all the type checkers (may there be as many as in metamath).

This is good, adjust the format however you need and I'll update the Haskell type checker accordingly.

gebner:
@dselsam I've already started in tc#7.  Should we expand let expressions when parsing, or add them to the type-checker?  The reason I added them to the export was to enable pretty-printing of definitions in the future.

dselsam:
@gebner I think for the Haskell TC, pretty-printing `let` is way more important than the code savings of not handling them. I will add support for them (once the format is stable).

gebner:
I've now added a small pretty-printer to leanchecker.  Right now it prints by default all axioms and the theorem statements that are specified as command-line arguments:
```
[gebner:~/lean/library] leanchecker+* ± time ~/lean/build/release/checker/leanchecker export.out nat.add_assoc
axiom quot.sound : Π α : Type u, Π s : setoid α, Π a b : α, setoid.r a b -> eq (quot.mk a) (quot.mk b)
axiom quot.lift : Π α : Type u, Π β : Type v, Π s : setoid α, Π f : α -> β, (Π a b : α, setoid.r a b -> eq (f a) (f b)) -> quot s -> β
axiom classical.strong_indefinite_description : Π a : Type u, Π p : a -> Prop, nonempty a -> subtype (λ x : a, Exists (λ y : a, p y) -> p x)
axiom quot : Π α : Type u, setoid α -> Type u
axiom quot.ind : Π α : Type u, Π s : setoid α, Π β : quot s -> Prop, (Π a : α, β (quot.mk a)) -> (Π q : quot s, β q)
axiom quot.mk : Π α : Type u, Π s : setoid α, α -> quot s
axiom propext : Π a b : Prop, iff a b -> eq a b
theorem nat.add_assoc : Π n m k : nat, eq (add (add n m) k) (add n (add m k))
checked 6933 declarations
16.04user 0.28system 0:16.34elapsed 99%CPU (0avgtext+0avgdata 717216maxresident)k
0inputs+0outputs (0major+195204minor)pagefaults 0swaps
```
Any suggestions for further info?

leodemoura:
@gebner It is debatable whether the following ones should be listed as axioms: `quot`, `quot.mk`, `quot.lift`,  and `quot.ind`. They are the type, constructor and eliminators for the quotient type. Note that we do not list inductive datatypes, their constructors and eliminators as axioms. Moreover, `quot/quot.mk/quot.lift/quot.ind` do no block reduction as axioms such as `propext` and `quot.sound`.

library/system/io.lean
```diff
@@ -3,17 +3,17 @@ Copyright (c) 2016 Microsoft Corporation. All rights reserved.
 Released under Apache 2.0 license as described in the file LICENSE.
 Authors: Luke Nelson, Jared Roesch and Leonardo de Moura
 -/
-constant {u} io : Type u → Type u
-constant io.functor : functor io
-constant io.monad   : monad io
+meta constant {u} io : Type u → Type u
+meta constant io.functor : functor io
```
leodemoura: It is not clear to me whether we want `io` to be `meta`. 
On one hand, it makes senses, we cannot prove anything about programs using `io`, and it is wrapping C++ code.
On the other hand, it will be awkward to write a function in `io` and tag it as `meta`. Suppose we are writing a program, and its interface is implemented using `io`. It feels awkward to say the interface is `meta` (-programming). Perhaps, we need a different keyword (e.g., `untrusted`). What do you think?

gebner: The reason I put them into meta is so that they we skip them during export, and they don't show up as additional axioms.

Personally, when I think of the meta keyword, I don't jump to "meta-programming", but rather think of the relaxed type checking requirements it entails in Lean.  You're right in that the meaning is closer to `untrusted`, and that `meta` might potentially confuse someone.

Nevertheless, I think we should cleanly separate the logically relevant and trusted parts of lean from the potentially unsound parts which are only meant to be executed.  I don't care whether we call them trusted/untrusted, regular/meta, safe/unsafe, total/partial, or whatever.  (The current meta name is the catchiest, I think, and also the one I've gotten used to.)  Whichever way we decide, it would be a good idea to have a single nomenclature everywhere (right now we use untrusted in C++, and meta in Lean code).

gebner:
> It is debatable whether the following ones should be listed as axioms: quot, quot.mk, quot.lift, and quot.ind.

The whole point of a proof checker is to give assurance that a theorem is correct.  From that point of view it seems vital to list all used axioms and constants (and probably the definitions that are used in the theorem statement as well).  More to the point, the quotient extension doesn't check the type of the provided axioms (or of `setoid`).  If you make a tiny mistake, you can easily derive `false` using the quotient extension:
```lean
prelude
import init.data.setoid

universe variables u v

constant quot : Π {α : Type u}, setoid α → Type u

namespace quot
protected constant mk        : Π {α : Type u}   [s : setoid α], α → quot s
notation `⟦`:max a `⟧`:0 := quot.mk a

constant sound     : Π {α : Type u} [s : setoid α] {a b : α}, a ≈ b → ⟦a⟧ = ⟦b⟧
constant lift      : Π {α : Type u} {β : Type v} [s : setoid α] (f : α → β), true → quot s → β
constant ind       : ∀ {α : Type u} [s : setoid α] {β : quot s → Prop}, (∀ a, β ⟦a⟧) → ∀ q, β q

attribute [elab_as_eliminator] lift ind

init_quotient

end quot

instance my_setoid : setoid bool :=
{ r := λx y, true,
  iseqv := ⟨(λ_, ⟨⟩),
            (λ_ _ _, ⟨⟩),
            (λ_ _ _ _ _, ⟨⟩)⟩ }

lemma lift_tt_eq_ff : ⟦tt⟧ = ⟦ff⟧ :=
quot.sound ⟨⟩

def strange_function : quot my_setoid → bool :=
quot.lift id ⟨⟩

eval strange_function ⟦tt⟧
eval strange_function ⟦ff⟧

lemma strange_prop : tt = ff :=
congr_arg strange_function lift_tt_eq_ff

lemma useful : false :=
bool.no_confusion strange_prop

print axioms useful
```

> Note that we do not list inductive datatypes, their constructors and eliminators as axioms.

I've been wondering why leanchecker doesn't list them.  Does `is_axiom()` return false for them?

leodemoura:
> If you make a tiny mistake, you can easily derive a contradiction with the quotient extension

Regular users cannot make this mistake. `quot` is defined in the `init` folder which was written by us.
By the same token, if we make a tiny mistake at `type_checker.cpp`, we can derive false.

We can easily protect against users that want to write their own `prelude` and define their own `quot` type. If you feel strongly about this, then adding the extra checks at `quot.cpp` is the way to go.
Another option is to declare `quot`, `quot.mk`, `quot.ind` and `quot.lift` at `quot.cpp`. We defined them in `quot.lean` just because it was more convenient. We can also remove the dependency from `setoid`.
There are different ways to protect against the kind of user abuse you have described.
That being said, I still think we should *not* list `quot`, `quot.mk`, `quot.ind` and `quot.lift` as axioms.

> I've been wondering why leanchecker doesn't list them. Does is_axiom() return false for them?

The environment provides an `is_builtin` method.

library/system/io.lean
leodemoura: Here are different scenarios for `meta/untrusted`.

1- I want to use unbounded recursion when defining `foo`. In this scenario, I cannot prove anything about `foo`. This is fine if `foo` is a function used for meta-programming (e.g., it is a tactic). 

2- We add a (meta-) primitive `foo` that breaks referential transparency. We do it because it is convenient. We really want to mark `foo` as `meta` because it doesn't even make sense to prove any property about `foo`.

3- I want to use some primitive `foo` available in the OS. I manually axiomatize properties of this primitive. We may even execute the code based on `foo` by using the `ffi` infrastructure (when it becomes stable). `foo` and its properties are declared using `constant/axiom`. Now, suppose we define `bla` using `foo`. We can prove things about `bla` using the axioms above. Sure, if we make a mistake in the axioms, we can prove false. My point is, I don't want `bla` to be tagged as `meta/untrusted`. I want it to be a regular definition, and when I type `print axioms bla`, I get `foo` and all its properties.

I think scenarios 1&2 are very different from 3. The scenarios 1&2 already exist in the stdlib. Scenario 3 is very natural, and @jroesch is planning to pursue this direction.

gebner:
> By the same token, if we make a tiny mistake at type_checker.cpp, we can derive false.

Yes, for Lean itself this is a good argument.  Although for reference type-checkers there is slight difference as `type_checker.cpp` is part of the trusted code base, and `quot.lift` is part of the untrusted user-provided environment export.

> Another option is to declare quot, quot.mk, quot.ind and quot.lift at quot.cpp. We defined them in quot.lean just because it was more convenient. We can also remove the dependency from setoid.

This option feels the most correct to me, at least in the context of reference type-checkers.  It is very similar to how we define inductive types.  But changing the quotient implementation is low on my priority list.

To be honest, I was just completely surprised to see that there are axioms in the standard library that we 1) hide by default, and 2) where changes can cause inconsistency even though the axioms are not wrong by themselves.

leodemoura:
> To be honest, I was just completely surprised to see that there are axioms in the standard library that we 1) hide by default, and 2) where changes can cause inconsistency even though the axioms are not wrong by themselves.

Well, we are also hiding all inductive datatype constructors and recursors `*.rec`, because they are not considered to be axioms in the literature.

> This option feels the most correct to me, at least in the context of reference type-checkers. It is very similar to how we define inductive types. But changing the quotient implementation is low on my priority list.

I will change it. I feel strongly about it. It is too asymmetric to not list the inductive datatype constructors and recursors,  but list the constructors and recursors for `quot`.

library/system/io.lean
gebner: I see.  Then we should probably only print the axioms (transitively) referenced by the specified theorems.

leodemoura:
> Right now it prints by default all axioms and the theorem statements that are specified as command-line arguments:

This is great.

> Any suggestions for further info?

Should we also export infix/prefix/postfix notation declarations?
It would make the output much more readable.
I think it would be too much work to export all notation declarations. 
What do you think?

gebner:
> Should we also export infix/prefix/postfix notation declarations?

I'll take a look at it.  Do you think reconstructing infix/prefix/postfix commands is easier than just dumping the parse table?  We should probably ignore foldr and binder notation for simplicity though.

leodemoura:
> I'll take a look at it. Do you think reconstructing infix/prefix/postfix commands is easier than just dumping the parse table? 

I was thinking about the export format. It seems easy to just add very simple commands for infix/prefix/postfix. 

> We should probably ignore foldr and binder notation for simplicity though.

I agree. After we add parser extensions, I'm not even sure we will keep it.

gebner:
I have now added support for infix/prefix/postfix notation using the simplest possible commands.  We can declare the notation only for patterns of the form `fn #1 #0`, where fn is a constant.  Also left/right associativity is unsupported.

The export looks like this:
```
#PREFIX 3977 100 𝒫
#POSTFIX 303 1034 ⁻¹
#INFIX 791 50  ∈ 
```

And the output now looks like this:
```
...
axiom propext : Π {a b : Prop}, (a <-> b) -> a = b
theorem nat.add_assoc : Π n m k : nat, (n + m) + k = n + (m + k)
...
```

PR #1333, opened 2017-01-24
Do not update info windows during text selection

### LINKED ISSUE #1332: (Emacs mode) Next Error mode interferes with editor features
In the buffer

```
check 42[C]
```

where `[C]` is a place of the caret, when switched to the Next Error mode, it's not possible to select text. 

Editor: Aquamacs 3.2 GNU Emacs 24.4.51.2 (OS X)

Kha:
Fixes #1332


Kha:
I have _no_ idea about the actual cause, but this seems like an acceptable workaround.

PR #1337, opened 2017-01-26
Allow VM code to create tasks

gebner:
In the category of "cool things we can do with the tactic monad", we can now define a tactic combinator in Lean that replaces the current goal with a lemma and proves it asynchronously in a different thread:
```lean
  def fill_shr (x : bitvec n) (i : ℕ) (fill : bool) : bitvec n :=
  bitvec.cong (by async { begin [smt] by_cases (i ≤ n), eblast end }) $
    repeat fill (min n i) ++ₜ taken (n-i) x
```

However I stumbled into a major issue: `rb_map` is not thread-safe, even with `ts_vm_obj`.  I think the reason is that we do not clone the comparison function, any ideas on how to solve that @leodemoura?

### DIFF
--- a/library/init/meta/async_tactic.lean
+++ b/library/init/meta/async_tactic.lean
@@ -1,0 +1,58 @@
+prelude
+import init.meta.tactic
+import init.meta.interactive
+
+namespace tactic
+
+private meta def run_or_fail {α} (s : tactic_state) (tac : tactic α) : α :=
+match tac s with
+| (tactic_result.success a s) := a
+| (tactic_result.exception .α fmt _ s') :=
+  undefined_core $ to_string $ fmt () ++ format.line ++ to_fmt s'
+end
+
+meta def run_async {α : Type} (tac : tactic α) : tactic (task α) := do
+s ← read, return $ task.delay $ λ _,
+  match tac s with
+  | (tactic_result.success a s) := a
+  | (tactic_result.exception .α fmt _ s') :=
+    undefined_core $ to_string $ fmt () ++ format.line ++ to_fmt s'
+  end
+
+private meta def get_undeclared_const (env : environment) (base : name) : ℕ → name | i :=
+let n := base <.> ("_aux_" ++ to_string i) in
+if ¬env^.contains n then n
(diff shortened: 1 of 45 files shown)


leodemoura:
> However I stumbled into a major issue: rb_map is not thread-safe, even with ts_vm_obj. I think the reason is that we do not clone the comparison function, any ideas on how to solve that @leodemoura?

I will try to fix this. I incorrectly assumed all `vm_external` objects did not contain nested `vm_obj`s. I forgot `rb_map` :(

library/init/meta/async_tactic.lean
```diff
@@ -0,0 +1,49 @@
...
+    undefined_core $ to_string $ fmt () ++ format.line ++ to_fmt s'
+  end
+
+meta def prove_goal_async (tac : tactic unit) : tactic unit := do
+ctx ← local_context, revert_lst ctx,
+tgt ← target, tgt ← instantiate_mvars tgt,
+env ← get_env, tgt ← return $ env^.unfold_untrusted_macros tgt,
+when tgt^.has_meta_var (fail $ "goal contains metavariables"),
+params ← return tgt^.collect_univ_params,
+lemma_name ← mk_fresh_name,
```
leodemoura: `mk_fresh_name` is not good for creating lemma names. It only guarantees freshness in the current session. For example, suppose we use `async` in files `A.lean` and `B.lean`. Then, when we import `A` and `B`, there is a risk of name collision.

One possible ugly solution is to pass the name `C` of the surrounding declaration, and create a name such as `C._aux_<idx>`. We can make it less ugly by storing the name of the surrounding declaration in the `tactic_state`.

gebner: Ok, I did the less ugly version.  The tactic now creates names such as `C._aux_1` etc., see the test for an example.

gebner:
It seems to work pretty well.  Even the info manager and show-goal work transparently!

library/init/meta/async_tactic.lean
leodemoura: Nice.

PR #1338, opened 2017-01-26
Default field value fixes

Kha:
I wanted to explore how we could make use of default field values in the monadic hierarchy, but quickly got sidetracked :) .


src/frontends/lean/structure_cmd.cpp
```diff
@@ -888,10 +890,10 @@ struct structure_cmd_fn {
                 }
                 name decl_name  = name(m_name + local_pp_name(field.first), "_default");
                 expr decl_type  = Pi(args, mlocal_type(field.first));
-                expr decl_value = Fun(args, *field.second);
+                expr decl_value = Fun(args, mk_explicit(*field.second));
```
Kha, line 893: Kind of a hack to make the loop in `mk_field_default_value` stop. I can document it, but I'm also open to better ideas.

Issue #1340, opened 2017-01-27
RFC: lightweight default parameters


leodemoura:
We currently do not have support for default parameter values in Lean. So, we use the following idiom.
1- Define a "core" function that takes "configuration parameters"
```lean
/- (apply_core t approx all insts e), apply the expression e to the main goal,
   the unification is performed using the given transparency mode.
   If approx is tt, then fallback to first-order unification, and approximate context during unification.
   If all is tt, then all unassigned meta-variables are added as new goals.
   If insts is tt, then use type class resolution to instantiate unassigned meta-variables. -/
meta constant apply_core    : transparency → bool → bool → bool → expr → tactic unit
```

2- Define the common case that fixes these parameters 
```lean
meta def apply : expr → tactic unit :=
apply_core semireducible tt ff tt
```

This style is quite annoying to use when the default is not what we want. 
We now have default structure values, then we can make the approach above less ugly.

```lean
structure delta_config :=
(max_steps       := default_max_steps)
(visit_instances := tt)

meta def delta (cs : list name) (e : expr) (cfg : delta_config) : tactic expr := ...
```

Then the user can write
```lean
delta cs e {}  -- default configuration
delta cs e {max_steps := 1000}   
delta cs e {visit_instances := ff}
...
```

However, it is still a little bit annoying to keep writing `{}` when we want the default configuration.

Here are some possibilities:

1- Add a new attribute `@[default_params]`. It would instruct the elaborator to replace `delta cs e` with `delta cs e {}`. To be precise, if `f` has the attribute `@[default_params]`, then for any term `(f ...)` or `f`, if it has type `S -> A` where `S` is a structure, we automatically add the `{}`. 
This is a little bit hackish, but we don't have to change the parser nor add a new kind of binder.
It is also a small change in the elaborator.

2- Use type classes. It is described here: https://github.com/leanprover/lean/issues/1121#issuecomment-274671634

This one has two problems. It has an extra setup step: we have to define a class, and then an instance with the default values. Second, it is annoying to use when we want to change the default for a single application. We would have to use `@f ...`. This is particularly annoying if `f` has many implicit arguments, and we would also need to use many `_`s.

The main advantage of this approach is that we don't have to change anything in the elaborator and/or parser.

[...]

PR #1346, opened 2017-01-30
fix(frontends/lean/parser): save noncomputable theory flag in snapshots

gebner:


gebner:
This is now the second flag that has been missing from the snapshots, and I'm still not sure if we're missing anything else.  I guess I'll refactor the parser so that we just always use snapshots, i.e. parse one command, restore from snapshot, and repeat again.

leodemoura:
> I guess I'll refactor the parser so that we just always use snapshots, i.e. parse one command, restore from snapshot, and repeat again.

I think this change can wait.
We are planning to refactor the parser to support the parser extensions written in Lean.
We can combine both changes in the same refactoring.

gebner:
> We can combine both changes in the same refactoring.

I just saw the design of the parser extensions in your comment at #1340, and I agree, it makes sense to combine the two changes.  AFAICT the parser extension monad will be very similar to the tactic monad, except that we have `parser_state` instead of `tactic_state`.  Once we have proper first-class parser states, we can drop the old snapshot infrastructure.
We can even move the saving of the snapshots out of the parser.  This could be handled by the module_mgr.

But I think in the meantime, we should still merge this PR.  Otherwise files with the `noncomputable theory` flag break if you edit them in the middle.

leodemoura:
> But I think in the meantime, we should still merge this PR. Otherwise files with the noncomputable theory flag break if you edit them in the middle.

I'm planning to merge it. Yesterday was an "intense" day of hacking. I will do it today.

Issue #1355, opened 2017-02-04
RFC: saying sorry


gebner:
Once upon a time, `sorry` was only used if the user actually typed `sorry` in a Lean file.  However, more and more failure conditions are now being excused by `sorry` as well--usually in a laudable and worthwhile effort to recover from parsing and elaboration errors.
 1. As intended, we use `sorry` when the user types `sorry`.
 2. We also use `sorry` as a placeholder for the value of a definition when elaboration fails.
 3. We might use `sorry` in the future as a placeholder in expressions that have parse errors (e.g. `(if x > 0 then tt)` would be parsed as `if x > 0 then tt else sorry`)
 4. While developing proof automation, we might want to insert `sorry` as proof terms.

Aside from 1., all these applications have one issue in common: we cannot syntactically determine from the pre-expression whether we will use sorry or not.  Up until now, we added a `sorry` axiom to the environment the first time it was used.  This worked well when Lean was single-threaded and the main source for `sorry`s was explicit use by the user.  However now `sorry` may first appear out of nowhere in a lemma that is elaborated asynchronously, and at that point we can no longer add it to the environment.
Hence I believe that we should just proactively add the sorry axiom as the first declaration to the environment.  I propose that we don't add the sorry axiom in the standard library but declare it in the C++ code instead, then it doesn't get exported and we can easily disable it using an option.

Another issue is that how we handle the computability lint for sorry: because sorry is noncomputable, we would normally get errors in all dependent definitions (because we would usually have to mark them noncomputable as well).  As a workaround, if the parser sees sorry (if the user enters it) we set the "noncomputable theory" flag for the current file.  Other files still show the noncomputability errors.  Actual noncomputability errors are hidden.
I think it would be much simpler if we just had a special case for sorry, and made it computable by fiat.  (We already have VM code for it!)

Concretely, I propose the following changes:
1. Add a global option `allow_sorry`.  This option is set to true by default, at least interactively.
2. If `allow_sorry` is set to to true, we add a `sorry` axiom to the initial environment.
3. The `noncomputable` extension treats `sorry` as computable.
4. We report a warning when we add a declaration to the environment that uses sorry.
5. The type checkers do not add the sorry axiom.

The main objection I can see is that `print axioms` would now always print one more axiom (by default), namely sorry.  But if I understand the plan for the native code interface correctly, we will get new constants and axioms anyhow--for example the constant `io` and axioms describing the properties of the IO monad.

leodemoura:
[...]

PR #1358, opened 2017-02-05
Implement sorry as a macro

gebner:
See #1355.  This PR implements sorry using a macro instead of using an axiom.

The macro takes a single argument, its type.  (We need some kind of polymorphism here, either universe or type.)  The pre-expression is `[sorry Prop]`.

1. `print axioms foo` shows `[sorry]` if foo uses the sorry macro.
2. We check imports for sorrys.  (~~~By searching all expressions, and this is a bit slow.  Caching this is not completely trivial since we might only find out that an import uses sorry long after having imported it, if sorry occurs in a proof.~~~ We now save that information as a flag in the olean files.)
3. We check each added declaration for sorry in `module::add`.  On the one hand this is positive since we now get a warning if sorry is used.  On the other hand we get a lot of warnings:
```
vm_sorry.lean:1:4: warning: declaration 'half_baked._main' uses sorry
vm_sorry.lean:1:4: warning: declaration 'half_baked._main.equations._eqn_1' uses sorry
vm_sorry.lean:1:4: warning: declaration 'half_baked.equations._eqn_1' uses sorry
```
4. The `sorry` macro is computable, and we generate VM code for it.

### DIFF
--- a/library/init/meta/expr.lean
+++ b/library/init/meta/expr.lean
@@ -104,6 +104,11 @@ const `true []
 meta def mk_false : expr :=
 const `false []
 
+/-- Returns the sorry macro with the given type. -/
+meta constant mk_sorry (type : expr) : expr
+/-- Checks whether e is sorry, and returns its type. -/
+meta constant is_sorry (e : expr) : option expr
+
 meta def app_of_list : expr → list expr → expr
 | f []      := f
 | f (p::ps) := app_of_list (f p) ps
--- a/library/init/meta/interactive.lean
+++ b/library/init/meta/interactive.lean
@@ -321,6 +321,9 @@ do e ← to_expr p,
 meta def trivial : tactic unit :=
 tactic.triv <|> tactic.reflexivity <|> tactic.contradiction <|> fail "trivial tactic failed"
 
+/-- Closes the main goal using sorry. -/
+meta def admit : tactic unit := tactic.admit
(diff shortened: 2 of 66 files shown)


gebner:
On a related note: now that we can easily use sorry, I wanted to make the elaborator a bit more resilient against errors.  That is, insert sorry instead of throwing exceptions.  This is useful in some cases (top-level definitions), and probably unwanted in others (pre-expression arguments to tactics).

The plan would be the following: unless we're inside a possible backtracking point, we return sorry (and report an error) instead of throwing exceptions.  Inside a backtracking point, we do not recover using sorry but keep throwing exceptions.  This should preserve the current backtracking behavior of the elaborator.  At the end, we solve all open meta-variables using sorry.  This mode would replace the double elaboration (second time with sorry) in `single_definition_cmd_core`.  Does this sound reasonable?

leodemoura:
> Does this sound reasonable?

Yes, this is reasonable, and useful.

We talked about adding error recovering when we were discussing type based auto completion in Paris.
After we have error recovering in the elaborator, we will only need to do it for the parser :)

gebner:
> After we have error recovering in the elaborator, we will only need to do it for the parser :)

I wanted to wait with the parser until the refactoring is over.  Then we can do error recovery and error messages with ranges.

PR #1359, opened 2017-02-05
chore(*): more warnings from clang

diakopter:


diakopter:
deleting; will resubmit

PR #1361, opened 2017-02-06
Recover from almost all elaboration errors using sorry

gebner:
This is a (non-exhaustive) list of errors that we can now recover from:
```lean
def half_baked : ℕ → ℕ
| 3  := 2
-- type mismatches
| 0  := 1 + ""
-- placeholders
| 5  := _ + 4
-- missing type-class instances
| 42 := if 2 ∈ 3 then 3 else _
-- exceptions during tactic evaluation
| 7  := by do undefined
-- nested elaboration errors
| _  := begin exact [] end
```
This is the elaborated term:
```lean
λ (a : ℕ),
  ite (a = 3) 2
    (ite (a = 0) (1 + sorry)
      (ite (a = 5) (sorry + 4)
        (ite (a = 42) (ite (2 ∈ 3) 3 sorry)
          (ite (a = 7) sorry sorry))))
```
The resulting definition gets added to the environment, and `vm_eval half_baked 3` returns 2, as expected.

The approach is pretty much as in described in #1358.  We only try do error recovery outside backtracking points, and set a boolean flag when backtracking is possible.  If error recovery is enabled, we try to return sorry instead of throwing an exception whenever possible.  In addition, the `visit` method catches exceptions and turns them into sorry as well.

There are still a few rough edges:
 * Missing type class instances are reported more than once.  That is, `x + y + z + w` generates 3 error messages about the missing has_add instance.
 * Partial applications are not sorry-padded, but completely replaced by sorry.  For example, `3 + (nat.add 4)` is elaborated to `3 + sorry` instead of `3 + (4 + sorry)`.  This situation is probably common in partial inputs, so we should keep it in mind for autocompletion.
 * Metavariables are solved using sorry in `ensure_no_unassigned_metavars`.  This is probably not the right place to do it, but I think it is the right time during elaboration.
 * The additional information in `visit_elim_app` is not added to the error in error-recovery mode.

### DIFF
--- a/src/frontends/lean/definition_cmds.cpp
+++ b/src/frontends/lean/definition_cmds.cpp
@@ -157,7 +157,8 @@ environment mutual_definition_cmd_core(parser & p, def_cmd_kind kind, decl_modif
     buffer<expr> fns, params;
     declaration_info_scope scope(p, kind, modifiers);
     expr val = parse_mutual_definition(p, lp_names, fns, params);
-    elaborator elab(p.env(), p.get_options(), local_pp_name(fns[0]), metavar_context(), local_context());
+    bool recover_from_errors = true;
+    elaborator elab(p.env(), p.get_options(), local_pp_name(fns[0]), metavar_context(), local_context(), recover_from_errors);
     buffer<expr> new_params;
     elaborate_params(elab, params, new_params);
     val = replace_locals_preserving_pos_info(val, params, new_params);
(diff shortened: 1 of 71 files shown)


src/frontends/lean/definition_cmds.cpp
```diff
@@ -678,6 +680,7 @@ class proof_elaboration_task : public task<expr> {
             finalize_theorem_proof(elab, params, val, m_finfo);
             if (m_is_rfl_lemma && !is_rfl_lemma(m_final_type, val))
                 throw exception("not a rfl-lemma, even though marked as rfl");
+            if (elab.has_errors()) throw std::exception();
```
leodemoura, line 683: I'm assuming that from now on, we use `std::exception()` when we want to interrupt the elaboration without reporting yet another error. Is that right?

gebner: The reason is the following workaround: in order not to forget any exceptions that occur during tasks, we catch exceptions that are subclasses of lean::throwable and report them automatically.  Throwing std::exception fails the task without producing an extra message.

So this is not related to the elaborator, but just to the tasks.

PR #1362, opened 2017-02-07
feat(shell/lean): add default memory limit in server mode

gebner:


gebner:
I'm not yet sure how to implement timeouts, but the memory limit is straightforward.  And it prevents the hang in Scott's example.

leodemoura:
I will try to add the deterministic timeout based on number of memory allocations. 

Today was not a productive day for me, I wasted a lot of time on lean-user issues.

PR #1368, opened 2017-02-10
Change induction, cases, and apply to return information how they changed the goal

johoelzl:
`induction` and `cases` return now information about added hypotheses and how they modified existing hypotheses. `apply` returns *all* generated mvars.

With this we do not need to analyze the applied rules beforehand, also we do not need to generate names used by `induction` and `cases`. It is also easy to check that the result is as expected (i.e. `do [[a, b]] <- cases r`).

For `apply` it may be important to know which mvars where generated and which ones are available in the generated subgoals. For some cases it can be hard to predict which mvars are assigned. This change should allow an easier map of rule parameters to subgoals.

### DIFF
--- a/library/init/meta/interactive.lean
+++ b/library/init/meta/interactive.lean
@@ -312,7 +312,8 @@ do e ← i_to_expr p,
    match rec_name with
    | some n := induction_core semireducible e n ids
    | none   := do I ← get_type_name e, induction_core semireducible e (I <.> "rec") ids
-   end
+   end,
+   return ()
 
 meta def cases (p : qexpr0) (ids : with_ident_list) : tactic unit :=
 do e ← i_to_expr p,
--- a/library/init/meta/interactive.lean
+++ b/library/init/meta/interactive.lean
@@ -317,7 +317,7 @@ do e ← i_to_expr p,
 
 meta def cases (p : qexpr0) (ids : with_ident_list) : tactic unit :=
 do e ← i_to_expr p,
-   if e^.is_local_constant then
+   (if e^.is_local_constant then
      cases_core semireducible e ids
    else do
      x ← mk_fresh_name,
@@ -327,7 +327,8 @@ do e ← i_to_expr p,
            get_local x >>= tactic.revert,
            return ()),
      h ← tactic.intro1,
-     cases_core semireducible h ids
+     cases_core semireducible h ids),
+   return ()
 
 meta def destruct (p : qexpr0) : tactic unit :=
 i_to_expr p >>= tactic.destruct
--- a/library/init/meta/relation_tactics.lean
+++ b/library/init/meta/relation_tactics.lean
@@ -14,7 +14,7 @@ do tgt ← target,
    env ← get_env,
    r   ← return $ get_app_fn tgt,
    match (op_for env (const_name r)) with
-   | (some refl) := mk_const refl >>= apply_core t tt ff tt
+   | (some refl) := mk_const refl >>= apply_core t tt ff tt >> return ()
    | none        := fail $ tac_name ++ " tactic failed, target is not a relation application with the expected property."
    end
 
--- a/library/init/meta/tactic.lean
+++ b/library/init/meta/tactic.lean
@@ -413,8 +413,13 @@ meta constant abstract_hash : expr → tactic nat
 meta constant abstract_weight : expr → tactic nat
 meta constant abstract_eq     : expr → expr → tactic bool
 /- (induction_core m H rec ns) induction on H using recursor rec, names for the new hypotheses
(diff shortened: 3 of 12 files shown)


library/init/meta/interactive.lean
```diff
@@ -310,13 +310,14 @@ do e_type ← infer_type e >>= whnf,
 meta def induction (p : qexpr0) (rec_name : using_ident) (ids : with_ident_list) : tactic unit :=
 do e ← i_to_expr p,
    match rec_name with
-   | some n := induction_core semireducible e n ids
+   | some n := do induction_core semireducible e n ids
    | none   := do I ← get_type_name e, induction_core semireducible e (I <.> "rec") ids
```
leodemoura: Is this `do` necessary?

library/init/meta/relation_tactics.lean
```diff
@@ -14,7 +14,7 @@ do tgt ← target,
    env ← get_env,
    r   ← return $ get_app_fn tgt,
    match (op_for env (const_name r)) with
-   | (some refl) := mk_const refl >>= apply_core t tt ff tt
+   | (some refl) := do r ← mk_const refl, apply_core t tt ff tt r, return ()
    | none        := fail $ tac_name ++ " tactic failed, target is not a relation application with the expected property."
```
leodemoura: It seems you do not like the `>>=` and `>>` notation :)
We can also write
``` 
mk_const refl >>= apply_core t tt ff tt >> return ()
```

library/init/meta/interactive.lean
johoelzl: Ups, no. If first added the `return ()` in each branch. I remove it.

library/init/meta/relation_tactics.lean
johoelzl: Ah, with `>>` it looks okay. I change it.

leodemoura: Thanks. BTW, there are other instances that can be written using `>>`

library/init/meta/tactic.lean
```diff
@@ -413,11 +414,20 @@ meta constant abstract_hash : expr → tactic nat
...
 meta constant abstract_eq     : expr → expr → tactic bool
 /- (induction_core m H rec ns) induction on H using recursor rec, names for the new hypotheses
-   are retrieved from ns. If ns does not have sufficient names, then use the internal binder names in the recursor. -/
-meta constant induction_core : transparency → expr → name → list name → tactic unit
+   are retrieved from ns. If ns does not have sufficient names, then use the internal binder names
+   in the recursor.
+   It returns for each new goal a list of new hypotheses and a substitution mapping for hypothesis
+   depending on H. The names in the substitution mapping are only valid in the original goal, not in
+   the type contexts of the new goals. -/
+meta constant induction_core : transparency → expr → name → list name → tactic (list (list expr × list (name × expr)))
```
leodemoura: This is cool, you are also returning the hypotheses that have been renamed.
BTW, we are not very precise what you mean by "renaming" here and in the rest of the API.
Each hypothesis has two names: internal and user. The user names are for user interaction (e.g., pretty printing). The internal names are unique. 
This tactic does not rename the "user" names, only the internal ones.
Any ideas on how to explain this stuff in the API?

library/init/meta/tactic.lean
```diff
@@ -391,8 +391,9 @@ meta constant set_goals     : list expr → tactic unit
    the unification is performed using the given transparency mode.
    If approx is tt, then fallback to first-order unification, and approximate context during unification.
    If all is tt, then all unassigned meta-variables are added as new goals.
-   If insts is tt, then use type class resolution to instantiate unassigned meta-variables. -/
-meta constant apply_core    : transparency → bool → bool → bool → expr → tactic unit
+   If insts is tt, then use type class resolution to instantiate unassigned meta-variables.
+   It returns a list of all introduced meta variables, even the assigned ones. -/
+meta constant apply_core    : transparency → bool → bool → bool → expr → tactic (list expr)
 /- Create a fresh meta universe variable. -/
```
leodemoura, line 397: This is also very useful.

library/init/meta/tactic.lean
johoelzl: I don't have any good suggestions. But I think *internal* name and *user* name are understandable words. I can change it to:

> The substitution maps internal names to their replacement terms. If the replacement is again a hypothesis the user name stays the same. The internal names are only valid in the original goal, not in the type context of the new goal.

leodemoura: Great!

johoelzl:
I repushed. I undid all `>>=`, `>>` to `do` changes. I updated to documentation. Also, please check the C++ code: I did run `valgrind` on simple tests but this is my first time to write C++ in a long time.

PR #1375, opened 2017-02-16
fix(kernel/type_checker): disable expression caching in asynchronous proof checking

gebner:
This fixes the performance problem I had with the example I posted on slack.

I noticed that we actually never use expression caching, at least with `-j0`: we disable caching in the parser and everything inherits that flag.  (There are tons of other places where we disable it, but none where we enable it.)  @leodemoura Should we just disable expression caching by default?

### DIFF
--- a/src/kernel/type_checker.cpp
+++ b/src/kernel/type_checker.cpp
@@ -760,6 +760,7 @@ public:
     }
 
     expr execute() override {
+        scoped_expr_caching disable(false);
         bool memoize = true;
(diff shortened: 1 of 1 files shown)


leodemoura:
> @leodemoura Should we just disable expression caching by default?

Yes, we should. I will also investigate how to make this cache better.

BTW, in server mode, this cache may be consuming a lot of memory since we don't remove anything from it. This may explain some memory consumption problems I had this week.

PR #1390, opened 2017-02-20
misc

Kha:


leodemoura:
@kha the files `vm_pos_info.*` are missing

Kha:
Fixed, thanks. Plus a new `expr` gdb pretty printer as bonus.

PR #1394, opened 2017-02-22
expr patterns

Kha:
This turned out to be much more intricate to implement than I expected, as always.

still missing:
* ignore implicit parameters
* more robust error handling

### DIFF
--- a/library/init/meta/expr.lean
+++ b/library/init/meta/expr.lean
@@ -212,23 +212,23 @@ is_constant_of (get_app_fn e) n
 meta def is_napp_of (e : expr) (c : name) (n : nat) : bool :=
 is_app_of e c ∧ get_app_num_args e = n
 
-meta def is_false (e : expr) : bool :=
-is_constant_of e `false
+meta def is_false : expr → bool
+| ``(false) := tt
+| _         := ff
 
+set_option eqn_compiler.max_steps 256
 meta def is_not : expr → option expr
-| (app f a)     := if is_constant_of f `not then some a else none
-| (pi n bi a b) := if is_false b then some a else none
-| e             := none
+| ``(not %%a)     := some a
+| ``(%%a → false) := some a
+| e               := none
 
-meta def is_eq (e : expr) : option (expr × expr) :=
-if is_napp_of e `eq 3
-then some (app_arg (app_fn e), app_arg e)
-else none
+meta def is_eq : expr → option (expr × expr)
+| ``((%%a: %%α) = %%b) := some (a, b)
+| _                    := none
 
-meta def is_ne (e : expr) : option (expr × expr) :=
-if is_napp_of e `ne 3
-then some (app_arg (app_fn e), app_arg e)
-else none
+meta def is_ne : expr → option (expr × expr)
+| ``((%%a: %%α) ≠ %%b) := some (a, b)
+| _                    := none
 
 meta def is_bin_arith_app (e : expr) (op : name) : option (expr × expr) :=
 if is_napp_of e op 4
@@ -247,13 +247,10 @@ is_bin_arith_app e `le
 meta def is_ge (e : expr) : option (expr × expr) :=
 is_bin_arith_app e `ge
 
-meta def is_heq (e : expr) : option (expr × expr × expr × expr) :=
-if is_napp_of e `heq 4
-then some (app_arg (app_fn (app_fn (app_fn e))),
-           app_arg (app_fn (app_fn e)),
-           app_arg (app_fn e),
-           app_arg e)
-else none
+set_option eqn_compiler.max_steps 512
+meta def is_heq : expr → option (expr × expr × expr × expr)
+| ``(@heq %%α %%a %%β %%b) := some (α, a, β, b)
+| _                        := none
 
 meta def is_pi : expr → bool
 | (pi _ _ _ _) := tt
--- a/library/init/meta/expr.lean
+++ b/library/init/meta/expr.lean
@@ -216,7 +216,6 @@ meta def is_false : expr → bool
 | ``(false) := tt
 | _         := ff
 
-set_option eqn_compiler.max_steps 256
 meta def is_not : expr → option expr
 | ``(not %%a)     := some a
 | ``(%%a → false) := some a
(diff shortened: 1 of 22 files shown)


library/init/meta/expr.lean
```diff
@@ -212,23 +212,23 @@ is_constant_of (get_app_fn e) n
 meta def is_napp_of (e : expr) (c : name) (n : nat) : bool :=
 is_app_of e c ∧ get_app_num_args e = n
 
-meta def is_false (e : expr) : bool :=
-is_constant_of e `false
+meta def is_false : expr → bool
+| ``(false) := tt
+| _         := ff
 
+set_option eqn_compiler.max_steps 256
```
Kha: The eqn compiler definitely needs a few more steps for these, but the output size looks still reasonable.

src/frontends/lean/elaborator.cpp
```diff
@@ -2305,7 +2306,7 @@ class validate_equation_lhs_fn : public replace_visitor {
     }
 
     virtual expr visit_macro(expr const & e) override {
-        if (is_inaccessible(e)) {
+        if (is_inaccessible(e) || is_name_macro(e)) {
```
Kha: hack

library/init/meta/expr.lean
```diff
@@ -247,13 +247,10 @@ is_bin_arith_app e `le
...
-if is_napp_of e `heq 4
-then some (app_arg (app_fn (app_fn (app_fn e))),
-           app_arg (app_fn (app_fn e)),
-           app_arg (app_fn e),
-           app_arg e)
-else none
+set_option eqn_compiler.max_steps 512
+meta def is_heq : expr → option (expr × expr × expr × expr)
+| ``((%%a: %%α) == (%%b : %%β)) := some (α, a, β, b)
+| _                             := none
```
leodemoura: The new definitions look much nicer, but what is the performance impact?
If it is too high, and we can try to optimize the pattern matching compilation code.

src/library/type_context.cpp
```diff
@@ -658,7 +659,7 @@ bool type_context::should_unfold_macro(expr const & e) {
 
 optional<expr> type_context::expand_macro(expr const & e) {
     lean_assert(is_macro(e));
-    if (should_unfold_macro(e)) {
+    if (should_unfold_macro(e) && !is_name_macro(e)) {
```
Kha: hack. @leodemoura The problem is that in https://github.com/leanprover/lean/blob/master/src/library/equations_compiler/elim_match.cpp#L330, values will still be unfolded by `whnf_core`.

leodemoura: I will try to improve `whnf_head_pred`. In the new code, it will also invoke the predicate for macros.
Thus, the caller can decide whether the macro should be unfolded or not.
What do you think?

leodemoura: BTW, I'm wondering if we need the `name_macro`. We can recognize a `name` value like we do for `nat`, `char`, etc.

Kha: I believe it should be fixed by `whnf_core` also respecting `m_unfold_pred`, if that makes sense.

library/init/meta/expr.lean
Kha: Using some very ad-hoc benchmarking:
```
def repeat {α : Type} (t : thunk α) : ℕ → unit
| 0     := ()
| (n+1) := match (t (), repeat n) with (_, _) := () end

run_command do e ← to_expr `(1 == 2), return $ timeit "old" (repeat (expr.is_heq_old e) 10000)
run_command do e ← to_expr `(1 == 2), return $ timeit "new" (repeat (expr.is_heq e) 10000)

    41   1 info            old 3.58899 secs
    42   1 info            new 0.87240 secs
```
:)

leodemoura: Nice!

src/library/equations_compiler/elim_match.cpp
```diff
@@ -502,17 +503,14 @@ struct elim_match_fn {
...
-                        if (value_is_contructor) has_constructor = true;
-                        return true;
-                    } else {
-                        return false;
-                    }
                 }
+                return false;
             });
         return r && has_variable && has_constructor;
     }
```
Kha, line 513: @leodemoura I _think_ we can just always try value transitions before complete transitions and remove the `value_is_constructor` flag...?

src/library/type_context.cpp
Kha: Okay, `name_macro` is gone now.

Kha: (no changes in the benchmark)

src/library/equations_compiler/elim_match.cpp
Kha: There is some fallout from this change in form of 'enum' inductives being compiled with `ite` and then failing in the app builder because they don't have a `decidable_eq` instance. What should we do about that?

leodemoura: Do you need this change for finishing this PR?
If not, let's discuss it after the paper deadline. 
When we use "value" transition, the equation lemmas do not hold definitionally.
One option is to *never* use value transition for 'enum' inductive datatypes.

Here are other pending issues regarding the equation compiler:
- We need to support nested inductive datatypes in the complete transition.
- Equation validation will miss invalid equations that require complete transition, and use inaccessible annotations.

leodemoura: > When we use "value" transition, the equation lemmas do not hold definitionally.

BTW, this is why the test `eq11.lean` failed here:
https://ci.appveyor.com/project/leodemoura/lean/build/1.0.634

Kha: > Do you need this change for finishing this PR?

Without the change, `name` values will always be eliminated as constructors, meaning we can only use term patterns after defining `decidable_eq string` (which uses `mk_dec_eq_instance` currently).

IIUIC, enum patterns would _always_ be eliminated as constructors before this fix, so I think we can just go back to that behavior for now.

leodemoura: @kha Ok. So, I think we should remove `enum` inductive from `is_value`. That is, we remove the test `if (optional<name> I_name = is_constructor(e)) return is_nontrivial_enum(*I_name);` from `is_value`.  As you pointed out, the value transition was never being used for them anyway. 
After, we make this change, we can always try the value_transition before complete_transition, as you wanted.

Kha: Done and done.

PR #1396, opened 2017-02-23
feat(library/init/meta/vm): add decidable_eq for vm_obj_kind

jroesch:
This was nice to have when writing a simple vm_monitor.


jroesch:
Any ideas why that build failed? the log seems to just say the build might of stalled?

PR #1398, opened 2017-02-24
Miscellaneous

gebner:
@leodemoura Re: the example sorry change: I know you changed this to reduce the number of sorry warnings.  But if we disable it for examples, then `example : false := sorry` is accepted without any warning at all.


robertylewis:
On the topic of build instructions: I built Lean on Windows for the first time today, and it went very smoothly, except I had to install Python too. Was that a quirk/me doing something wrong? Or should I add it to the Windows build instructions?

gebner:
@rlewis1988 The python dependency should be completely optional.  We only use it for the C++ style checker.  At least on Linux you can build Lean without Python (you might get a warning when you run cmake, though).

robertylewis:
I just looked again, and it definitely appears to fail without Python:
```
$ cmake -G Ninja ../../src
...

CMake Error at C:/msys64/mingw64/share/cmake-3.7/Modules/FindPackageHandleStandardArgs.cmake:138 (message):
  Could NOT find PythonInterp (missing: PYTHON_EXECUTABLE)
Call Stack (most recent call first):
  C:/msys64/mingw64/share/cmake-3.7/Modules/FindPackageHandleStandardArgs.cmake:378 (_FPHSA_FAILURE_MESSAGE)
  C:/msys64/mingw64/share/cmake-3.7/Modules/FindPythonInterp.cmake:151 (FIND_PACKAGE_HANDLE_STANDARD_ARGS)
  CMakeLists.txt:273 (find_package)


-- Configuring incomplete, errors occurred!

...
$ pacman -S python2
...
$ cmake -G Ninja ../../src
...
(succeeds)
```

Issue #1364, opened 2017-02-08
Lean mode freezing Emacs and sluggish VSCode 


leodemoura:
As @rlewis1988 also reported, Emacs often freezes when we are editing the standard library.
Yesterday, I had to disable lean-mode to be able to edit `tactic.lean`, otherwise I would have to wait 5 secs after each keystroke.

I built a repro the exposes the problem. To execute the repro, we need to use the latest version which adds the command `sleep` and `long_sleep` to the Lean server.

As far as I can tell, the problem happens when we open files `A.lean` and `B.lean`, where `B.lean` depends on `A.lean` (and many other files that also depend on `A.lean`). Then, we start editing `A.lean` (e.g., introducing an fixing errors).

In my first repro: https://gist.github.com/leodemoura/0e3be8b2f2f364e0677392db9e8f2dc7

I open `tactic.lean`,  open `vector.lean`, break the `tactic_result` declaration at `tactic.lean`, and then fix it.
For this sequence of operations, the lean server produces a 210Mb output trace. This trace corresponds to 2 keystrokes. I imagine that yesterday when I was editing the file, several terabytes of information were being sent between Emacs and Lean server.

By inspecting the trace, it seems the server is producing longer and longer messages `{"msg", ...}`, where message `i` is a prefix of message `i+1`.

In the second repro: https://gist.github.com/leodemoura/d9e78ba6f2fffdbe0400344c3d6f10ee

The sequence of events is similar, but I do not open `vector.lean`. In this case, the lean server produces a 5Mb trace

This problem is not specific to the standard library. It seems it will affect any project that has many files, and non trivial dependencies. 

I also tried the experiment with VS Code. It doesn't freeze, but everything is very slow, and the feedback produced by the editor is more or less useless since it takes a long time to stabilize. 

@rlewis1988: I'm using the following workaround to edit stdlib files, I open a fresh Emacs process, and open only the file I want to edit.

gebner:
Thanks for making an easily reproducible example.  Not that for reproduction you need to replace `\/home\/leo\/projects\/lean` with your local lean checkout.

In this case the server produces 3857904 messages.  Many of those are full message updates, that is, where we send all available messages again, because an old message got removed.  I guess we can just limit these full refreshes to once every 200ms or so.

gebner:
Just some more data.  Virtually all of the messages are full refreshes.  Limiting them to once every 100ms would cut down the responses by a factor of over 8000!
```
[...]

PR #1405, opened 2017-02-28
[RFC] New parallel compilation infrastructure

gebner:
TL;DR This is why you want it:

2017-02-28-152124_1920x1080_scrot

I have been using this branch ~almost~ exclusively for the last few ~weeks~ months, and it
should be pretty stable now.

~There is one nasty memory leak though (that is also in master): open
`init/core.lean` and `standard.lean`. Then change something in `init/core.lean`.
Every change uses an additional few hundred megabytes for me. I would really
appreciate any ideas on how to debug this. So far I've tried valgrind (too slow
on the whole library, small examples don't show any leaks) and heaptrack
(doesn't give any clues either), without success.~

## New parallel compilation infrastructure

While the current parallel compilation and server mode works pretty well, there
a few issues:

 * Defining tasks in C++ is cumbersome.
 * Message reporting, cancellation, task queues are terribly tightly coupled.
   The code for `mt_task_queue` is horribly complex at the moment because it has
   lots of responsibilities, and there are probably lots of bugs.
 * The approach requires a hand-rolled task queue, and does not easily transfer
   to other programming languages, or even just multiple Lean instances inside
   the same process (which might reasonably happen when Lean is used as a
   library).
 * We cannot work on a file without compiling its dependencies.  The
   dependencies (in particular proof elaborations) are always executed.
 * Similarly, we eagerly compile all reverse dependencies.  This is a huge
   problem if you edit `init/core.lean` and have `data/list/comb.lean` open at
   the same time.
 * When handling info command we indicate that the server is busy (with the
   "hourglass" icon in Emacs), just because the command spawns a task.

This PR solves all of these issues.  I think a good way to view the problem is
to consider 4 very similar, but subtly different graphs:

 1. Task dependency graph.  (Contains an edge if a task needs to be run after
    another one.  For example, type-checking must run after elaboration.)
 2. Module dependency graph.  (Reflects the import statements.)
 3. Message tree.  (How messages can be reused.  When reparsing a file, we
    keep the messages form the old proof elaborations around, and only replace
    them once they are re-elaborated as well.)
 4. Cancellation graph.  (Which tasks need to be cancelled when we reload a
    file?)

These are now separate and mostly explicitly represented.  Currently tasks have
metadata which we use for 3. and 4., this metadata is replaced by explicit data
structures.

Additionally, this PR adds the following new features:
 * By default, we only check the visible *lines*.  You can also choose to check
   nothing, or the complete visible files.
 * Greatly reduced number of sorry warnings.
 * The command-line exit status is 1 if and only if an error was reported.  This
   matches the error reporting in Emacs. Concretely, we missed two error in the
   run tests due to the current mismatch: `quote1.lean` and `exact_perf.lean`.
   @leodemoura `quote1` was an easy fix, but `exact_perf.lean` reports an error
   because `try { exact h2 }` fails, and `h2` has an elaboration error. You
   added this error message at
   https://github.com/leanprover/lean/commit/add5266df75bf04c721670ea5ac0e3c59209c15b
   Any ideas?

I have removed all incremental message/tasks updates from the server.  The
server resends all running tasks / all messages every time (about every 100ms).
(See #1364 for how we arrived at this approach.)
However these are only sent for the region that is checked: for example if
you choose "check visible lines", then you only see errors for the visible lines.

There are two backwards-compatible changes in the server protocol:
 * The `additional_message` command got removed.
 * There is now a `roi` (region of interest) command:
```json
{
  "seq_num": 45,
  "command": "roi",
  "files": [
    {
      "file_name": "/home/gebner/lean/library/init/core.lean",
      "begin_line": 1,
      "end_line": 69
    }
  ]
}
 ```

## Task dependency graph (and tasks in general)

As before, the task dependency graph is specified by the dependencies of each
task.  Every task has a method that computes a list of its dependencies:

```c++
virtual void get_dependencies(buffer<gtask> &) {}
```

The reason this is a method is because the dependencies can change dynamically.
Consider for example a dependency on `task<task<expr>>`, here the
`get_dependencies` method would first return the outer task, and then both the
outer and inner task once the outer one is finished.

Since tasks no longer require so much metadata, it is feasible to construct
them with lambdas, including nice combinators for mapping, etc.:
```c++
add_library_task(map<unit>(error_already_reported(), [_d] (bool already_reported) {
    if (!already_reported && has_sorry(_d)) {
        report_message(...);
    }
    return unit();
}).depends_on(_d.is_theorem() ? _d.get_value_task() : nullptr));
```
(The snippet above causes `sorry` to be only reported at most once per
declaration, and only if no other error was reported.  It is also completely
deterministic.)

Tasks now have the the type `task<T>` (much easier to type).  Tasks are
typically constructed using the `task_builder<T>` helper.  A common operation is to
wrap the execution function, e.g. to set thread-local variables.  This is
accomplished using the `.wrap()` function; implementation-wise, each wrapper
adds a new heap-allocated `gtask_imp` object that delegates the execution and
dependency methods to the previous `gtask_imp`.  (We could avoid the
heap-allocation with some template trickery, but I'd rather wait for C++14 for
that--it has the equivalent of Rust's impl Trait feature.)

Dependencies are usually constructed with combinators as well, and are tasks as
well (with the generic `gtask` type).

In general, tasks are now no longer submitted immediately.  They are stored in
the `log_tree` for lazy evaluation--see the corresponding section for details.

## Module dependency graph

This part remains essentially unchanged.  In the `module_mgr` every module has
a list of dependencies.  We might want to use the cancellation graph to
invalidate modules as well, but this has the (positive?) side-effect that
invalidated modules are immediately cancelled.

I'm waiting with this refactoring until we have the parser monad.

## Message tree (`log_tree`)

The `log_tree` is a mutable, concurrent data structure.  It stores produced
messages, the info managers, as well as the tasks that need to be executed to
produce these messages.  More concretely, a node of the `log_tree` has the
following fields:
```c++
name_map<node> m_children;

std::vector<log_entry> m_entries;  // e.g. messages

location m_location;
std::string m_description;

gtask m_producer;
```

Every node has an associated location (file name + position range). The tree is
structured in a way such that the children's location is always a subset of the
parent's location.  We could enforce this invariant, as well as that messages
need to be contained in the node's location.  This would solve the problem where
we run a command, and report error messages in a different part of the file.

The tree is designed for lazy evaluation: we first insert nodes that contain
unsubmitted tasks. The tasks are then submitted by a separate listener, and only
when the editor/command-line requests the error messages for that location. When
a task is finished, it resets the producer field in its node to null. We only
modify nodes where an ancester has a non-null producer. Nodes that have a
non-null producer are shown as in-progress in the editor.

When sending information to the editor, we simply traverse this tree.  (Recall
that we only send messages for the currently visible region, so this is pretty
efficient. We don't even look at subtrees whose location is disjoint.)
```c++
std::vector<message> msgs;
m_lt->for_each([&] (log_tree::node const & n) {
    if (roi.intersects(n)) {
        for (auto & e : n.get_entries()) {
            if (auto msg = dynamic_cast<message const *>(e.get())) {
                if (roi.intersects(*msg))
                    msgs.push_back(*msg);
            }
        }
        return true;
    } else {
        return false;
    }
});
```

The progress messages at the command line and in the server mode are obtained by
traversing the `log_tree` as well, instead of looking at the state of the task
queue.

Tasks are typically inserted by using the `add_library_task<T>` function.

## Cancellation graph

At the moment, we cancel tasks based on their position in the file, by storing
the file name and position in every task, and sending a predicate to the task
queue when we want to cancel something.  This PR takes an approach inspired by
vscode, where every command gets an explicit `CancellationToken` object, where
you can attach listeners. We have a thread-local variable for a
`cancellation_token`, which can be cancelled and where we can add children.

It is notable to describe the ownership convention: parents hold weak references to
the children, children hold (strong) references to the parents. This ensures,
that as long as you have a `cancellation_token`, you will receive cancellation
events from every parent, without creating memory leaks (by never freeing
children).

## To do

* ~Different priorities for the tasks. Currently we prioritize parsing over
  elaboration. This has been lost in the refactoring. It does not seem to be a
  major issue for now.~

* ~Do not elaborate proofs in dependent modules. With this PR, this would happen
  for a "stupid" reason: we check for every import whether it contains sorry.
  This causes the proofs to be elaborated. We just need to find a clean way to
  skip this sorry check, probably only when "checking visible lines".~
  
* Once we refactor the parser to process one command at a time, I'd like to
  schedule one parsing task per command. This brings a few immediate advantages:
  * Better visual feedback about the parsing status.
  * We can stop parsing right below the edge of the screen. Parsing is
    automatically resumed when scrolling down.
  * We can reuse snapshots even when we did not finish parsing the file. Right
    now we only get the snapshots after we finish parsing the whole file.

## Performance

There is no (significant) performance change in either direction when compiling
the standard library. The editor interface seems to be more responsive now.


Kha:
> Every change uses an additional few hundred megabytes for me. I would really
> appreciate any ideas on how to debug this.

Still reading through the RFC, but https://clang.llvm.org/docs/LeakSanitizer.html might be worth a try.

leodemoura:
> There is one nasty memory leak though (that is also in master): open
init/core.lean and standard.lean. Then change something in init/core.lean.
Every change uses an additional few hundred megabytes for me. 

I tried a few experiments. 

- Disabled memory pools and the small object allocators (they could be fragmenting memory, and/or preventing memory from being reclaimed). No change: memory consumption keeps increasing.

- Printed amount of memory released by memory pools and small object allocators (in the end of execution). The amount of reclaimed memory increases as memory consumption increases. My point here is that I don't think the issue is due to a memory leak. I think a module may be preventing memory from being reclaimed because it contains references to many live objects. Today, I will add a compilation flag that when enabled will display the total amount of memory reclaimed by all these subsystems. 

- I disabled `g_expr_cache_enabled`, and other caches. No change: memory consumption keeps increasing.

- The memory keeps increasing even if I instruct emacs to use `-j0 -T0 -M0`. 

- I tried valgrind, but it detected only minor leaks (the ones I posted before).

leodemoura:
> @leodemoura quote1 was an easy fix, but exact_perf.lean reports an error
because try { exact h2 } fails, and h2 has an elaboration error. You
added this error message at
add5266
Any ideas?

I will take a look.

gebner:
> I tried a few experiments.

I tried adding this in module_mgr on file invalidation:
```c++
m_modules.clear();
```

When I check in the debugger, the task queue is empty, there is just a single loaded module (init.core), log_tree is also cleared, yet we're using a gigabyte of memory.  I really can't see where we still keep references to all that memory.

leodemoura:
> When I check in the debugger, the task queue is empty, there is just a single loaded module (init.core), log_tree is also cleared, yet we're using a gigabyte of memory. I really can't see where we still keep references to all that memory.

It could be the thread local caches.
We may consume a lot of memory by storing references to a few different `environment` objects,

leodemoura:
I have added the following compilation flags 

1)
```
-D TRACK_CUSTOM_ALLOCATORS=ON
```
Tracks the amount of memory deallocated by `memory_pool` and `small_object_allocator`.

2)
```
-D TRACK_LIVE_EXPRS=ON
```
Enable counter for the number of live `expr`

BTW, Lean displays the amount of deallocated memory and live expressions before/after finalization.

3) 
```
-D CUSTOM_ALLOCATORS=OFF
```
Disable `memory_pool` and `small_object_allocator`.

Here is some data for an example that consumes 1Gb by making simple modifications to `tactic.lean` while `vector.lean` is open:

a) with `-D TRACK_CUSTOM_ALLOCATORS=ON` and `-D TRACK_LIVE_EXPRS=ON`
```
memory deallocated by memory_pool and small_object_allocator (before finalization): 117205272
number of live expressions (before finalization): 105
memory deallocated by memory_pool and small_object_allocator (after finalization): 608798768
number of live expressions (after finalization): 0
```
So, the custom allocators have released 608Mb in the end of the execution. Note that this amount is not including the overhead introduced by `malloc`. I believe this is evidence that the high memory consumption is **not** due to a memory leak. 608Mb is a lot of memory. Moreover, when I edited the file manually, the memory consumption was around 250Mb after the first modification.

b) With `-D CUSTOM_ALLOCATORS=OFF` (memory consumption goes down to 828Mb). However this is expected since we may have unused space in pools for different threads. 828Mb is still a lot of memory.

gebner:
The compilation flags are nice.  I just had to move them around and make them write to stderr, so that I can see them from emacs.  I also added an extra output right before thread finalization:
```
memory deallocated by memory_pool and small_object_allocator (before finalization): 96532824
number of live expressions (before finalization): 4309228
memory deallocated by memory_pool and small_object_allocator (after finalization): 96532824
number of live expressions (after finalization): 4309140
memory deallocated by memory_pool and small_object_allocator (after thread finalization): 173477928
number of live expressions (after thread finalization): 4309140
```

(This is from while the server is running.)  So it is really the thread local data that uses up memory.  However all worker threads are stopped as soon as they are idle.  The lthread objects are destroyed as well (there was a reference cycle there), but that doesn't help either.

leodemoura:
I have just added `-D SAVE_INFO=OFF` and `-D SAVE_SNAPSHOT=OFF`.
Then, I tried the following experiment with the setting
```
-D SAVE_INFO=OFF -D SAVE_SNAPSHOT=OFF -D CUSTOM_ALLOCATORS=OFF
```
and lean server running with `-j2`.
Then, I tried the usual example two files (`tactic.lean` and `bitvec.lean`), and kept adding/fixing an error in `tactic.lean`.
Initially memory consumption stabilized at 1.05Gb. Then, I started adding different kinds of errors, and it started to increase again. However, it stabilized many times. I managed to get it to 1.8Gb but it took hundreds of modifications. With the `info_manager` and `snapshots` enabled, I get to 2Gb very quickly.

gebner:
Hmm, apparently all the thread-local data is on the *main* thread: if I comment out the following line, then `delete_thread_finalizer_manager` does not change the value of `get_memory_deallocated`:
```
    ~thread_finalizers_manager() {
//        finalize_thread(get_pair()); // finalize main thread
        pthread_key_delete(g_key);
    }
```
That is, I get:
```
memory deallocated by memory_pool and small_object_allocator (before finalization): 104291536
number of live expressions (before finalization): 125
memory deallocated by memory_pool and small_object_allocator (after finalization): 104291536
number of live expressions (after finalization): 0
memory deallocated by memory_pool and small_object_allocator (after thread finalization): 104291536
number of live expressions (after thread finalization): 0
```

@Kha LeakSan is awesomely fast, thanks for pointing me to it.  However it doesn't detect the leak above.

leodemoura:
@gebner I tried to reproduce your last experiment. Here is the data I get (I'm using the master branch)
Before:
```
% memusg ../../bin/lean -j 2 --server=../../tmp/trace2.txt > /dev/null
memory deallocated by memory_pool and small_object_allocator (before finalization): 375631624
number of live expressions (before finalization): 105
memory deallocated by memory_pool and small_object_allocator (after finalization): 375631624
number of live expressions (after finalization): 105
memory deallocated by memory_pool and small_object_allocator (after thread finalization): 636353912
number of live expressions (after thread finalization): 0
memusg: peak=974244
```
After I commented the line you have referenced in your message
```
memusg ../../bin/lean -j 2 --server=../../tmp/trace2.txt > /dev/null
memory deallocated by memory_pool and small_object_allocator (before finalization): 325759920
number of live expressions (before finalization): 105
memory deallocated by memory_pool and small_object_allocator (after finalization): 325759920
number of live expressions (after finalization): 105
memory deallocated by memory_pool and small_object_allocator (after thread finalization): 325759920
number of live expressions (after thread finalization): 105
~/projects/lean/build/track$ memusg: peak=889312
```

leodemoura:
@gebner Here is a conjecture:
1- worker threads allocate many `expr`s and other kinds of object using the `memory_pool`. 
2- the memory pool for the worker thread is often empty, and it requests more memory using `malloc`
3- some/most of these object are deallocated by the main thread. These objects are stored in the main thread `memory_pool`.
4- The main thread `memory_pool` keeps growing since the main thread doesn't do much. This is why we have a 300Mb in the `memory_pool` of the main thread in the last experiment.
What do you think?

I will test this conjecture by putting a limit on the `memory_pool` size.

leodemoura:
@gebner Here is the same experiment, but `memory_pool`s can have at most 1024 objects.
```
memusg ../../bin/lean -j 2 --server=../../tmp/trace2.txt > /dev/null
memory deallocated by memory_pool and small_object_allocator (before finalization): 10004080
number of live expressions (before finalization): 105
memory deallocated by memory_pool and small_object_allocator (after finalization): 10004080
number of live expressions (after finalization): 105
memory deallocated by memory_pool and small_object_allocator (after thread finalization): 13335936
number of live expressions (after thread finalization): 0
memusg: peak=1048732
```
We are still consuming a lot of memory 1Gb, but the main thread is holding only 3Mb instead of 300Mb during finalization. 
When using the editor, the memory seems to stabilize around 1.5Gb. It keeps growing but in a much slower pace, and I hit several plateaus where the memory doesn't increase.

Another thing to keep in mind that there is overhead and fragmentation going on in the `malloc` provided by the runtime. I will keep running more experiments.

gebner:
> 3- some/most of these object are deallocated by the main thread. These objects are stored in the main thread memory_pool.

This is very true.  We deallocate all the environments on the main thread, as well as the info managers, etc.

I tried the memory_pool change, but it doesn't change the memory consumption for me.  I found a different memory leak though that sometimes happens at the end of a server process, I'm looking into it right now.

leodemoura:
> I tried the memory_pool change, but it doesn't change the memory consumption for me.

Yes, the memory consumption is high, but it seems to stabilize for me. 
Does it still keep growing for you? For me, it stops around 2Gb.
Moreover, it doesn't grow as fast as before. Before the `memory_pool` commit, every modification would consume between 100Mb and 200Mb.

The commit also seems to address the following observation you made:

> Hmm, apparently all the thread-local data is on the main thread: if I comment out the following line, then delete_thread_finalizer_manager does not change the value of get_memory_deallocated:

Do you still observe this behavior after the `memory_pool` commit? 
I cannot observe it in the `master` branch.

> I found a different memory leak though that sometimes happens at the end of a server process, I'm looking into it right now.

Does this leak affect the `master` branch too?
Valgrind is not reporting any significant memory leak in the `master` branch when I tried it yesterday.
There were minor (a few kilobytes) in the server code, but they do not explain the crazy memory consumption.

gebner:
> Does it still keep growing for you? For me, it stops around 2Gb.

It still keeps growing for me.  I need to test master again to see the difference.

> Do you still observe this behavior after the memory_pool commit?

No, the main thread finalizer does not change the deallocation number anymore, so that seems to be fixed.

gebner:
> Does it still keep growing for you? For me, it stops around 2Gb.

Okay, there were quite a few leaks all over the place.  Now it stops at 1.5G for me as well.

gebner:
I've added the region-of-interest support to the vscode extension as well: https://github.com/leanprover/vscode-lean/compare/master...gebner:parallel2

Unfortunately, it's not possible to determine which lines are visible with the current vscode API: https://github.com/Microsoft/vscode/issues/14756

leodemoura:
> @leodemoura quote1 was an easy fix, but exact_perf.lean reports an error
because try { exact h2 } fails, and h2 has an elaboration error. You
added this error message at
add5266
Any ideas?
I will take a look.

I investigated this one. However, I didn't find a good solution yet (only hacks).
Here is the problem:
When we parse tactic notation, we statically know whether we should report errors or not.
Example:
```
begin
     tac_1, tac_2
           --^ if tac_2 fails, we want the red squiggly line here
end
```
On the other hand,
```
begin
     tac_1, try {tac_2}
                --^ we don't want a red squiggly line here if tac_2 fails
end
```
So, we use the `tactic.rstep` and `tactic.istep` combinators for controlling this.
The `tactic.rstep line col tac` will report the error at the given line and column if `tac` fails.
The tactic notation module decides which one should be used.
This works for everything, **but** `to_expr`.
The elaboration errors are reported by the elaborator depending on the flag `report_errors`.
Right now, `tactic.interactive.exact` is defined as
```
/- doesn't allows metavars and report errors -/
meta def i_to_expr_strict (q : pexpr) : tactic expr :=
to_expr q ff tt -- << the `tt` means report errors

meta def exact (q : parse texpr) : tactic unit :=
do tgt : expr ← target,
   i_to_expr_strict ``(%%q : %%tgt) >>= tactic.exact
```

leodemoura:
Here is one ugly hack to fix the problem above.
We add a combinator `tactic.ignore_elab_errors tac` which will execute `tac` without reporting elaboration errors even if `tac` uses `to_expr e ff tt -- << tt here means "report_errors"`.

gebner:
I think that in the long run all error messages during tactic execution should be kept in the tactic_state.  Then `<|>` behaves as expected and suppresses error messages when backtracking.  Even if we add `ignore_elab_errors` for `try`, we'd still have the problem with `<|>`.

If I recall correctly, the main motivation for `rstep` to ignore the exception position, and then for `to_expr` to directly report an error is because of the issue in `quote_error_pos.lean`: if we call to_expr on a pre-expression from another definition, the we might get elaboration errors at that definition, and not at the tactic use site.
This issue was the reason we now keep track of position ranges (begin and end of the current command) in the log_tree data structure; we could take the position if it lies inside the range (when the pre-expression comes from the current tactic block), or assign it another useful position if it lies outside (when it comes from another definition).

For now, the `ignore_elab_erors` hack is probably the easiest and would fix the test and the `try` problem.  I'm on vacation for a week, I can solve this afterwards in a nicer way.

leodemoura:
> I think that in the long run all error messages during tactic execution should be kept in the tactic_state. 

This would be bad. The error messages are thunks, because we want to discard them at `<|>`. This is important for performance. We originally did not use thunks and it was a performance bottleneck.
It is also problematic to store the thunk since we would have to use the thread safe `vm_obj`.

> Then <|> behaves as expected and suppresses error messages when backtracking. Even if we add ignore_elab_errors for try, we'd still have the problem with <|>.

This is a bug in the current support for `<|>` in interactive mode. I can fix this, and it will work with `ignore_elab_errors`.

gebner:
> This would be bad. The error messages are thunks, [...]

I did not suggest to switch from thunks to error messages.  I was referring to the fact that some operations like `add_declaration` and `to_expr` directly report errors that remain after backtracking.  And that those error messages should go into the tactic state instead (as thunks).

leodemoura:
> And that those error messages should go into the tactic state instead (as thunks).

This problematic since to put them into the `tactic_state` we need to convert the thunks into thread safe `vm_obj`s (which deep copy `vm_obj`s closures and constructors).

Kha:
I'm not sure I'm following yet - why would we ever need to report errors immediately or store them in the tactic state, instead of returning them in `result.exception`?

gebner:
> I'm not sure I'm following yet - why would we ever need to report errors immediately or store them in the tactic state, instead of returning them in result.exception?

In very rare cases we produce warnings instead of errors, for example when adding declarations with the wrong noncomputability annotation.  The elaborator can also produce more than one error (when error recovery is enabled).  Maybe we want to produce more warnings in the future.

gebner:
@leodemoura Did you already start with `ignore_elab_errors`?

I think there are two options here:
1. Modify `to_expr` so that it doesn't report errors directly, but makes sure that they are inside the current declaration.  Then we can remove `rstep`.
2. Add an `ignore_elab_errors` combinator that sets a thread-local flag that causes `to_expr` to not report errors directly.  (Or a flag in the tactic state instead?)  This flag is always set on the left branch of `<|>` (also in the non-interactive version).  Maybe rename `ignore_elab_errors` to `possible_backtracking`; we could enable error recovery if we're not backtracking.

Any opinions?  Should we do 1, 2, or 1+2?

gebner:
So, I found another possible explanation for the high memory usage, this is the output of `malloc_stats`:
```
Total (incl. mmap):
system bytes     = 1095458816
in use bytes     =  533526976
max mmap regions =          2
max mmap bytes   =    8396800
```
That is, the Lean process uses 1095M of memory but we have only malloc'd 533M.  The rest seems to be due to fragmentation and/or glibc's reluctancy to free memory.  No idea yet on how to improve that.

gebner:
One possibility: we switch to jemalloc.  It has a nice API to query how much memory we use, and it also doesn't seem to fragment that much.

With jemalloc, when switching between `init/core.lean` and `standard.lean`, we allocate 700M, and use 1200M (after a few iterations).

Kha:
> In very rare cases we produce warnings instead of errors, for example when adding declarations with the wrong noncomputability annotation. The elaborator can also produce more than one error (when error recovery is enabled). Maybe we want to produce more warnings in the future.

I see, thanks. I do wonder if we can't come up with simpler implementations, however. For example, for warnings we could traverse and check all new declarations only after the top-level command (eg `run_cmd`) has finished.

> Modify to_expr so that it doesn't report errors directly, but makes sure that they are inside the current declaration. Then we can remove rstep.

Sounds perfect. Are there any downsides?

Kha:
Btw, gperf (and jemalloc in particular) seem to have some nice heap profilers: https://github.com/jemalloc/jemalloc/wiki/Use-Case%3A-Leak-Checking

gebner:
> I see, thanks. I do wonder if we can't come up with simpler implementations, however. For example, for warnings we could traverse and check all new declarations only after the top-level command (eg run_cmd) has finished.

The noncomputability/sorry/type-checking errors are a bit of a hack at the moment, the otherwise pure `module::add` functions directly reports them--even if we never use the resulting environment.  I'm open for better suggestions.  The problem with checking the new declarations after the top-level command is mainly that we don't really know which declarations are new.  But looking at the code, these checks seem to be better at home in the definition command, then you couldn't circumvent them using `environment::add` either.

> > Modify to_expr so that it doesn't report errors directly, but makes sure that they are inside the current declaration. Then we can remove rstep.
> Sounds perfect. Are there any downsides?

The issue with the easy one-line implementation is that the error would appear at the begin of the declaration, not at the tactic.

gebner:
Ok, I just looked a bit through the tactic notation code.  I'm surprised to see that we apparently directly report errors in `rstep`, completely bypassing the built-in exception handling of the tactic monad.

I don't really remember the rationale for this behavior.  Why can't we just replace `rstep` with `istep`?  Concretely, my proposal would be the following:
 * Change `istep` to have `begin_line begin_col end_line end_col : nat` arguments, with the intention that error positions in the exception case should be adjusted to lie in the given range.
 * Remove `rstep`, `report_error`, as well as `silent_fail`.
 * Use `step` instead of `istep` in auto-quote blocks.

Kha:
Still sounds good to me - do we actually need to clamp downwards (end_*)
too though? (edit: we do in recursive tactics I guess :smile: )

gebner:
> Still sounds good to me - do we actually need to clamp downwards (end_*)

I have now implemented this approach.  I left out the downward clamping for simplicity: in the worst case, `report_message` will do some clamping anyhow (to the current declaration, basically).  The code is quite a bit simpler now.

> (edit: we do in recursive tactics I guess ￼ )

I'm not sure I follow.  We don't use `istep` in auto-quote blocks (`` `[begin ... end] ``), so this shouldn't happen.

leodemoura:
I will play with this branch soon. 
In the meantime, I will just write a few use cases and mark them as OK or BUG based on the master branch. Some of the them used to work in the master branch, but don't work anymore (e.g., `repeat`).

```lean
example (p q : Prop) (h1 : p) (h2 : q) : p ∧ q :=
begin
  exact (or.inl _)
       --^ type mismatch here (OK)
end

example (p q : Prop) (h1 : p) (h2 : q) : p ∧ q :=
begin
  exact (and.intro _ h2)
                 --^ don't know how to synthesize placeholder here (OK)
end

example (p q : Prop) (h1 : p) (h2 : q) : p ∧ q :=
begin
  abstract {exact (and.intro _ h2)}
                           --^ don't know how to synthesize placeholder here (OK)
end

example (p q : Prop) (h1 : p) (h2 : q) : p ∧ q :=
begin
  {exact (and.intro _ h2) } <|> apply and.intro
                  --^ should not show error here (BUG)
end

example (p q : Prop) (h1 : p) (h2 : q) : p ∧ q :=
begin
  try {exact (and.intro _ h2) }
                      --^ should not show error here (BUG)
end

example (p q : Prop) (h1 : p) (h2 : q) : p ∧ q :=
begin
  repeat {exact (and.intro _ h2) }
                         --^ should not show error here (BUG)
end

open tactic
meta def mytac : tactic unit :=
`[apply (or.inl _)]

example (p q : Prop) (h1 : p) (h2 : q) : p ∧ q :=
begin
  mytac
--^ failed to unify error here, it shoult *not* be at mytac or.inl (OK)
end
```

gebner:
The `import foo uses sorry` warnings are now disabled in the editor by default, since they would require elaboration of all proofs in the dependent modules.  Nothing has changed with `lean --make`, they are still there.  It is easy enough to re-enable it in some cases (for example if "check all files" is selected), but I'd like to hear how people use them first.

gebner:
I just found another problem with show-goal: apparently the goal is rendered in the empty environment (or maybe with pp.all set, but I don't know where that would come from).

Kha:
> I just found another problem with show-goal: apparently the goal is rendered in the empty environment (or maybe with pp.all set, but I don't know where that would come from).

It's using `get_closest_snapshot`. Should we fall back to `m_loaded_module` (when available) when there are no snapshots?

gebner:
> It's using get_closest_snapshot. Should we fall back to m_loaded_module (when available) when there are no snapshots?

The environment and/or snapshots should not make a difference, since the environment at the time of the goal is stored in `vm_obj_format_info`.

leodemoura:
> I just found another problem with show-goal: apparently the goal is rendered in the empty environment (or maybe with pp.all set, but I don't know where that would come from).

> The environment and/or snapshots should not make a difference, since the environment at the time of the goal is stored in vm_obj_format_info.

@gebner It is not clear to me what the problem is. The environment is stored in the goal, and your second comment makes it clear you are aware of this fact.

leodemoura:
> The import foo uses sorry warnings are now disabled in the editor by default, since they would require elaboration of all proofs in the dependent modules. 

Good for me.

gebner:
@leodemoura This is the problem (as seen from emacs), I'm still trying to figure out what's going on:
```
α : Type.{u},
_inst_1 : discrete_linear_ordered_field.{u} α,
a b : α,
ha :
  gt.{u} α (strong_order_pair.to_has_lt.{u} α (linear_strong_order_pair.to_strong_order_pair.{u} α (linear_ordered_ring.to_linear_strong_order_pair.{u} α (linear_ordered_field.to_linear_ordered_ring.{u} α (discrete_linear_ordered_field.to_linear_ordered_field.{u} α _inst_1))))) a (zero.{u} α (no_zero_divisors.to_has_zero.{u} α (integral_domain.to_no_zero_divisors.{u} α (discrete_field.to_integral_domain.{u} α (discrete_linear_ordered_field.to_discrete_field.{u} α _inst_1) (discrete_linear_ordered_field.to_discrete_field.{u} α _inst_1))))),
h :
  le.{u} α (weak_order.to_has_le.{u} α (strong_order_pair.to_weak_order.{u} α (linear_strong_order_pair.to_strong_order_pair.{u} α (linear_ordered_ring.to_linear_strong_order_pair.{u} α (linear_ordered_field.to_linear_ordered_ring.{u} α (discrete_linear_ordered_field.to_linear_ordered_field.{u} α _inst_1)))))) (div.{u} α (division_ring_has_div.{u} α (field.to_division_ring.{u} α (linear_ordered_field.to_field.{u} α (discrete_linear_ordered_field.to_linear_ordered_field.{u} α _inst_1))) (field.to_division_ring.{u} α (linear_ordered_field.to_field.{u} α (discrete_linear_ordered_field.to_linear_ordered_field.{u} α _inst_1)))) (one.{u} α (zero_ne_one_class.to_has_one.{u} α (division_ring.to_zero_ne_one_class.{u} α (field.to_division_ring.{u} α (linear_ordered_field.to_field.{u} α (discrete_linear_ordered_field.to_linear_ordered_field.{u} α _inst_1)))))) a) b
⊢ le.{u} α (weak_order.to_has_le.{u} α (strong_order_pair.to_weak_order.{u} α (linear_strong_order_pair.to_strong_order_pair.{u} α (linear_ordered_ring.to_linear_strong_order_pair.{u} α (linear_ordered_field.to_linear_ordered_ring.{u} α (discrete_linear_ordered_field.to_linear_ordered_field.{u} α _inst_1)))))) (div.{u} α (division_ring_has_div.{u} α (field.to_division_ring.{u} α (linear_ordered_field.to_field.{u} α (discrete_linear_ordered_field.to_linear_ordered_field.{u} α _inst_1))) (field.to_division_ring.{u} α (linear_ordered_field.to_field.{u} α (discrete_linear_ordered_field.to_linear_ordered_field.{u} α _inst_1)))) (one.{u} α (zero_ne_one_class.to_has_one.{u} α (division_ring.to_zero_ne_one_class.{u} α (field.to_division_ring.{u} α (linear_ordered_field.to_field.{u} α (discrete_linear_ordered_field.to_linear_ordered_field.{u} α _inst_1)))))) b) a
```

leodemoura:
@gebner Can you reproduce the problem in batch mode? If you can, one option is to put a breakpoint at `format tactic_state:pp() const` at `tactic_state.cpp`. Then, we can see whether the options are correct there or not. We can also put a breakpoint at 
```
vm_obj tactic_save_info_thunk(vm_obj const & pos, vm_obj const & thunk, vm_obj const & s) {
```
at `info_manater.cpp`. This is invoked when we save the goal information. We can compare the `options` object at saving and pp times.

gebner:
> @gebner Can you reproduce the problem in batch mode?

It does not happen with `-j0`.  I'll debug further when I get home.

gebner:
The issue was that the global ios was not set, and we would hence fall back to the dummy ios without the fancy formatter.  It's fixed now.

Kha:
@gebner I just pushed a small Emacs 24 fix (always wanted to try out this Github feature :) )

gebner:
@kha Thanks!  I didn't realize you could push to other people's repositories.  This only works for PRs, right?

Kha:
> This only works for PRs, right?

Yes, I would hope so :)

Kha:
> The import foo uses sorry warnings are now disabled in the editor by default, since they would require elaboration of all proofs in the dependent modules.

I'm a bit anxious about this change. A colleague just asked me whether Lean has a sorry command just like Isabelle - he had proved something in Isabelle, which turned out to still depend on a sorry, but he didn't notice because there is no warning at all in Isabelle (compiling on the command line, however, would give an error by default). How about an opt-out setting that elaborates such proofs at lowest priority?
If we have fully checked a dependent module that does not depend on any opened modules, it may also make sense to write out an olean file even in server mode. I have accidentally processed the whole standard library when working in another project on more occasions than I can count, only to redo all those computations when changing to the next project (e.g. another `tests/` folder).

johoelzl:
@Kha I absolutely agree. I also find the behavior of Isabelle to *not* show `sorry` (neither as a dot in a scroll bar nor as warning at the import statement) very annoying.

Would it be possible to set a /tainted/ flag or something in the `olean` file after it was produced? Would it be possible to add a special /end-of-file/ block describing the used axioms?

gebner:
> he had proved something in Isabelle, which turned out to still depend on a sorry, but he didn't notice because there is no warning at all in Isabelle

Ok, maybe I should reiterate some of the changes in this PR.  By default, we only show errors for the *currently visible lines*.  We do this for a few reasons.
  * Flycheck.  Updating many errors makes emacs unresponsive.  The easiest way to improve flycheck performance is to show fewer errors.
  * Long files.  If we always check the whole file, then working on the beginning of a long file is very uncomfortable.  Only checking the first part improves responsiveness.  This will get even better when we stop parsing as soon as we leave the visible part.
  * Consistency.  We could show all produced errors for the current file, even if they are outside of the part that we actively check.  But I feel that this would be even more confusing, since whether or not these errors are produced depends on where you have scrolled before.  (E.g. you would not normally see an error in a proof at the end of a file, but you would see it if you had previously scrolled to the end and back.)

Now, this behavior is configurable.  You can switch to "check visible file" mode, which checks (and shows errors for) the whole file.  I think including sorry checks would fit well in this extended mode.

My intention was that "check visible lines" would be used when you're working on a proof and just want to see errors in the current proof and have the editor be as responsive as possible.  You'd then switch to "check visible files" to get a more global overview.

> I have accidentally processed the whole standard library when working in another project on more occasions than I can count,

The change with regards to sorry reporting addresses precisely this issue, but in a slightly different way: we no longer process the whole standard library.  Essentially all proofs get skipped.
This may be a bit overkill for the standard library where we have mostly definitions, but I think it will greatly improve responsiveness in formalizations with many proofs.

gebner:
> Would it be possible to set a /tainted/ flag or something in the olean file after it was produced?

Yes, we have such a flag.  But in general, the files containing sorry will not be loaded from olean anyhow.

> Would it be possible to add a special /end-of-file/ block describing the used axioms?

Do you want to see import statements annotated with "uses propext"?  Or would you like to see warnings for theorems that depend on axioms?  I think this would result in a huge number of warnings.  We always have `#print axioms`, which will also tell you if a proof uses sorry.

Or maybe I'm misunderstanding you: would you like the editor modes to insert an uneditable block of text at the end of each module summarizing the used axioms?

gebner:
> How about an opt-out setting that elaborates such proofs at lowest priority?

Just to make a concrete proposal:
 * "check visible lines" would skip the sorry check.
 * "check visible files" would perform them, but at lowest priority.

@kha If I understand you correctly, you'd like "check visible files" to be the default.  This would also show all errors for the visible files, not restricting them to the currently visible lines.

Kha:
> @kha If I understand you correctly, you'd like "check visible files" to be the default

I'm fine with "visible lines", I just think we shouldn't hide errors by default in any non-obvious way. So perhaps we should just be a bit more verbose, where applicable:

* add an info message "dependencies not checked/still checking: ..." at the beginning of the file
* add an info message "checked up to here" at the end of the ROI (Isabelle uses its annotated scroll bar for that)

johoelzl:
> Yes, we have such a flag. But in general, the files containing sorry will not be loaded from olean anyhow.

Ah, then it is mostly a misunderstanding on my side. Now I understand @Kha's comment on elaborating the dependent modules on low priority. And I also think the "check visible files" should contain a check on low-priority. (edited)

My suggestion with the list of used axioms was more a generalization of the flag. So Lean would check at the import if `sorry` was in the list of axioms. And something like `#print axioms` would work without full elaboration. And you are right it would be only useful for `propext`. `sorry` is reported anyway, and `choice` is marked with `noncomputable`.

When the dependent modules are not fully elaborated, what is the current behavior of `#print axioms`? Does it show a warning that not all theorems are fully elaborated?

Btw, I didn't test this PR yet, but I love its idea :-)

gebner:
> When the dependent modules are not fully elaborated, what is the current behavior of #print axioms? Does it show a warning that not all theorems are fully elaborated?

The output of `#print axioms my_lemma` is completely deterministic (and will also include user-defined axioms, in case it wasn't clear).  Right now it will block parsing until all the referenced proofs have been elaborated, but it is not difficult to change it so that this becomes async.

Kha:
> We could show all produced errors for the current file, even if they are outside of the part that we actively check.

By the way, that is Isabelle's behavior and I find it very acceptable (especially with an indicator of what parts of the file have been processed). Conversely, I just got very confused by an error in Lean not showing up because it was above the ROI.

gebner:
> Conversely, I just got very confused by an error in Lean not showing up because it was above the ROI.

I agree.  The default is now to show all available error messages, and show the hourglass icon if anything in the file still needs to be processed.

gebner:
@leodemoura How do we proceed with this branch?  BTW, I added [your examples](https://github.com/leanprover/lean/pull/1405#issuecomment-286833474) as a test case and everything works as expected.

kim-em:
@gebner, I just tried out this branch with my code. I can get the Lean server to consistently crash, by opening VS Code with just [monoidal_categories/monoidal_category.lean](https://github.com/semorrison/lean-category-theory/blob/master/monoidal_categories/monoidal_category.lean) open. It parses for a few seconds and then crashes. It's fine on the master branch.

    git clone https://github.com/semorrison/lean-category-theory
    cd lean-category-theory
    git reset --hard 7cf08593e2b1749105d68fc9b88c32a6c6c403a4

gebner:
@semorrison Thanks for testing!  This branch requires an updated version of the vscode extension from https://github.com/gebner/vscode-lean/tree/parallel2
But I just noticed that the updated extension no longer works with the current state of the branch, I'll fix it and notify you when it should work again.

leodemoura:
@gebner I can merge it today. Is it ready?

gebner:
@leodemoura Great.  I'm still tracking down the bug reported by Scott.  Aside from that the branch is ready.

gebner:
@leodemoura The bug is fixed, feel free to merge.
@semorrison vscode now works again.

gebner:
Just so that I don't lose track of it: I think this is the upstream vscode issue that would allow us to implement the "visible lines" mode. https://github.com/Microsoft/vscode/issues/588

Edit: here is yet another issue for it.  https://github.com/Microsoft/vscode/issues/17362

Issue #1385, opened 2017-02-18
Equation compiler failure: unknown variable


dselsam:
```Lean
def S := list ℕ
def T (s : S) : Type := sorry

inductive op : Π (ishapes : list S) (oshape : S),  Type
| binary : Π (shape : S), op [shape, shape] shape
| gemm : Π {m n p : ℕ}, op [[m, n], [n, p]] [m, p]

namespace op

set_option trace.eqn_compiler true

def f : Π {ishapes : list S} {oshape : S}, op ishapes oshape → T oshape
| [shape, .shape] .shape (binary .shape) := sorry
| [[m, n], [.n, p]] [.m, .p] gemm := sorry

end op
```
yields:
```
/home/dselsam/projects/lean/issues/eqn_compiler.lean:12:4: error: equation compiler failed to create auxiliary declaration 'op.f._main'
nested exception message:
infer type failed, unknown variable
  p
```
It works if I change the result type from `T oshape` to something that does not depend on `oshape`, i.e. `nat`.

Here is the full trace:
```
[eqn_compiler] compiling
def f : Π {ishapes : list S} {oshape : S}, op ishapes oshape → T oshape
| [shape, .shape] .shape (binary .shape) := sorry
| [[m, n], [.n, p]] [.m, .p] gemm := sorry
[eqn_compiler] recursive: 0
[eqn_compiler] nested recursion: 0
[eqn_compiler.elim_match] depth [1]
match f [ishapes, oshape, a]
  ([shape, .shape]) (.shape) (binary .shape) := sorry
  ([[m, n], [.n, p]]) ([.m, .p]) (gemm) := sorry
[eqn_compiler.elim_match] step: constructors only
[eqn_compiler.elim_match] depth [2]
match f.nil [oshape, a]
[eqn_compiler.elim_match] step: constructors only
[eqn_compiler.elim_match] step: variables only
[eqn_compiler.elim_match] depth [3]
match f.nil [a]
[eqn_compiler.elim_match] step: constructors only
[eqn_compiler.elim_match] depth [2]
match f.cons [a, a_1, oshape, a_2]
  (shape) ([.shape]) (.shape) (binary .shape) := sorry
  ([m, n]) ([[.n, p]]) ([.m, .p]) (gemm) := sorry
[eqn_compiler.elim_match] step: variables and constructors
[eqn_compiler.elim_match] depth [3]
match f.cons [a, a_1, oshape, a_2]
[...]

Issue #1334, opened 2017-01-24
Equation compiler failure


robertylewis:
```lean
inductive nlist
| atom : nlist
| mk : list nlist → nlist

open nlist list

meta def fn : nlist → nlist
| (mk l) := mk [] 
| _ := atom
```
The equation compiler fails on `fn`, but I'm not sure how to interpret its output. It also fails with a generic pattern `e` in place of the placeholder in the last line, but succeeds if you replace the placeholder with `atom`.

jroesch:
This is a known limitation, the pattern match compiler currently can not compile wildcard patterns for nested inductive types.

If you do:

```lean
meta def fn : nlist → nlist
| (mk l) := mk [] 
| atom := atom
```

It should work, we hope to fix this soon, my understanding is the nested compiler needs to be extended.

leodemoura:
@jroesch Do you have "energy" to work with me on this issue next week? We are currently missing a lot of features for nested inductive datatypes. Example: the compiler generates very naive code for them (the problem we have discussed in Paris). The goal would be: finish support for well-founded recursion (needed when using nested inductives); "complete" transition for nested (this issue); avoid coercions from/to auxiliary datatypes produced by the inductive compiler when generating code; cleanup inductive compiler code.

PR #1406, opened 2017-02-28
Fix equations compilers elim_match: it fails when argument is variable or inaccessible term

johoelzl:
The `elim_match` part of the equations compiler fails when there are two equations, one where the argument is matched against a inaccessible pattern and one where the argument is matched against a variable. An example:
```
example {α : Type} {p q : α → Prop} {x : α} (h : ∀ y, p y → q y) (hx : q x) :
  ∀ y, x = y ∨ p y → q y
| x  (or.inr p)   := h x p
| ._ (or.inl rfl) := hx
```

It looks like when one inaccessible pattern is discovered, all substitutions for the all equations are thrown away. But for /true/ variables the substitution is still needed.

The change in this pull request fixes the problem, and all tests go through. Put it looks like there are not many tests for this part of the equations compiler and I'm not sure if I understand all the invariants for the `problem`s and `equation` in `elim_match`.

### DIFF
--- a/src/library/equations_compiler/elim_match.cpp
+++ b/src/library/equations_compiler/elim_match.cpp
@@ -601,7 +601,7 @@ struct elim_match_fn {
         for (equation const & eqn : P.m_equations) {
             equation new_eqn   = eqn;
             new_eqn.m_patterns = tail(eqn.m_patterns);
-            if (is_var_transition) {
+            if (is_var_transition || is_local(head(eqn.m_patterns))) {
                 new_eqn.m_subst  = add_subst(eqn.m_subst, head(eqn.m_patterns), head(P.m_var_stack));
             }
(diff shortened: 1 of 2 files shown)


leodemoura:
@johoelzl In the next two weeks, @dselsam will help us with problems in the inductive compiler, and the interface with the equation compiler. The plan is to fix issues like #1334
I will also take the opportunity to write an equation validator that will produce better error messages for invalid set of equations such as #1385.

> It looks like when one inaccessible pattern is discovered, all substitutions for the all equations are thrown away. But for /true/ variables the substitution is still needed.

This is correct.

> Put it looks like there are not many tests for this part of the equations compiler

It would be great if you could create tests for it that exercise corner cases. 
It would be great to have a test suite for the new equation validator.

PR #1407, opened 2017-03-01
replace strong indefinite description with indefinite description

avigad:
Moving the pieces around, we can reduce the choice principle to a simpler and prettier axiom.


avigad:
@digama0 Sorry, I meant to ping you on this, to see if this is the reduction you had in mind.

PR #1413, opened 2017-03-03
`expr` literals ```(...)

Kha:
(because more backticks, more good)

This change mostly reuses the existing logic for `expr` patterns to also support them as literal values. Literals will be stored efficiently in the existing `quote_macro`, instead of in a full reflected expression tree as for patterns.
Also change syntax for expr patterns to the same notation for consistency.


Kha:
@leodemoura Your example
```
```(@add_comm %%A %%inst %%a %%b)
```
compiles but introduces a new universe parameter. Should we check for that?

src/frontends/lean/parser.cpp
```diff
@@ -533,6 +533,10 @@ void parser::add_local_level(name const & n, level const & l, bool is_variable)
 }
 
 void parser::add_local_expr(name const & n, expr const & p, bool is_variable) {
+    if (!m_in_quote) {
+        // certainly not in a pattern
+        m_in_pattern = false;
```
Kha: @leodemoura I'm not proud of this hack. The issue is that in code like https://github.com/leanprover/lean/blob/d9da6f05b599fb4e1af1bba8b118993eec02f2e5/library/tools/super/prover.lean#L57, it's too late to elaborate the quotation in `patexpr_to_expr`. We need to know whether we're in a pattern before we leave the binder.

library/init/meta/expr.lean
```diff
@@ -282,18 +284,8 @@ meta def binding_body : expr → expr
...
 meta def imp (a b : expr) : expr :=
-pi `a binder_info.default a b
-
-meta def and_ (a b : expr) : expr :=
-app (app (const ``and []) a) b
-
-meta def not_ (a : expr) : expr :=
-app (const ``not []) a
-
-meta def false_ : expr := const ``false []
```
Kha, line 296: I'm glad to get rid of most of these functions, but `imp` turned out to be just too convenient.

leodemoura: me too!

src/frontends/lean/parser.cpp
Kha: Ah well, it does break the error message in https://github.com/leanprover/lean/blob/master/tests/lean/inaccessible2.lean#L11. Suggestions?

leodemoura:
> @leodemoura Your example
```
(@add_comm %%A %%inst %%a %%b)
```
> compiles but introduces a new universe parameter. Should we check for that?

@kha I'm not sure the triple backtick will work for universe polymorphic definitions.
In the current implementation, the elaborator uses a finalization step where unassigned universe metavariables are converted into universe parameters. 
https://github.com/leanprover/lean/blob/master/src/frontends/lean/elaborator.cpp#L3312

src/frontends/lean/parser.cpp
leodemoura: It looks fine to me. Could you please add a comment explaining the motivation for the hack?
I regularly forget the motivations for the hacks I have added. 
So, I'm trying to comment them before I forget them :)

Kha:
Yes, I just meant whether we should flag such quotations as errors eagerly.
Even if we had universe antiquotations, just using `to_expr` sounds much more convenient to me for universe polymorphic quotations. So I'm not very worried about the restriction.

leodemoura:
> Yes, I just meant whether we should flag such quotations as errors eagerly.

Yes, I think we should flag them as errors eagerly.

src/frontends/lean/parser.cpp
Kha: Done.

library/init/meta/simp_tactic.lean
```diff
@@ -372,7 +372,7 @@ meta def to_simp_lemmas : simp_lemmas → list name → tactic simp_lemmas
 | S (n::ns) := do S' ← S^.add_simp n, to_simp_lemmas S' ns
 
 meta def mk_simp_attr (attr_name : name) : command :=
-do t ← to_expr ``(caching_user_attribute simp_lemmas),
+do let t := ```(caching_user_attribute simp_lemmas),
    v ← to_expr ``({name     := %%(quote attr_name),
```
Kha, line 376: Not sure whether we'd want a `quote` equivalent for `expr`. We'd be back to the problem of polymorphic instances, but `name` would be fine.

leodemoura: I'm also not sure. Do you have a name for it? 
We would also be able to add instances for `nat`, `string`, `list name`, `list nat`, `list string`, etc.
It may look weird to have multiple instances of `list`.
Do you see a general solution for this?

leodemoura:
@kha I haven't merged this PR because I'm getting errors when compiling the stdlib.

Kha:
Yeah, the universes check turned out to be too restrictive - it's ok if the antiquotations' types contain universe mvars. Now I just check for universe params in the body of the substitution.

library/init/meta/simp_tactic.lean
Kha: @rlewis1988 I believe you expressed slight concern about the old `to_expr` functions going away, any opinions about bringing them back? It would be a typeclass function of type `α → expr`, so no `tactic`.

@leodemoura I can't think of a general solution. Even if we also passed the reflected type to the function, it would still have to do some partial elaboration to infer universe parameters. In that case, the user may as well use `to_expr ∘ quote`.

robertylewis: @Kha Yes, but only very slight. If they're there, there's at least one place I would use them. I don't have any strong feelings about it though.

PR #1415, opened 2017-03-04
fix(inductive_compiler/ginductive.cpp): debug code did not compile

dselsam:
Thanks @johoelzl for pointing this out. It compiled in Release mode because the offending lines are preprocessed out.


dselsam:
An assertion fires in `init`. I have a meeting, I'll fix this in an hour.

PR #1420, opened 2017-03-05
Default field values fixes

Kha:


src/frontends/lean/structure_cmd.cpp
```diff
@@ -621,8 +625,10 @@ struct structure_cmd_fn {
                 default_value = m_p.parse_expr();
             }
         }
-        if (default_value && !is_explicit(bi)) {
-            throw parser_error("invalid field, it is not explicit, but it has a default value", start_pos);
+        if (default_value) {
+            if (!is_explicit(bi))
+                throw parser_error("invalid field, it is not explicit, but it has a default value", start_pos);
+            default_value = some(mk_annotation(*g_default_val_body, *default_value));
```
Kha: @leodemoura So I guess fixing this issue via annotations doesn't work at trust level 0. I'm out of ideas, thoughts?

leodemoura: Yes, all macros are expanded at trust level 0.
We can fix this by wrapping the value with the identity function.
https://github.com/leodemoura/lean/commit/045fe4ad252a76a1939ce60e7fc8abbb2c55454b

Kha: Hah, nice. I've removed my commit.

PR #1428, opened 2017-03-07
[WIP] feat(frontends/lean/inductive_cmds.cpp): better resultant universe in…

dselsam:
…ference

- [x] Confirm the behavior of all examples in the test file.
- [x] Test a few more edge cases.

### DIFF
--- a/src/frontends/lean/inductive_cmds.cpp
+++ b/src/frontends/lean/inductive_cmds.cpp
@@ -11,6 +11,8 @@ Authors: Daniel Selsam, Leonardo de Moura
 #include "util/fresh_name.h"
 #include "util/sexpr/option_declarations.h"
 #include "kernel/replace_fn.h"
+#include "kernel/for_each_fn.h"
+#include "kernel/find_fn.h"
 #include "kernel/instantiate.h"
 #include "kernel/inductive/inductive.h"
 #include "kernel/abstract.h"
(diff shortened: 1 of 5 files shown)


leodemoura:
Is this still WIP?

dselsam:
I have finished my draft. I wouldn't mind going over the test/example file with you before merging though.

leodemoura:
> I have finished my draft. I wouldn't mind going over the test/example file with you before merging though.

I went over the test file this morning, and the output looks reasonable.
Do you have any concerns about it?
Anyway, I will merge this PR, we can fix problems later.

dselsam:
Can you confirm that this behavior is acceptable:
```Lean
inductive foo (α : Sort (u+1)) : Sort (u+1) 
| mk : α → foo 

inductive bar 
| mk   : foo bar → bar

check @bar -- bar : Type
```
It seems a little funny to me that `bar` is not universe polymorphic like `foo`, but I am fine with this.

leodemoura:
> It seems a little funny to me that bar is not universe polymorphic like foo, but I am fine with this.

I thought it is ok because `bar` does not have any parameters.
For example, `nat` is `Type` instead of `Type u`. 
I think this is good because if they were universe polymorphic, they would generate tons of unwanted universe parameters in definitions that use them. For example, suppose we have:
```lean
def foo : bar × bar × bar := 
...
````
This definition would have three universe parameters. The only way to avoid this is by providing universe parameters manually to each `bar`. Some users have defined a universe polymorphic `nat`, and they eventually got definitions with hundreds of universe parameters.

dselsam:
Great. I see you merged already so I'll close the PR.

PR #1435, opened 2017-03-08
RFC: Implement transfer method. Use it for int and dlist

johoelzl:
This adds a `transfer` method to lean.

It is used to prove that `int` is a commutative ring. For this it transfers along a relation between `int` (i.e. a disjoint sum of nonnegative and negative numbers) and pairs of natural numbers. Using `transfer` the instantiation of `comm_ring int` is much shorter and nicer structured. For example: proving `add_assoc` required a case distinction on all 3 variables, this is now completely gone. Now it requires a proof for each constant (i.e. `zero`, `add`, `mul`, ...) that it is proper wrt to the relation. This turns out to be a straight forward case distinction on the function arguments.

For `dlist` it transfers along the isomorphism between `dlist` and `list`.

This currently only supports _simple_ types. It does not yet support type constructors, i.e. it can not transfer `list (dlist α)` to `list (list α)` or `list int` to `list (nat * nat)`.

### DIFF
--- a/library/data/dlist.lean
+++ b/library/data/dlist.lean
@@ -85,4 +85,57 @@ by cases l; rsimp
 lemma to_list_concat (x : α) (l : dlist α) : to_list (concat x l) = to_list l ++ [x] :=
 by cases l; simp; rsimp
 
+section transfer
+
+protected def rel_dlist_list (d : dlist α) (l : list α) : Prop :=
+to_list d = l
+
+instance bi_total_rel_dlist_list : @relator.bi_total (dlist α) (list α) dlist.rel_dlist_list :=
+⟨take d, ⟨to_list d, rfl⟩, take l, ⟨of_list l, to_list_of_list l⟩⟩
+
+protected lemma rel_eq :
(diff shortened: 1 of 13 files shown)


library/data/dlist.lean
```diff
@@ -85,4 +85,57 @@ by cases l; rsimp
...
+protected meta def transfer : tactic unit := do
+  transfer.transfer [`relator.rel_forall_of_total, `dlist.rel_eq, `dlist.rel_empty,
+    `dlist.rel_singleton, `dlist.rel_append, `dlist.rel_cons, `dlist.rel_concat]
+
+example : ∀(a b c : dlist α), a ++ (b ++ c) = (a ++ b) ++ c :=
+begin
+  dlist.transfer,
+  intros,
+  simp
+end
```
leodemoura, line 130: Nice.

avigad:
Beautiful!

leodemoura:
I have already merged it. 
Sorry, I missed the `RFC` prefix.

johoelzl:
I marked it as RFC as its only tested on `int` and `dlist`. And as you already have them fixed, I didn't run the test suite.

Issue #1431, opened 2017-03-07
RFC: lenses


leodemoura:
We have been talking about adding a lenses library similar to the one available in Haskell.
Here is a very basic `lens` in Lean: https://github.com/leanprover/lean/blob/master/tmp/micro_lenses.lean
I'm not suggesting we use this basic definition, but I'm using this example to emphasize one issue: performance.
I want a "lenses library" that is "destructive update friendly".
The current VM performs destructive updates on arrays when their reference counter is one.
In the file `micro_lenses`, we have the following example:
```lean
vm_eval let p₀ := (mk_array 10 0, mk_array 5 tt),
            p₁ := (idx 1 ∙ fst)^.set p₀ 1,
            p₂ := (idx 2 ∙ snd)^.set p₁ ff in
        p₂
```
We have a pair of arrays, and all array updates are performed destructively because:
1- Lenses provide a `modify` operation which reads and updates a reference. If we had only `set` and `get`. We would have to define the `lenses.compose` `set` as:
```lean
λ a d, s^.set a (t^.set (s^.get a) d)
```
This definition would prevent destructive updates from being performed in the example above because `s` would still be alive when we execute `t^.set`.
2- We define `fst` and `snd` using `match`.
The following alternative definition would also prevent destructive updates
```lean
def fst {α β} : lens (α × β) α := 
{ modify := λ p f, (f p.1, p.2),
  ... }
```
since `p` would still be alive when we execute `f`.

Ok, we can use lenses to destructively update arrays nested in structures, but the code is not optimal since it reconstructs the pair instead of destructively updating `p`. This is not a big deal in this case, but suppose we have many layers of nested structures. We use the idiom above in the `hash_map` library: https://github.com/leanprover/lean/blob/master/library/data/hash_map.lean#L78
We can improve that by making sure the compiler generates better code for structure field lenses. First, they should be automatically generated. Second, the generated code should use a destructive update (when the reference counter is 1) instead of "reconstructing" the structure.
With these modification, the `hash_map` above will be almost as fast as an imperative one when the reference counter is 1.

What else do you want from a lenses library for Lean?

dselsam:
I'll just list a few simple features I make heavy use of in Haskell. The types are wild, but the basic functionality is simple.

1. Basics
```
Turn lens into projection:
view :: FoldLike b a a' b b' -> a -> b

Update inside a structure:
over :: ASetter a a' b b' -> (b -> b') -> a -> a'
```
[...]

PR #1438, opened 2017-03-08
feat(library/init/data/list): add array_list

digama0:
Add support for an array list type, using the "mutable" `array` implementation.

### DIFF
--- a/library/data/list/array_list.lean
+++ b/library/data/list/array_list.lean
@@ -1,0 +1,163 @@
+import init.data.quot
+universe u
+
+def defined_array.equiv (α) {n} (s : set (fin n)) (a b : array α n) := ∀ x ∈ s, a^.read x = b^.read x
+
+def defined_array.setoid (α) {n} (s : set (fin n)) : setoid (array α n) :=
+⟨defined_array.equiv α s,
+ λa x _, rfl,
+ λa b ab x h, eq.symm (ab x h),
+ λa b c ab bc x h, eq.trans (ab x h) (bc x h)⟩
+
+def defined_array (α) {n} (s : set (fin n)) := quotient (defined_array.setoid α s)
+
+def defined_array.lift_on {α n s} (q : @defined_array α n s) (β) (f : array α n → β)
+  (h : ∀ (a b : array α n), (∀ x ∈ s, a^.read x = b^.read x) → f a = f b) : β :=
+@quotient.lift _ β (defined_array.setoid α s) f h q
+
+def defined_array.read {α n s} (q : @defined_array α n s) (x) (h : x ∈ s) : α :=
+q^.lift_on (Π x ∈ s, α) (λa x _, a^.read x) (λ a b ab, funext $ λ x, funext $ λh, ab x h) x h
+
+def defined_array.write {α n s} (q : @defined_array α n s) (x) (h : x ∈ s) (v : α) : defined_array α s :=
+q^.lift_on _ (λa, @quotient.mk _ (defined_array.setoid α s) $ a^.write x v)
+  (λa b h, quot.sound (λi hi, show (if x = i then v else a^.read i) = (if x = i then v else b^.read i), by rw h i hi))
+
+def array_list.valid (size capacity) : set (fin capacity) := λ ⟨x, h⟩, x < size
+
+structure array_list (α : Type u) :=
+(capacity length : nat)
+(data : defined_array α (array_list.valid length capacity))
+(len_le : length ≤ capacity)
+
+def array_list.mk_data {α length capacity} (a : array α capacity) : defined_array α (array_list.valid length capacity) :=
+@quotient.mk _ (defined_array.setoid α (array_list.valid length capacity)) a
+
+local notation `⟦`:max a `⟧` := array_list.mk_data a
+
+def mk_array_list {α} [inhabited α] (capacity := 10) : array_list α :=
+{ capacity := capacity,
+  length := 0,
+  data := ⟦mk_array capacity (default α)⟧,
+  len_le := nat.zero_le capacity }
+
+def array.to_array_list {α n} (l : array α n) : array_list α :=
+{ capacity := n,
+  length := n,
+  data := ⟦l⟧,
+  len_le := le_refl n }
+
+def mk_array_list_of_fn {α n cap} [inhabited α] (f : fin n → α) (H : n ≤ cap) : array_list α :=
+{ capacity := cap,
+  length := n,
+  data := ⟦{data := λ⟨x, _⟩, if h : x < n then f ⟨x, h⟩ else default α}⟧,
+  len_le := H }
+
+def array.to_array_list_cap {α n cap} [inhabited α] (arr : array α n) (H : n ≤ cap) : array_list α :=
+mk_array_list_of_fn arr^.read H
+
+namespace array_list
+
+def nil {α} : array_list α :=
+{ capacity := 0,
+  length := 0,
+  data := ⟦array.nil⟧,
+  len_le := le_refl 0 }
+
+lemma lt_capacity {α} (l : array_list α) {n} (h : n < l^.length) : n < l^.capacity :=
+nat.lt_of_lt_of_le h l^.len_le
+
+def read {α} (l : array_list α) (x : fin l^.length) : α :=
+let ⟨n, h⟩ := x in l^.data^.read ⟨n, l^.lt_capacity h⟩ h
+
+def to_array {α} (l : array_list α) : array α l^.length :=
+{data := l^.read}
+
+def write {α} (l : array_list α) : fin l^.length → α → array_list α :=
+λ ⟨n, h⟩ v, { l with data := l^.data^.write ⟨n, l^.lt_capacity h⟩ h v }
+
+def trim {α} (l : array_list α) : array_list α :=
+l^.to_array^.to_array_list
+
+def truncate {α} (l : array_list α) (n) (nsz : n ≤ l^.length) : array_list α :=
+{ capacity := l^.capacity,
+  length := n,
+  data := l^.data^.lift_on _ (λa, ⟦a⟧) (λa b H, quot.sound (λ⟨x, h⟩ (hn : x < n), H ⟨x, h⟩ (lt_of_lt_of_le hn nsz))),
+  len_le := le_trans nsz l^.len_le }
+
+def ensure_capacity {α} [inhabited α] (l : array_list α) (cap : nat) : array_list α :=
+if H : l^.capacity ≥ cap then l else
+@mk_array_list_of_fn α l^.length (max cap (2 * l^.length)) _ l^.read $
+  le_trans l^.len_le $ le_trans (le_of_lt $ lt_of_not_ge H) (le_max_left _ _)
+
+theorem ensure_capacity_len {α} [inhabited α] (l : array_list α) (cap : nat) : (ensure_capacity l cap)^.length = l^.length :=
+if H : l^.capacity ≥ cap
+then begin delta ensure_capacity, rwa dif_pos H end
+else begin delta ensure_capacity, rw dif_neg H, reflexivity end
+
+theorem ensure_capacity_ge {α} [inhabited α] (l : array_list α) (cap : nat) : (ensure_capacity l cap)^.capacity ≥ cap :=
+if H : l^.capacity ≥ cap
+then begin delta ensure_capacity, rwa dif_pos H end
+else begin delta ensure_capacity, rw dif_neg H, exact le_max_left _ _ end
+
+def append {α} (l : array_list α) (v : α) : array_list α :=
+let L := @ensure_capacity _ ⟨v⟩ l (l^.length + 1) in
+have llen : L^.length = l^.length, from @ensure_capacity_len _ ⟨v⟩ l (l^.length + 1),
+have lc : l^.length < L^.capacity, from @array_list.ensure_capacity_ge _ ⟨v⟩ l (l^.length + 1),
+{ capacity := L^.capacity,
+  length := l^.length + 1,
+  data := L^.data^.lift_on _
+    (λa, ⟦a^.write ⟨l^.length, lc⟩ v⟧)
+    (λa b H, quot.sound $ λ ⟨y, yl⟩ hi,
+      let lf := fin.mk l^.length lc in
+      show (if lf = ⟨y, yl⟩ then v else a^.read ⟨y, yl⟩) = (if lf = ⟨y, yl⟩ then v else b^.read ⟨y, yl⟩), from
+      if h : lf = ⟨y, yl⟩ then by repeat {rw if_pos h} else
+      have y < L^.length, from eq.symm llen ▸ lt_of_le_of_ne (nat.le_of_succ_le_succ hi)
+        (λe : (fin.mk y yl)^.val = (fin.mk l^.length lc)^.val, h $ fin.eq_of_veq $ eq.symm e),
+      by repeat {rw if_neg h}; exact H ⟨y, yl⟩ this),
+  len_le := lc }
+
+def append_list {α} [inhabited α] : array_list α → list α → array_list α
+| l [] := l
+| l (x :: xs) := append_list (l^.append x) xs
+
+-- make sure this is compiled with overwrite optimization
+def insert_aux {α} : Π n (l : array_list α) (h : n ≤ l^.length), α → array_list α
+| 0 l h v := append l v
+| (n+1) l h v :=
+  let pos := fin.mk (l^.length - 1 - n) (array.lt_aux_3 h) in
+  let u := l^.read pos in insert_aux n (l^.write pos v) (le_of_lt h) u
+
+def insert {α} (l : array_list α) (v : α) (n : ℕ) (h : n ≤ l^.length) : array_list α :=
+insert_aux (l^.length - n) l (nat.sub_le _ _) v
+
+-- TODO (Mario): add operations for adding and removing several elements at a time
+def remove_aux {α} : Π n (l : array_list α) (h : n < l^.length), array_list α
+| 0 l h := l^.truncate (l^.length - 1) $ nat.sub_le _ _
+| (n+1) l h :=
+  let pos := l^.length - 1 - n in
+  let u := l^.read ⟨pos, array.lt_aux_3 $ le_of_lt h⟩ in
+  remove_aux n (l^.write ⟨pos - 1, by rw nat.sub_sub; exact array.lt_aux_3 h⟩ u) (le_of_lt h)
+
+def remove {α} (l : array_list α) (n : ℕ) (h : n < l^.length) : array_list α :=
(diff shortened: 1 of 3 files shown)


library/data/list/array_list.lean
```diff
@@ -0,0 +1,163 @@
...
+
+def read {α} (l : array_list α) (x : fin l^.length) : α :=
+let ⟨n, h⟩ := x in l^.data^.read ⟨n, l^.lt_capacity h⟩ h
+
+def to_array {α} (l : array_list α) : array α l^.length :=
+{data := l^.read}
+
+def write {α} (l : array_list α) : fin l^.length → α → array_list α :=
+λ ⟨n, h⟩ v, { l with data := l^.data^.write ⟨n, l^.lt_capacity h⟩ h v }
+
```
leodemoura: The VM will not be able to perform a destructive update on `l^.data` even if the reference counter to `l` is 1 and it owns `data`. This structure update is currently desugared into
```lean
array_list.mk l^.capacity l^.length (l^.data^.write ⟨n, l^.lt_capacity h⟩ h v) l^.len_le
```
In the current code generator we evaluate arguments from left to right.
So, after we execute `l^.data`, the array will have at least two references to it: one on the stack, and one from `l`. So, a destructive update is not performed.
To make sure we can perform a destructive update at run time, we need to write this function using a pattern matching
```lean
def write {α} : ∀ (l : array_list α), fin l^.length → α → array_list α 
| ⟨cap, len, data, hle⟩ ⟨n, h⟩ v :=
  ⟨cap, len, data^.write ⟨n, nat.lt_of_lt_of_le h hle⟩ h v, hle⟩ 
```
We can inspect whether we are performing destructive updates or not at run time by using the option `trace.array.updat`.
```lean
set_option trace.array.update true
vm_eval ((list.to_array_list [1, 2, 3])^.write (fin.of_nat 1) 10)^.read (fin.of_nat 1)
```
I have considered using match-with to compile structure updates, but this has nasty consequences.
For example, in the current approach we have the following definitional equality
```lean
{ l with data := ...}^.capacity   =?=   l^.capacity
```
That is, the projection of a non-updated field is definitionally to the same projection before the update.
This property is essential when we are building hierarchies of type classes (e.g., algebra).
If we use `match-with` to compile structure updates, then these two terms will be provably equal but not definitionally equal.
This is why I want to support lenses in Lean (#1431). With lenses, we can perform this kind of update in a much more convenient way and make sure a destructive update will be performed at run time if the reference counters are 1. Note that, in the lenses definition we will have to use a `match-with`. Otherwise we will have the same problem.

That being said, I realize that it would be great to have a mechanism to statically enforce that the reference counters are 1 at run time. To be able to do this, we would have to add linear types  to Lean.
We have considered this, but the complexity is too high and they are super inconvenient to use.

digama0: @leodemoura I tried your version and a few others:

    def write2 {α} : ∀ (l : array_list α), fin l^.length → α → array_list α 
    | ⟨cap, len, data, hle⟩ ⟨n, h⟩ v :=
      ⟨cap, len, data^.write ⟨n, nat.lt_of_lt_of_le h hle⟩ h v, hle⟩ 
    
    def write3 {α} (l : array_list α) : fin l^.length → α → array_list α
    | ⟨n, h⟩ v :=
      match l, h with
      ⟨cap, len, data, hle⟩, h := ⟨cap, len, data^.write ⟨n, nat.lt_of_lt_of_le h hle⟩ h v, hle⟩
      end

    def write4 {α} (l : array_list α) : fin l^.length → α → array_list α
    | ⟨n, h⟩ v :=
      let cap := l^.capacity, len := l^.length, data := l^.data, hle := l^.len_le in
      ⟨cap, len, data^.write ⟨n, nat.lt_of_lt_of_le h hle⟩ h v, hle⟩
    
    def write5 {α} (l : array_list α) : fin l^.length → α → array_list α
    | ⟨n, h⟩ v :=
      match l^.capacity, l^.length, h, l^.data, l^.len_le with
      cap, len, h, data, hle := ⟨cap, len, data^.write ⟨n, nat.lt_of_lt_of_le h hle⟩ h v, hle⟩
      end

    set_option trace.array.update true
    vm_eval ([1, 2, 3]^.to_array_list^.write (fin.of_nat 1) 10)^.read (fin.of_nat 1) -- non-destructive
    vm_eval ([1, 2, 3]^.to_array_list^.write2 (fin.of_nat 1) 10)^.read (fin.of_nat 1) -- destructive
    vm_eval ([1, 2, 3]^.to_array_list^.write3 (fin.of_nat 1) 10)^.read (fin.of_nat 1) -- destructive
    vm_eval ([1, 2, 3]^.to_array_list^.write4 (fin.of_nat 1) 10)^.read (fin.of_nat 1) -- non-destructive
    vm_eval ([1, 2, 3]^.to_array_list^.write5 (fin.of_nat 1) 10)^.read (fin.of_nat 1) -- destructive
    
    example {α} (l : array_list α) (n : ℕ) (h : n < l^.length) (v : α) : (l^.write ⟨n, h⟩ v)^.capacity = l^.capacity := rfl
    example {α} (l : array_list α) (n : ℕ) (h : n < l^.length) (v : α) : (l^.write2 ⟨n, h⟩ v)^.capacity = l^.capacity := rfl --failed
    example {α} (l : array_list α) (n : ℕ) (h : n < l^.length) (v : α) : (l^.write3 ⟨n, h⟩ v)^.capacity = l^.capacity := rfl --failed
    example {α} (l : array_list α) (n : ℕ) (h : n < l^.length) (v : α) : (l^.write4 ⟨n, h⟩ v)^.capacity = l^.capacity := rfl
    example {α} (l : array_list α) (n : ℕ) (h : n < l^.length) (v : α) : (l^.write5 ⟨n, h⟩ v)^.capacity = l^.capacity := rfl

Version 2 is the one you presented above, and version 3 is the suggested match expression. Both of them fail to reduce projections definitionally. I have a few alternatives, versions 4 and 5.

I assume that `let` expressions allow controlling the evaluation order, so I was hoping that version 4 would work. But it doesn't achieve the destructive write - do you know why? Perhaps it is just because `l` is still in scope, even though it is not being used. If this version can be made to work, it is a good alternative compilation for `{s with ...}`.

The only version which has both properties is version 5, which uses a match expression on the projections, but this involves an extra construction / destruction step.

leodemoura: > I was hoping that version 4 would work. But it doesn't achieve the destructive write - do you know why? 

The problem is that there are implicit references to `l` in the `data^.write` application (i.e., in the implicit arguments, and one of them is being captured by a closure that is just dead code).
They can be eliminated. I will try to fix this.

That being said, I also think `write4` is a great alternative for compiling `{s with ...}`. I will try to work on this too.

leodemoura: I pushed a fix that erases the artificial dependency that was preventing the destructive update.
`write4` is performing a destructive update after this fix.
I will implement the new approach for `{s with ...}`

leodemoura:
@digama0 I added the new encoding for `{s with ...}`, but I'm not sure whether we should keep it or not. It is not a complete solution for "enabling" destructive updates. That is, it is only guaranteed to work if the user is updating a single field. Example `{s with sz := s^.sz + 1, data := s^.data^.write i v}` may or may not update `s^.data` destructively. It depends on the evaluation order and where the fields `sz` and `data` are located in the structure. If we have two array updates, then for sure one of them will not be destructive `{s with data1 := s^.data1^.write i1 v1, data2 := s^.data2^.write i2 v2}`. There is also a performance problem. For example, suppose we have a structure `s` with many fields (e.g. `field`), then the new `{s with ...}` is much bigger than the previous one. The problem is that the `let` stores the type of each variable, and the types are not small in a structure like `field`. I tried to minimize the problem by creating a `let` only for fields which are not propositions. Another problem is when proving things about definitions that use `{s with ...}`. We will probably have to execute an initial `zeta` expansion step. To eliminate these auxiliary lets, but this can get complicated when we want to eliminate *only* the ones introduced by the `{s with ...}` notation.

digama0:
@leodemoura Do you have a specific example with this behavior? I have the following, but it's not showing what you say.

    structure test :=
    (data1 : array nat 3)
    (data2 : array nat 3)
    (sz: nat)
    
    def test.write (s : test) (i : fin 3) (v : nat) :=
    {s with data1 := s^.data1^.write i v, data2 := s^.data2^.write i v}
    
    set_option trace.array.update true
    vm_eval let a := {test . sz := 4, data1 := {data := λ_, 5}, data2 := {data := λ_, 1}}^.write
      (fin.of_nat 1) 10 in
      (a^.data1^.read (fin.of_nat 1), a^.data2^.read (fin.of_nat 2)) -- destructive write

leodemoura:
@digama0 Here is an example
```lean
structure test :=
(data1 : array nat 3)
(data2 : array nat 3)
(sz: nat)

def test.write (s : test) (i : fin 3) (v : nat) :=
{s with data1 := s^.data1^.write i v, data2 := s^.data2^.write i v}

def mk_test (n : nat) : test :=
{ data1 := mk_array 3 0,
  data2 := mk_array 3 1,
  sz    := n }

set_option trace.array.update true

#eval test.write (mk_test 10) (fin.of_nat 1) 10
/- 
Output:
[array.update] non-destructive write at #1
[array.update] destructive write at #1
-/
def test.write2 (s : test) (i : fin 3) (v : nat) : test :=
match s with
| ⟨d₁, d₂, s⟩ := ⟨d₁^.write i v, d₂^.write i v, s⟩
end

#eval test.write2 (mk_test 10) (fin.of_nat 1) 10
/-
Output:
[array.update] destructive write at #1
[array.update] destructive write at #1
-/
```

leodemoura:
In your example
```lean
vm_eval let a := {test . sz := 4, data1 := {data := λ_, 5}, data2 := {data := λ_, 1}}^.write
  (fin.of_nat 1) 10 in
  (a^.data1^.read (fin.of_nat 1), a^.data2^.read (fin.of_nat 2)) -- destructive write
```
The compiler simplifies `a^.data1` and `a^.data2` at compilation time because `a` is a constructor after inlining `test.write`.

leodemoura:
BTW, your example help me to find a potential performance problem in the compiler.
It is being too aggressive in the function inlining step, and may take a very long time in some cases, and/or consume a lot of time. It should have not unfolded/inlined `test.write` at compilation time.

digama0:
@leodemoura My recommended solution to your `{s with ... }` problem is to search for references to any let-bound projections in the body, and replace them with the variables. That is, in the example
```
structure test :=
(data1 : array nat 3)
(data2 : array nat 3)
(sz: nat)

def test.write (s : test) (i : fin 3) (v : nat) :=
{s with data1 := s^.data1^.write i v, data2 := s^.data2^.write i v}
```
we compile this to
```
let _sz := s^.sz in
let _data1 := s^.data1 in
let _data2 := s^.data2 in
{test . sz := _sz, data1 := _data1^.write i v, _data2 := data2^.write i v}
```
(where let bindings are only generated for data, as you suggested).

leodemoura:
>@leodemoura My recommended solution to your {s with ... } problem is to search for references to any let-bound projections in the body, and replace them with the variables. That is, in the example

At commit  c58f61e, the transformation was implemented at `elaborator.cpp`. Your suggestion cannot be reliably implemented there. The problem is that the term is full of "holes" at that point, and we can't collect all occurrences of `s^....` since some of these occurrences will be in holes that will be "filled" later. 
I think we can try to implement your suggestion in the code generator after we have erases proofs. Then,
1- All holes have already been filled and we can reliably collect all occurrences of `s^....`.
2- Proofs have been erased, and we don't have to be concerned about occurrences of `s^....` in proof terms.
3- Users will not be puzzled by alien `let _fname := ...` terms in their declarations .

digama0:
@leodemoura That actually makes a lot of sense. Evaluation order isn't even supposed to be a thing users need to worry about in dependent type theory, and the code generator is the one supplying the evaluation order on top of the original structure anyway.

What you really need is a proper analysis of the `arr^.write` terms in the expression relative to `arr^.read`, and ensure that the former are always evaluated first, unless a read depends on a write, in which case a non-destructive write is required. In general, it seems like linear types would definitely be useful, but I am not an expert.

> Just a quick comment. I didn't read the Haskell paper carefully, but I think their approach is not very useful for us (i.e., performing destructive updates) because given `f : A -o B`, they allow the caller to use `f` with non unique types. So, we cannot perform destructive updates on `A`.

In their implementation, there are two heaps, one for regular objects and one for linear objects, which must be routed through linear functions and have no GC. For our purposes, you can make sure that `A` is a linear object before entering the function, by copying it if the refcounter is > 1, that is, assuming that any function with a linear argument will modify its argument and so preempt the copying when the write in the function is found. That way, the code for `f` can always safely apply destructive updates.

leodemoura:
@jroesch and I needed a char buffer yesterday. We defined it as 
https://github.com/leanprover/lean/blob/master/library/data/buffer.lean#L8
We realized it is essentially `array_list` because the internal array implementation already has support for capacity/size.
Sorry for not addressing this PR earlier. I will not merge it since `buffer` is simpler.

digama0:
I am confused by your assertion. Exactly how is array being stored
internally, that would cause size and capacity to diverge?

leodemoura:
@digama0 
`parray` is used to implement Lean arrays.
https://github.com/leanprover/lean/blob/master/src/library/vm/vm_array.cpp#L14
The capacity in `parray` is stored before the elements.
https://github.com/leanprover/lean/blob/master/src/library/parray.h#L26
The size is stored in a cell:
https://github.com/leanprover/lean/blob/master/src/library/parray.h#L65

digama0:
@leodemoura Although it is nice to know that the array implementation has
these additional features, I would really like to have access to a raw
array type (without even array size stored), from which more complicated
verified programs can be built. Perhaps parray can also be exposed by a
different name, but for proper complexity control for lean programs, the
real underlying array structure should be exposed as much as possible, so
that more of it can be formalized and verified and fewer e.g. bounds checks
are needed. In particular, defined_array is meant to be an abstraction for
uninitialized or partially-initialized arrays, and it would be nice if the
VM could actually avoid array initialization for this structure.

leodemoura:
@digama0 Yes, it would be nice to have a primitive C-like array. We will need this kind of array to interface with external code. We can call them `c_array` or `carray`.

digama0:
@leodemoura Here is a slight variation on `defined_array` that doesn't require `inhabited α` for its operation. Would it be feasible to reflect this in C?

```lean
structure partial_carray (α : Sort u) (n) (s : fin n → Prop) :=
(data : Π (n : fin n), s n → α)

def carray (α : Sort u) (n) := partial_carray α n (λ_, true)
```

If I understand correctly, the parameter `s` will not be erased by the VM, but I can't think of any way to extract data from it (since equality on `fin n → Prop` is not decidable). Do you know if it is consistent to erase it? The idea is that `s` represents the subset of values of the array which have been initialized (which is not stored in the data structure). The management of this parameter is only used in order to ensure validity of the data access.

leodemoura:
> If I understand correctly, the parameter s will not be erased by the VM, but I can't think of any way to extract data from it (since equality on fin n → Prop is not decidable). Do you know if it is consistent to do so? 

An object of type `carray` has only one field: the function `Π (n : fin n), s n → α`.
The lambda lifting step may capture unnecessary dependencies, but we can refine it.
We cannot treat this function as an unary function because of partial applications.
The problem is easy to see with `safe_div : Pi (x y : nat), y > 0 -> nat`. 
An user may create the closure `safe_div 0 0`. This is fine, but he will not be able to build a proof for `0 > 0, and execute `safe_div` which would probably crash the program. However, if we erase the third argument, then we would be able to crash the program using `safe_div 0 0` (which is a fine term).

BTW, we erase irrelevant data (i.e., types, propositions (which are also types) and proofs) by replacing them with unit (i.e. `()`). We currently do not erase functions that produce irrelevant data.
The plan is to replace a term `t : Pi (a_1 : A_1) ... (a_n : A_n), B` where `B` is irrelevant with `mk_unit_closure_n` which is a pre generated byte code for `fun a_1 ... a_n, ()`.
This is needed in cases such as 
```lean
list.length (list.map (fun _, true) [1, 2, 3])
```
That is, the user creates a list of propositions using `list.map`. At runtime, we would be creating a list of units. By using this approach we make sure the code generation is uniform.

digama0:
@leodemoura 
> We currently do not erase functions that produce irrelevant data.
The plan is to replace a term `t : Pi (a_1 : A_1) ... (a_n : A_n), B` where `B` is irrelevant with `mk_unit_closure_n` which is a pre generated byte code for `fun a_1 ... a_n, ()`.

I think you can take this one step farther: Just replace `t` with `()` in the VM (i.e. `Pi (a : A), B` is irrelevant if `B` is). Then if apply is called on `()`, you ignore the argument and return `()` again.

> The lambda lifting step may capture unnecessary dependencies, but we can refine it.
We cannot treat this function as an unary function because of partial applications.
The problem is easy to see with `safe_div : Pi (x y : nat), y > 0 -> nat`.
An user may create the closure `safe_div 0 0`. This is fine, but he will not be able to build a proof for `0 > 0`, and execute `safe_div` which would probably crash the program. However, if we erase the third argument, then we would be able to crash the program using `safe_div 0 0` (which is a fine term).

Hm, that's too bad. This seems like an edge case, so it might be worth working around it. What is lambda lifting exactly? If I'm reading the code correctly, you are uncurrying sequences of lambdas. Does a function of type `A -> B -> C` always get compiled to a two arg function, or will it remain curried if the function is defined that way? If yes, you could eliminate all irrelevant args, and then compile a fully applied function application as normal, and a function application where only irrelevant args are unapplied to a thunk that calls the function. More generally, a member of type `Pi (a :: A), B` compiles to a function whose arguments are only the relevant arguments in `A`, unless there are none, in which case it compiles to a thunk.

I'm starting to see why a solid semantics for the VM data types is useful.

leodemoura:
> I think you can take this one step farther: Just replace t with () in the VM (i.e. Pi (a : A), B is irrelevant if B is). Then if apply is called on (), you ignore the argument and return () again.

For the bytecode interpreter, this would be reasonable. However, for native code generation, we don't want this kind of overhead.

leodemoura:
> What is lambda lifting exactly?

It is one of the preprocessing steps. For each nested lambda (except the ones used for `cases_on`), we create an auxiliary definition `foo a_1 ... a_n`. Where `a_1`, ... `a_n` are the local constants used in the lambda expression. At runtime, `foo a_1 ... a_n` becomes a tuple which contains a pointer to `foo` and `a_1` ... `a_n`. When we get the remaining arguments, we execute `foo`.

leodemoura:
We reduce the overhead of `safe_div` by creating an auxiliary definition: `safe_div.aux` which takes only 2 arguments. That is, the generated code is something like
```lean
def safe_div.aux x y :=
... -- code for the original safe_div is here

def safe_div x y h :=
safe_div.aux x y
```
Then, for fully applied `safe_div x y h` application, we inline and get
```
safe_div.aux x y
```
That is, we only pay for the overhead for partial applications.

PR #1439, opened 2017-03-08
fix(shell/server): remove unnecessary dependencies of info_task and use intermediate envs

Kha:
@gebner Looks good?

### DIFF
--- a/src/shell/server.cpp
+++ b/src/shell/server.cpp
@@ -320,11 +320,16 @@ server::cmd_res server::handle_sync(server::cmd_req const & req) {
 }
 
 std::shared_ptr<snapshot const> get_closest_snapshot(std::shared_ptr<module_info const> const & mod_info, pos_info p) {
-    auto snapshots = mod_info.get()->m_result.get().m_snapshots;
-    auto ret = snapshots.size() ? snapshots.front() : std::shared_ptr<snapshot>();
-    for (auto & snap : snapshots) {
-        if (snap->m_pos < p)
-            ret = snap;
+    auto ret = std::shared_ptr<snapshot const>();
+    if (auto res = mod_info->m_result.peek()) {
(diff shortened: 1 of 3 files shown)


src/shell/server.cpp
```diff
@@ -500,9 +491,10 @@ class server::info_task : public task<unit> {
                 auto opts = m_server->m_ios.get_options();
                 auto env = m_server->m_initial_env;
                 if (auto mod = m_mod_info->m_result.peek()) {
-                    if (mod->m_loaded_module->m_env)
-                        env = mod->m_loaded_module->m_env.get();
-                    if (!mod->m_snapshots.empty()) opts = mod->m_snapshots.back()->m_options;
+                    if (!mod->m_snapshots.empty()) {
```
Kha: I suppose there's no way for the snapshots to be empty here

gebner: Only if we disable snapshots completely.

Kha: Right, in that case we should fall back to the completed environment if available.

Kha:
I've always wondered why asynchronicity didn't break the interactive tests... makes sense if `info` has been basically synchronous up to now. We should probably add a flag `blocking` to `sync` and use it in interactive tests.

gebner:
> I've always wondered why asynchronicity didn't break the interactive tests... makes sense if info has been basically synchronous up to now. We should probably add a flag blocking to sync and use it in interactive tests.

Because we don't do anything asynchronously in the tests. :smile:   We use the `-j0` flag in the tests, which disables all multi-threading (and potential non-determinism).

Kha:
Huh, I guess I should actually look at the failed test before thinking about how to fix it :smile: . I now use `get_closest_snapshot` so that infos are printed in the correct local context (at least eventually, the `break_at_pos` parser may overtake the actual one).

PR #1440, opened 2017-03-08
Cyclic field defaults and overriding superclass defaults

Kha:
This change allows overriding a structure field's default value after its declaration, enabling both features mentioned in the title.

```lean
class Eq (α : Type) :=
(eq : α → α → Prop)
(ne : α → α → Prop := λ a b, ¬eq a b)
(eq := λ a b, ¬ne a b)

check ({eq := (=)} : Eq ℕ)
check ({ne := (≠)} : Eq ℕ)
```

```lean
class functor (f : Type u → Type v) : Type (max u+1 v) :=
(map : Π {α β : Type u}, (α → β) → f α → f β)

class applicative (f : Type u → Type v) extends functor f : Type (max u+1 v) :=
(pure : Π {α : Type u}, α → f α)
(seq  : Π {α β : Type u}, f (α → β) → f α → f β)
(map  := λ _ _ x y, seq (pure x) y)

check ({pure := @some, seq := @seq_app option _} : applicative option)
```

The extension to elaborating structure instances is quite naive for now; instead of computing a topological ordering of the default values, it repeatedly tries to insert applicable default values until it has defined all fields or gets stuck.

### DIFF
--- a/src/frontends/lean/elaborator.cpp
+++ b/src/frontends/lean/elaborator.cpp
@@ -2477,10 +2477,11 @@ expr elaborator::visit_structure_instance(expr const & e, optional<expr> const &
     get_intro_rule_names(m_env, S_name, c_names);
     lean_assert(c_names.size() == 1);
     expr c = copy_tag(e, mk_constant(c_names[0]));
-    expr c_type = m_env.get(c_names[0]).get_type();
-    unsigned i = 0;
     name_map<expr> field2value;
-    while (is_pi(c_type)) {
+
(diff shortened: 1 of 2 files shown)


leodemoura:
> The extension to elaborating structure instances is quite naive for now; instead of computing a topological ordering of the default values, it repeatedly tries to insert applicable default values until it has defined all fields or gets stuck.

I think this is fine. At least, it will be fine for a very long time. I do not expect we will have structures with more than 100 fields.

PR #1448, opened 2017-03-09
Default recover_from_error to false for most commands

Kha:


tests/lean/type_error_at_eval_expr.lean.expected.out
```diff
@@ -1,6 +1,6 @@
 type_error_at_eval_expr.lean:3:0: error: eval_expr failed due to type error
 nested exception message:
-type mismatch at definition '_eval_expr.16.423', has type
+type mismatch at definition '_eval_expr.16.436', has type
   list ℕ
 but is expected to have type
   ℕ
```
Kha: This test is quite flaky

Kha:
It can be argued whether commands such as `eval` should change the exit code, but I don't think any of these need error recovery per se.

tests/lean/type_error_at_eval_expr.lean.expected.out
Kha: @leodemoura Should we just replace integral `name` parts with `_` or something by default? Of course, ideally users should never see private names at all (by default).

leodemoura: I think this is a good idea. BTW, we should have an option for disabling this kind of cleanup. The option is mostly for ourselves when debugging crazy bugs.

gebner:
I think that for now this change is fine.  But in the long term, we want to have error recovery wherever possible, so that we can do type-directed autocompletion (e.g. of projector notations).

If I see it correctly, the motivating issue for this change was that we didn't fail on errors in run_cmd, right?  This will be fixed when we merge #1405.  Are there any other reasons to avoid error recovery?

PR #1451, opened 2017-03-10
feat(inductive_compiler): generate injectivity lemmas

dselsam:
- [x] Fix performance problem in proving basic injective lemmas.
- [x] Refactor so we use subst to prove the pack lemmas instead of the nested inj lemmas.


dselsam:
My laptop is too slow to run the entire test suite again right now to double check my last change didn't break anything. We can let travis run it.

PR #1456, opened 2017-03-13
chore(frontends/lean/inductive_cmds): allow sorrys

dselsam:


dselsam:
(I just noticed the old inductive-sorry test is failing. I'll push a fix later today)

PR #1461, opened 2017-03-14
Use attribute for registering attributes

Kha:
We could do the same for `vm_monitor.register`. Any others?


leodemoura:
> We could do the same for vm_monitor.register. 

This would be great!

> Any others?

None that I can think of.

avigad:
Nice! :+1:

PR #1464, opened 2017-03-16
Pretty-printing interactive tactics

Kha:
2017-03-16-182146_532x342_scrot

Implementing this was kind of terrible. I started out with a nice and general design where you can register arbitrary functions that may or may not format the type of an interactive response (info/complete) (third commit). But then I started questioning the usefulness of this design - the tactics pretty printer, in particular, should really only be used in interactive contexts. So now that pretty printer is hard-coded and directed by a thread-local indicating the context. And since I had already started hard-coding for tactics, I spent this afternoon on highlighting the current parameter:

2017-03-16-141052_933x109_scrot

It should also work, more or less, while editing a tactic script.


leodemoura:
> Implementing this was kind of terrible.

@kha it looks awesome. Are you not happy with the result?

src/frontends/lean/builtin_exprs.cpp
```diff
@@ -964,7 +964,7 @@ parse_table init_nud_table() {
     r = r.add({transition("`(", mk_ext_action(parse_lazy_quoted_pexpr))}, x0);
     r = r.add({transition("``(", mk_ext_action(parse_quoted_pexpr))}, x0);
     r = r.add({transition("```(", mk_ext_action(parse_quoted_expr))}, x0);
-    r = r.add({transition("`[", mk_ext_action(parse_auto_quote_tactic_block))}, x0);
+    r = r.add({transition("`[", mk_ext_action(parse_interactive_tactic_block))}, x0);
     r = r.add({transition("`", mk_ext_action(parse_quoted_name))}, x0);
```
leodemoura, line 968: @kha BTW, I recently noticed that 
```
`[ ... ]
```
blocks are resolving names dynamically, and this triggers nasty bugs. 
I think the names in this kind of block should be resolved at compilation time like you did for 
```
``( ... )
```
What do you think?

Kha:
> @kha it looks awesome. Are you not happy with the result?

Thanks! I'm not satisfied with some implementation details, but now that I've committed to hard-coding it for tactics, I'm happy with the outcome.

Kha:
> I think the names in this kind of block should be resolved at compilation time

Sounds reasonable. Should it be ``[...] for consistency then?

leodemoura:
> Sounds reasonable. Should it be ``[...] for consistency then?

I agree.

tests/lean/interactive/focus.lean
```diff
@@ -4,7 +4,7 @@ begin
   focus {  subst a
        --^ "command":"info"
          },
-     --^ "command":"info"
+       --^ "command":"info"
```
Kha, line 7: @leodemoura A slight regression re. goal info. For doc support while typing a tactic, we want the trailing whitespace to logically belong to the tactic, whereas previously it belonged to the end separator so that they share the same goal position. I don't think there's an easy way to reconcile these conflicting demands.

src/frontends/lean/tactic_notation.cpp
```diff
@@ -156,34 +160,49 @@ static expr parse_interactive_param(parser & p, expr const & ty, expr const & qu
...
+            p.check_break_before();
+            throw;
         }
-        type = binding_body(type);
-    }
-    while (p.curr_lbp() >= get_max_prec()) {
-        args.push_back(p.parse_expr(get_max_prec()));
+    } catch (break_at_pos_exception & e) {
+        e.m_token_info.m_tac_param_idx = args.size();
+        throw;
```
Kha, line 205: The `break_at_pos_exception` handling is slowly getting ridiculous, but when(/if?) we switch to embedding info in server-generated highlighting instead, I'll have to redo it anyway.

PR #1471, opened 2017-03-17
Run tests in parallel

Kha:
Try to run all run tests in the same Lean process first and skip individual tests if successful. This effectively cuts the run time of the complete test library in half.

When any of the Lean modules fail, we still go back to running the individual tests to minimize interference. Alternatively, we could teach Lean to write out individual flags reporting the success of each module.

Individual tests can still be run separately with `ctest -R/-E` without running `leanruntest_all`.


Kha:
> Alternatively, we could teach Lean to write out individual flags reporting the success of each module.

We could also do a similar thing for `tests/lean`.

leodemoura:
@kha It is very unlikely, but we may miss a failure due to interference. That is, the test succeeds because we reused the process. Should we ignore this kind of situation?

We may also try to delete obsolete tests (we never remove old tests, we just add them), and combine many small tests in a single one.

Kha:
> It is very unlikely, but we may miss a failure due to interference. That is, the test succeeds because we reused the process. Should we ignore this kind of situation?

It would sound like a miracle to me if that corruption (which we would fail to notice with or without this change) managed to do this in parallel to the other test in every single Travis configuration :) . So I'm not too concerned about it.

leodemoura:
> It would sound like a miracle to me if that corruption

@kha I'm not concerned about memory corruption, but the thread local caches we use. The cached information may be reused and hide the bug.

gebner:
I think this is a great idea.  Right now every test has a startup overhead of almost half a second, and our init directory is not going to get smaller.

> Alternatively, we could teach Lean to write out individual flags reporting the success of each module.

Is there a way to do this nicely with ctest?  The touch file as flag seems a bit hacky.

> We could also do a similar thing for tests/lean.

With #1405, it is easy to access the error messages for each module separately, and in a deterministic order.  So I think this should possible.

> @kha It is very unlikely, but we may miss a failure due to interference. That is, the test succeeds because we reused the process. Should we ignore this kind of situation?

I'd be more worried about the other direction actually, i.e. a test sometimes fails because it is run in parallel.  This might happen for example if we depend on the output of `mk_fresh_name`.

Kha:
Actually, it shouldn't be too hard to implement perfect separation: we load the standard library, then dispatch each argument file to a forked child process via a "process-level task queue". Edit: on Linux, that is.

leodemoura:
> Actually, it shouldn't be too hard to implement perfect separation: we load the standard library, then dispatch each argument file to a forked child process via a "process-level task queue". Edit: on Linux, that is.

Yes, this is a good approach for Linux.
Windows does not provide `fork`. Cygwin tries to simulate it, but it is very slow.

leodemoura:
@kha I think we can get a reasonable separation in Lean by using a fresh thread. We do exactly what you said, but we create a fresh thread instead of using `fork`.

Kha:
> @kha I think we can get a reasonable separation in Lean by using a fresh thread. We do exactly what you said, but we create a fresh thread instead of using fork.

Ok, we can do that if you're not concerned about interference from mutable global variables. For `mk_fresh_name`, we should implement hiding numeric name parts as we discussed. Any other globals that could pose a problem?

gebner:
> Ok, we can do that if you're not concerned about interference from mutable global variables. For mk_fresh_name, we should implement hiding numeric name parts as we discussed. Any other globals that could pose a problem?

I can think of one more kinds of "global state".  This should not happen with correct tests, but might occur during development.  Namely segfaults and other kinds of crashes.  If we run every test as a single process it is obvious which test is crashing.  With all tests in the same process, the responsible test becomes slightly more difficult to find.

Nevertheless I think that we should not try so hard to make the test setup different from the normal setup.  If a test fails depending on the state of mutable global variables, then it might also fail when run from emacs, or when run with `lean --make`.

Kha:
> With all tests in the same process, the responsible test becomes slightly more difficult to find.

In the setup of this PR, if the `all` test fails, we still run the individual tests. So I would suggest to do the following:

* upgrade this PR not to reuse threads
* possibly implement process isolation for Unix platforms in the future, which would then allow us to skip the individual tests even on failure

With both approaches, I believe we need some temporary files in the file system due to the limited nature of ctest.

gebner:
> upgrade this PR not to reuse threads

I don't think there is a way to do this at the moment.  If we spend time on writing a test runner in C++, I think we should rather write one that makes full use of the parallelization infrastructure that we have: then we could run all tests in a bit over 10 seconds (well, 30 seconds on my laptop).

Fundamentally speaking, I don't see the point of isolation here: if a test needs a perfectly isolated process to work, then it is broken from my point of view.  (And even if we decide to keep such flaky tests, we can always put them in a separate directory and execute them separately.)

> With both approaches, I believe we need some temporary files in the file system due to the limited nature of ctest.

An alternative would be to have one ctest test for all the lean tests, and have it show the failed tests in the output.

Kha:
> Fundamentally speaking, I don't see the point of isolation here: if a test needs a perfectly isolated process to work, then it is broken from my point of view.

I'm all for running tests in parallel in order to expose concurrency bugs, but I also understand Leo's concern that cache bugs may make broken tests _pass_, even if it's quite unlikely.

> An alternative would be to have one ctest test for all the lean tests, and have it show the failed tests in the output.

The downside is that you can't run the tests selectively any more (`-R`/`-E`/`--rerun-failed`).

leodemoura:
@kha Are you working on this PR? Should we close it, or merge as is?

Kha:
@leodemoura I've implemented a version that does not reuse threads on top of #1405, if you prefer that: https://github.com/Kha/lean/commit/b5e689fb77a9797e1b65f1190eaa58285ba26b38 (@gebner does that look right to you?)
The runtime overhead is just 2%.

gebner:
@kha I don't think right now is a good time to change `mt_task_queue`.  This causes a big merge conflict with #1405.

Practically, I think the biggest source of non-determinism when running tests in parallel is the use of `mk_fresh_name`.  There a quite a few tests whose output changes depending on the result of `mk_fresh_name`.  I don't think it matters for the run tests, but the thread reuse change does nothing to remove this non-determinism.

I don't think we've *ever* had a case where reusing a thread caused problems due to the changes to the thread local variables.  Most `scope_*` helpers restore the old value, after all.
Most problems with thread-local variables are due to difference between 1) calling a function somewhere inside the parser, vs. 2) calling it on a separate thread with fresh thread-locals.

Kha:
> This causes a big merge conflict with #1405.

Hopefully not, it's based on that PR :) .

> There a quite a few tests whose output changes depending on the result of mk_fresh_name.

Yes, we'll want to normalize that output when we parallelize `leantest`s (see above).

> I don't think we've ever had a case where reusing a thread caused problems due to the changes to the thread local variables.

I see. I'll be happy to implement this PR in whichever way you two see fit.

gebner:
> Hopefully not, it's based on that PR :) .

Oops, sorry, you even mentioned it in the text.  The change looks fine.

I still think it's the wrong approach.  But if you want to do it, it's fine with me.

leodemoura:
> I still think it's the wrong approach. But if you want to do it, it's fine with me.

@gebner Is the right approach the C++ test runner?

> I don't think we've ever had a case where reusing a thread caused problems due to the changes to the thread local variables

My main concern is silent failures. We had them in the past. 
I agree with you that `scope_*` helpers prevent the problems raised in this thread, but they are not used everywhere. 
For example, consider this recent commit: https://github.com/leanprover/lean/commit/7a58da11813b5ff8a3e02a030f39518c5174c7b0
Before this fix, the cache that was not tracking the environment, and would not reset like the ones wrapped with `scope_*` helpers.

> I think the biggest source of non-determinism when running tests in parallel is the use of mk_fresh_name.

I agree. @kha has a good solution for this problem: we erase the "numerical ids". We can add a pretty printer option that is only used by the test suite.

gebner:
> @gebner Is the right approach the C++ test runner?

My comment was directed at the thread reuse change.  While it does prevent some non-deterministic behavior, the biggest source (mk_fresh_name) is still there.  The reason why I think it is the wrong approach is because we only do it using tests.  The bugs caused by reusing threads are still there, we just don't get them when running tests.

I think the right approach would be to write a small test runner (or maybe just a flag to the main lean executable), that uses the same code as the command-line interface and the server.  That is, compile the lean files using the module_mgr, check the result and write the failing ones to stdout (or to files, whatever ctest likes better).  Basically very much like @Kha's original proposal, we would not even execute each test individually.
As a second step, we can extend this runner to the tests where we compare the output.  I think the easiest approach here is to just create an output file for each test that contains the messages.  A wrapper shell script can then call `diff` and fail if the results differ (as it does now).

I share your concerns that tests will behave differently when run in this way.  But at least they will fail or pass in the same way as if you used `lean --make` or the editor interface.

> My main concern is silent failures. We had them in the past.

Are you talking about the run tests, where some passed even though they produced errors?  This was a essentially a pure logic error, where we ignored some errors when determining the exit value of the process.  This error would have happened regardless of any parallelization.

To combat this issue, I have added some tests in `tests/lean/fail` that check whether we actually fail on those errors.  If we decide to implement a test runner, we should probably run those tests via the runner as well, to make sure that the runner does not ignore any failures.

> I agree. @kha has a good solution for this problem: we erase the "numerical ids". We can add a pretty printer option that is only used by the test suite.

I like this approach as well.  Should we maybe even hide the numerical ids by default?

johoelzl:
So is the idea still to use `fork()` on Linux? Even on Windows the fork implementation by Cygwin's (or is it now msys?) may be faster than Lean's startup time.

Kha:
> Basically very much like @Kha's original proposal, we would not even execute each test individually.

Then we are back to potentially hard to debug segfaults etc.
If we want to keep good ctest integration, I think the best we can do is to have a flag that makes Lean:
* always return with exit status 0 (on normal termination)
* write out the output and exit status of each input file. We then use these in the individual ctests if available, else run them individually.

gebner:
> Then we are back to potentially hard to debug segfaults etc.

We can still use a debugger.  We can also do the following to narrow down the segfault:
```
for i in *.lean; do echo $i; lean $i || break; done
```

Kha:
With some help from @gebner, I implemented the approach in my last comment so that leantests are also run in parallel and failing tests do not mean that all tests have to be re-run individually. Running the full test suite in release mode went from ~3min to ~30s for me. And it uncovered a thread locals bug, so no isolation may indeed be a good thing to test (too)?

src/shell/lean.cpp
```diff
@@ -567,27 +576,51 @@ int main(int argc, char ** argv) {
...
+            log_tree::node m_lt_node;
+        };
+        std::vector<input_mod> mods;
         for (auto & mod : module_args) {
+            auto lt_node = lt_root.mk_child(mod, "", {}, log_tree::DefaultLevel, true);
+            scope_log_tree lt(lt_node);
+            mod_mgr.set_log_tree(lt_node);
             auto mod_info = mod_mgr.get_module(mod);
-            mods.push_back({mod, mod_info});
+            mods.push_back({mod, mod_info, lt_node});
```
gebner: Did the `mod_info->m_lt` field not work out?

tests/lean/bad_unification_hint.lean.expected.out
```diff
@@ -1,4 +1,4 @@
-bad_unification_hint.lean:7:0: warning: declaration 'append_cons_hint' uses sorry
 bad_unification_hint.lean:7:0: error: invalid unification hint, failed to unify pattern after unifying constraints
-bad_unification_hint.lean:12:0: warning: declaration 'cons_append_hint'' uses sorry
+bad_unification_hint.lean:7:0: warning: declaration 'append_cons_hint' uses sorry
 bad_unification_hint.lean:12:0: error: invalid unification hint, failed to unify constraint #1
+bad_unification_hint.lean:12:0: warning: declaration 'cons_append_hint'' uses sorry
```
Kha, line 4: There is some reordering compared to `-j0`, but it doesn't look less right in any way, and should be deterministic.

tests/lean/caching_user_attribute.lean.expected.out
```diff
@@ -3,10 +3,12 @@
 eq.refl
 eq.mp
 
+[user_attributes_cache] no cached result for [foo]
+[user_attributes_cache] recomputing cache for [foo]
 eq.refl
 eq.mp
 
-[user_attributes_cache] cached result for [foo] has been found, but cache fingerprint does not match
+[user_attributes_cache] no cached result for [foo]
```
Kha: I didn't catch this one at first, will have to look into it.

src/shell/CMakeLists.txt
```diff
@@ -104,12 +104,26 @@ endif()
 
 # LEAN TESTS
 file(GLOB LEANTESTS "${LEAN_SOURCE_DIR}/../tests/lean/*.lean")
+add_test(NAME leantest_clear_all
+        WORKING_DIRECTORY "${LEAN_SOURCE_DIR}/../tests/lean"
+        COMMAND bash -c "rm *.test_suite.out *.status || true")
+set_tests_properties(leantest_clear_all PROPERTIES FIXTURES_SETUP AllFiles)
```
Kha, line 110: Note: `FIXTURES_SETUP` is cmake >= 3.7. In older cmakes,running a leantest/leanruntest t will show an outdated result in the unlikely sequence
* run `lean(run)test_all` but not t
* make a change that changes t's result
* run t but not `lean(run)test_all`

tests/lean/caching_user_attribute.lean.expected.out
gebner: This test is probably flaky now after the merge of #1489 (each of the `run_cmd` commands is executed as a separate task, and potentially on different threads).

tests/lean/bad_unification_hint.lean.expected.out
gebner: I was actually amazed how little the order changed.

src/shell/lean.cpp
Kha: You mean with a separate `log_tree` per file? That should work too.

tests/lean/io_bug2.lean.expected.out
```diff
@@ -1,3 +1,3 @@
+
 "t1"
 "t2"
```
Kha, line 3: Note: these are due to the `#eval` result (unit) being printed before trace messages.

src/shell/lean.cpp
gebner: No, I mean this [little field](https://github.com/leanprover/lean/blob/68c1ff62199a2db4a3471de19645c76ee163b648/src/library/module_mgr.h#L55), which contains the log_tree node for the given file.  You should just be able to write `mod_info->m_lt` instead of `lt_node`.

Kha: Ahh, that's much better. I thought you were referring to the same-named field in a previous revision.

leodemoura:
@kha what is the plan for fixing tests? I regularly use
```
test_single.sh ../../bin/lean tst.lean yes
```
`test_single.sh` uses `-j 0`.

Kha:
> test_single.sh uses -j 0.

Oh, right. It should instead use `--test-suite` too then.

gebner:
@Kha I just pushed a commit refactoring the fail tests to use the test runner.  This github feature is really handy.

tests/lean/fail/test_all.sh
```diff
@@ -0,0 +1,20 @@
...
+do
+    ff=$(../readlinkf.sh "$f")
+    if [[ "$OSTYPE" == "msys" ]]; then
+        # Windows running MSYS2
+        # Replace /c/ with c:, and / with \\
+        ff=$(echo $ff  | sed 's|^/\([a-z]\)/|\1:/|' | sed 's|/|\\\\|g')
+    fi
+    fs+=("$ff")
+done
+"$LEAN" --test-suite "${fs[@]}" || (rm *.test_suite.out *.status; false)
```
Kha, line 20: You can use the simpler `run/test_all.sh` since we don't care about filenames in outputs.

Kha:
@gebner Nice, thanks.

gebner:
Once this PR is merged, we can take a look at two other issues that have been blocked due to test performance:
 * Enable all tests in the debug builds on travis.  Right now the debug builds only run a limited amount of the tests (every 10th test between test 100 and 600...).  We ran into the travis timeout without this limit.
 * Tests in the emscripten version.  While we could always theoretically run the tests in the emscripten version, we might now actually be able to run them while drinking a single cup of coffee.  It would also be nice to have a travis build for emscripten, no idea how difficult this would be.

leodemoura:
I tried to merge this PR. `ctest` now freezes my machine. It is running for several minutes, and the VM is non responsive. I have tried my standard test command:
```
yes "A" | ctest -j 20
```
I've also tried
```
ctest
```
Both take much longer than before.

leodemoura:
I have more data. First, the performance problem only happens after I rebase this PR (`git rebase origin/master`). Can someone reproduce this problem?

Here are my numbers
1) current master branch and `time ctest -j 20` : 2 min 9 secs
2) this PR and `time ctest -j 20`: 1min 17 secs
3) this PR and `time ctest`: 2 min 32 secs
4) this PR + `rebase origin/master` and `time ctest -j 10` : 10min 32secs
5) this PR + `rebase origin/master` and `time ctest -j 20` : 12min 3secs

Remark: VM was not responsive on experiments 4, 5.

Kha:
Mmh, freezing sounds like swapping? It's not slower for me after rebase, but I can confirm that memory  consumption grows from 0.5GB to 2.5GB - #1489 I'd assume?

leodemoura:
> Mmh, freezing sounds like swapping? It's not slower for me after rebase, but I can confirm that memory consumption grows from 0.5GB to 2.5GB

Yes, it is probably swapping for me. Note that, the it gets slower as I increase `<num>` at `ctest -j <num>`.
Which `<num>` do you use?

Kha:
This was for -j4.

gebner:
I can not reproduce any slowdowns or large memory consumption.  This is after rebasing on master:
```
± time ctest -j20
[...]
100% tests passed, 0 tests failed out of 1054

Total Test time (real) =  32.50 sec
163.22user 4.73system 0:32.50elapsed 516%CPU (0avgtext+0avgdata 462552maxresident)k
1024inputs+53744outputs (10major+2463997minor)pagefaults 0swaps
```
I'm not sure how reliable `time`'s memory measurement is.  Sampling the output of `free -m` gives a peak memory usage of 1060M (all testing processes combined).

leodemoura:
> I'm not sure how reliable time's memory measurement is. Sampling the output of free -m gives a peak memory usage of 1060M (all testing processes combined).

@gebner I had to wait a long time for `ctest` to finish, the time reported by `time` seems to reflect the amount of time I had to wait. The machine was also non-responsive most of the time.

gebner:
@Kha I rebased the branch so that we can compare the same state of the code.

If we want, we can reduce memory usage further by setting the `RUN_SERIAL 1` property on the all tests.  This will prevent them from being executed in parallel with any other test (only needs 500M here).  We can alternatively set a `RESOURCE_LOCK USES_FULL_CPU` which would only prevent two all tests from being run at the same time (800M).

gebner:
@leodemoura Are you using a debug build?  Any non-standard options?

I have to wait for ctest at the end as well, but this is because the style_check tests is now the slowest test.

leodemoura:
> @leodemoura Are you using a debug build? Any non-standard options?

No, I'm using the same option I have always used. 
The numbers reported above were collected using the same build directory `build/release`.

leodemoura:
@gebner Recall that I had a similar freezing problem with the first the parallel build PR. 
We fixed the problem by using `-j0` in `test_single.sh`.

leodemoura:
The VM may be the source of the problem. I just found the following link:

http://stackoverflow.com/questions/29543339/adding-more-cores-to-virtualbox-makes-applications-slower-starting-at-16-cores

gebner:
Ok, I retried with a debug build and also on my underpowered laptop.  Both remain responsive and finish in a bit over two minutes.

> @gebner Recall that I had a similar freezing problem with the first the parallel build PR.

There the situation was different, we started a few *hundred* lean processes with multi-threading enabled.  Here we only start three of them.

But we can try a few things:
 * Do you have any tests that are not checked in? (Just making sure.)  Have you run cmake to update the ctest configuration?
 * What if we only run one of the all tests at a time? https://github.com/gebner/lean/tree/test_all_serial

gebner:
> There the situation was different, we started a few hundred lean processes with multi-threading enabled. 

@leodemoura I think this might have happened to you here.  If you did not run `cmake` after merging, then ctest would execute each test separately using the `test_single.sh` script *without* calling `test_all.sh` before.  If `test_all.sh` is not called before, then each `test_single.sh` script runs lean with multi-threading enabled.

I have now pushed a "fix" for this issue by passing -j0 in the test_single script (shouldn't be necessary with test_all, but we pass -j0 in the other test_single files as well).  We still use multi-threading in test_all.sh.

leodemoura:
@gebner I'm at home now. I will try your fix tomorrow.

leodemoura:
@gebner I tried your fix, without your fix, built from scratch, etc. Nothing works. In all cases, it takes between 10 and 12 mins to finish the tests. The machine is non responsive the whole time. 
I have `top` running on a separate terminal, and the problem doesn't seem to be memory consumption, nor the number of lean processes running. 

Then, I replaced
```lean
    num_threads = hardware_concurrency();
```
with
```lean
   num_threads = 4
```
and managed to finish the tests in 1min 30secs, and the machine was responsive the whole time.

leodemoura:
@gebner I'm trying to reconfigure the VM and make sure it uses only processor to avoid the issue described in the link I posted yesterday.

leodemoura:
Here is more data.
My machine has 2 processors with 8 (real) cores each. It reports 32 cores because of hyper threading.
If I allow the VM to use all 32 cores. Then, it takes 12 mins to run all tests. 
If I allow the VM to use only 16 cores. Then, it takes 4 mins to run all tests. I think in this setting the VM may still be using both processors and I may be experiencing the problem described in the link above.
If I allow the VM to use only 8 cores. Then, it takes 1 min to run all tests.
The runtime for `leanruntest_all` also looks reasonable now: 27.13 secs.
After this nasty VM experience, I will try to get a linux desktop for my office.

PR #1489, opened 2017-03-27
Refactor top-level parser

### LINKED ISSUE #1434: request: updating the 'parsing' message
When the parsing message appears in emacs or VS Code, it shows the line number at which Lean started parsing, and from this I can work out if snapshots are being used, and how many commands back it is rewinding before using a snapshot.

> Would it be possible for this parsing message to update at some interval, to tell me where it's currently up to?

I have by now many cases where files sometimes take minutes to compile, and it would be very helpful to quickly monitor where the slow compilation is. (I'm aware of the --profile flag on the command line, but it would be great to have some of this information readily available in the editor.)

gebner:
This PR removes all snapshot handling from the `parser` class and moves the top-level command parsing loop into a new `module_parser` class.  By default, every command is parsed in a separate task.  The snapshots are stored in a lazy async linked list.

Fixes #1434.

As discussed in #1405, this approach has a few advantages:
 * Better visual feedback about the parsing status.  You can see the parsing progress in emacs by following the fringe indicators.  For a nice illustration, open @semorrison's `lean-category-theory/universal.lean` file.
 * We stop parsing right below the edge of the screen (in the "visible lines" and "visible lines and above" modes).  Parsing is automatically resumed when scrolling down.  We do not need to reparse anything.
 * We can reuse snapshots even when we did not finish parsing the file.  Right now we only get the snapshots after we finish parsing the whole file.

At first I expected that executing every command in a separate task would be a massive performance hit, since it would presumably invalidate thread-local caches more often.  However, in practice this PR seems to reduce both runtime and memory usage (when compiling the standard library).  The old behavior of parsing the whole file in the same task can be re-enabled by calling `module_parser::use_separate_tasks`, if it turns out to be a problem.


kim-em:
I haven't narrowed this down to a reproducible case, but since this pull request was merged I've started seeing the Lean server pin at 100% CPU utilisation (2400%!), and stop ignoring edits.

PR #1496, opened 2017-03-29
[RFC] change precendence of field names to match application

gebner:
This was the second part of my projection syntax suggestion on slack. I'm posting this here mainly to have concrete data supporting the change. We can close this PR if we don't want to adopt the change.

I propose to change the precedence of the field-name token to match that of application.  At the moment we allow the following code:
```lean
(env s).get n
list.range x .succ -- parsed as (list.range (nat.succ x))
```
I find the second one particularly confusing as the space suggests that they are different arguments.  With this change, the code would look as follows:
```lean
env s .get n
list.range (x .succ) -- or: list.range x.succ
```

Practically speaking, this PR removes 158 parentheses in the standard library, often greatly increasing readability.  This is a particularly striking example:
```diff
 return $ new_hyps.map $ λnew_hyp,
-  (((clause.mk 0 0 new_hyp.2 c.local_false c.local_false).close_const h).inst
-               (op.1.close_const lci).proof).close_constn
-  (qf.2 ++ op.2.remove_nth i ++ new_hyp.1)
+  clause.mk 0 0 new_hyp.2 c.local_false c.local_false
+    .close_const h
+    .inst (op.1.close_const lci .proof)
+    .close_constn (qf.2 ++ op.2.remove_nth i ++ new_hyp.1)
```

Disadvantages:
 * `x .succ` is parsed differently than `x.succ`
 * It does not play well with the `.2` projections, since they are always parsed as tokens (and keep their old precedence).  This could be fixed by scanning `x.2.get` as a single name, and resolving the `.2` in the `id_to_expr` function.

### DIFF
--- a/library/data/bitvec.lean
+++ b/library/data/bitvec.lean
@@ -101,7 +101,7 @@ section arith
 
   protected def mul (x y : bitvec n) : bitvec n :=
   let f := λ r b, cond b (r + r + y) (r + r) in
-  (to_list x).foldl f 0
+  to_list x .foldl f 0
 
   instance : has_mul (bitvec n)  := ⟨bitvec.mul⟩
 end arith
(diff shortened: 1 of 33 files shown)


leodemoura:
@gebner I'm closing this PR based on our discussions at slack.

PR #1504, opened 2017-04-03
fix(.travis.yml): execute all tests in debug builds

gebner:
Let's see whether we still get timeouts on travis.


gebner:
Thanks to #1471, the tests now take about 3 minutes for the release build, and only 12 minutes for the debug build on travis.  The slowest build takes 36 minutes (the time limit is 50 minutes), this is where we build the standard library in trust-level zero (and in debug mode).

PR #1508, opened 2017-04-09
feat(library/system/io,shell/lean): add --run switch

### LINKED ISSUE #1289: Add an easy way to run a program with the VM from the CLI
It would be nice to have `--run` or something similar that looks for a known symbol (e.g `main`) and tries to `vm_eval` it. We could add a configuration flag and share this symbol for both the `--run` command and `--compile`.

gebner:
This PR adds the following:
 * A `--run` switch to the main lean executable, so that you can do `lean --run foo.lean --further -arguments to- for the main function--`.  The switch looks for a `main` definition of type `io α` or `tactic α`.
 * An `io.cmdline_args : io (list string)` function that returns the list of command-line arguments.

Fixes #1289.

Both the `#eval` command and the `--run` switch use the same `eval_helper` code.  This class takes a declaration in an environment, for example `main : ∀[io.interface], io unit`, adds arguments for the function arguments (well, just io.interface), and then executes the resulting function depending on its type either as a tactic or an io computation.  (Adding the `to_string` call still happens in `eval_cmd`.)

The command-line arguments are added as an extra field in the io interface, the arguments are specified when calling `mk_io_interface`.  Only the `--run` switch passes a non-empty list.  In particular, the command line arguments are (and should?) not be available to io code in tactics.

I did not add `io.exit` yet, since we can set the return value using `io.fail` for now.  Access to environment variables will come next, I'd just like to get comments on the general design first.

Example:
```lean
import system.io
variable [io.interface]

def main := io.cmdline_args >>= io.print_ln
```
```
$ lean --run main.lean -a 3
["-a", "3"]
```

Issues:
  * If you execute a tactic using `--run`, then `run_io io.cmdline_args` will return the empty list.

Alternatives:
  * Store command-line arguments in a global variable.
  * Add an extra command-line argument to main: `def main (args : io.cmdline_args) := ...`
  * Implement the argument injection and execution using an `unsafe_eval` type-class with instances for `∀ [io.interface], ...`, `tactic α`, `@io vm_io_iface α`, `result α`, `α` if `to_string α` is available, etc.


jroesch:
I think it looks good 👍, two questions though.

Should we organize this more like Haskell and create either an `io.env` or `system.env` namespace for placing functions like this? I would imagine our API will end up being pretty similar in the long run. We can then use a name like `io.env.args` or something else short. 

It also seems that the primitive list should be in IO, since  reading command line arguments should be contained inside of the `io` effect.

A final thought: we might want to ensure the underlying implementation is compatible with native code generation as much as possible to avoid special casing.

For example having a global variable that is setup during initialization is easy for the native compiler.

gebner:
> Should we organize this more like Haskell and create either an io.env or system.env namespace for placing functions like this? I would imagine our API will end up being pretty similar in the long run. We can then use a name like io.env.args or something else short.

I agree.  This seems like a good organization.

> It also seems that the primitive list should be in IO, since reading command line arguments should be contained inside of the io effect.

From my point of view, it doesn't cause any IO effect.  It is just a constant parameter that is passed down.  In particular, I don't think there is a sensible global value for the command-line arguments.  Calling `run_io io.env.args` is completely meaningless inside a tactic.

Do you know why rust decided to make `std::env::args()` a function instead of passing a list of strings to the `main()` function?  I couldn't find any discussion on that, as that decision seems to have happened before the RFC process.

> A final thought: we might want to ensure the underlying implementation is compatible with native code generation as much as possible to avoid special casing.

I don't think there will be compatibility problems with the native compiler, no matter which approach we choose.  With the current state of the compiler, you need to call `mk_io_interface` at some point anyhow, and then you can just add the command-line arguments.  For backends other than the C++ one, we'll need other io.interface implementations.

On the other hand, we will probably duplicate the logic in `eval_helper` and I don't see an easy way around it, as we need to perform different tasks: for the VM, we want to construct vm_objs as arguments; for the native compiler, we want to generate C++ code that constructs these objects.

emberian:
> Do you know why rust decided to make std::env::args() a function instead of passing a list of strings to the main() function? I couldn't find any discussion on that, as that decision seems to have happened before the RFC process.

I guess two reasons, but these are somewhat post-hoc, and don't need to be fundamental:

1. Keep function sig simple. it's always been `fn main() { ... }` or `pub fn main() { ... }`, since the first public version in 2010.
2. On Windows, getting the args is a CRT function.
3. On most Unix, the args aren't actually immutable. In fact, I've seen this abused in a C argument parser (it wasn't very nice). Rust pretends they are immutable though.

I don't think Lean needs to make the same choice here.

PR #1510, opened 2017-04-09
add new interactive tactics skip, ginduction, exacts

digama0:
This is used in some of my other work. The `exacts` tactic is discussed on the slack, although it has some issues with using variables in the local context of goals after the first, by comparison to a sequence `exact t1, exact t2, ...`.

The `ginduction` tactic performs induction on any expression with an inductive type. In the inductive step, you have access to the fact that the expression is equal to the current constructor. It is essentially reverting the goal `P t` to `Π x, x = t → P x` and performing induction on `x`.

`skip` is just `tactic.skip`, useful in tactic combinators in interactive mode.

### DIFF
--- a/library/init/meta/interactive.lean
+++ b/library/init/meta/interactive.lean
@@ -236,6 +236,13 @@ term of the goal. Let `T` be our goal, let `p` be a term of type `U` then
 meta def exact (q : parse texpr) : tactic unit :=
 do tgt : expr ← target,
    i_to_expr_strict ``(%%q : %%tgt) >>= tactic.exact
+/--
+Like `exact`, but takes a list of terms and checks that all goals
+are discharged after the tactic.
+-/
+meta def exacts : parse qexpr_list_or_texpr → tactic unit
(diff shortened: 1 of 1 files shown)


johoelzl:
There is `cases`, which handles also inductive cases and subsumes `ginduction`. It not only introduced the equalities, but also takes care of them later and simplifies them modulo constructor injectivity, i.e. `succ a = succ b` is replaced by `a = b` and when one of them is a local constant it is eliminated. This is for example the tactic which is used internally by the equations compiler.

gebner:
@johoelzl AFAICT `cases` neither adds an induction hypothesis, nor does it add an equation for the major premise.

johoelzl:
@gebner thanks for the clarification, I didn't think about this differences.

I think it would be nice to have a similar unification for the introduced equality as `cases` provides. Also I think it would be nice if the interactive version of `ginduction` would just replace `induction`. Then we do not need to explain to new users which version they should use.

PR #1514, opened 2017-04-10
Add a tactic which uses z3 to prove goals

jroesch:
This is an initial tactic which uses z3 to prove a subset of Lean goals, and uses sorry to finish the proof. 

Currently it supports a limited subset of Lean, it will translate uninterpreted functions, propositional logic & arithmetic into a z3 query.

Doing a clean-up pass right now, but we should merge the initial version. Since we are lacking arithmetic support in Lean currently this might be generally useful.

### DIFF
--- a/tests/lean/run/smt2_P_implies_Q_fail.lean
+++ b/tests/lean/run/smt2_P_implies_Q_fail.lean
@@ -1,0 +1,5 @@
+example (P Q : Prop) : P -> Q :=
+begin
+    intros,
+    z3
+end
--- a/tests/lean/run/smt2_p_implies_p.lean
+++ b/tests/lean/run/smt2_p_implies_p.lean
@@ -1,0 +1,7 @@
+import tools.smt2
+
+example (P : Prop) : P → P :=
+begin
+    intros,
+    z3 $ "/tmp/P.smt2"
+end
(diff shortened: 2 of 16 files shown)


tests/lean/run/smt2_P_implies_Q_fail.lean
```diff
@@ -0,0 +1,5 @@
+example (P Q : Prop) : P -> Q :=
+begin
+    intros,
+    z3
```
gebner, line 4: This introduces a hard build dependency on Z3 in order to run the tests.  I think we should move the Z3 tests into a separate directory and only run them if we can find the `z3` executable.

tests/lean/run/smt2_p_implies_p.lean
```diff
@@ -0,0 +1,7 @@
+import tools.smt2
+
+example (P : Prop) : P → P :=
+begin
+    intros,
+    z3 $ "/tmp/P.smt2"
```
gebner, line 6: Does this work on windows?

jroesch: This was leftover from testing, got distracted yesterday afternoon by other things and haven't had a chance to come back yet.

tests/lean/run/smt2_P_implies_Q_fail.lean
jroesch: I agree, hadn't figured out the right place to put them yet. Maybe we should just move this code out of the standard library into a package? I don't know how far along the package management tooling is.

leodemoura:
@jroesch Are you going to fix this PR? Or, should I close it?
It cannot be merged as is because of the problems raised by @gebner.

leodemoura:
This PR was merged into a separate repo: https://github.com/leanprover/smt2_interface

PR #1516, opened 2017-04-11
io.process: add exit status and working directory

gebner:
* The `system/process.lean` file is merged into `system/io.lean`.
 * `io.process.child` is now an abstract type.  You can still do `ch.stdin`.
 * `io.proc.spawn` no longer waits for the process to finish.  There is a separate call, `io.proc.wait`, that also returns the exit status.
 * I also fixed a bug where we closed stdout if the child process inherits it.

To do:
 * [ ] Add windows support for exit status and current working directory.


library/system/io.lean
```diff
@@ -68,6 +91,14 @@ interface.iterate e α a f
...
 iterate () $ λ _, a >> return (some ())
 
+def catch {e₁ e₂ α} (a : io_core e₁ α) (b : e₁ → io_core e₂ α) : io_core e₂ α :=
+interface.catch e₁ e₂ α a b
+
+instance : alternative io :=
+{ interface.monad _ with
+  orelse := λ _ a b, catch a (λ _, b),
+  failure := λ _, io.fail "failure" }
+
```
leodemoura, line 101: @gebner I'm not sure whether `io` should be an instance of `alternative` or not.
Right now, it is not an issue because we don't have laws for `alternative`, nor `laws` showing how `failure` interacts with `bind`. We will eventually need laws for `fail` and `catch` too.
This is messy in Haskell (@Kha posted a link to the Haskell documentation where they discuss the `alternative` laws). That being said, I think it would be nice if we had laws such as `a >> failure = failure` for the `alternative` type class, which are not true for `io`. 
Anyway, I'm merging this PR as-is. We can decide this issue later.

gebner: I agree.  `io` should not be a lawful instance of alternative, in particular because of the law you posted.  I only added the instance to have `<|>` as a lightweight error handling mechanism, and for this it would be enough if `io` just implemented `has_orelse`.

gebner: FWIW, I just noticed that Parsec-style parser combinators violate this law as well.

leodemoura: @gebner I think we should add a new type class that has the `catch` and `fail` actions (e.g., we can call it `exceptional` or something like that). For this type class `fail >> a = fail`, but `a >> fail` is not equal to `fail`. We can also add a nice notation for `catch`. Then `io` and `parsec` would implement it instead of `alternative`.  What do you think?

Kha: Note that the `alternative` laws are upheld for non-backtracking parsers, which I believe are better for error reporting, and also what the current `interactive` parser is doing (and what `tactic` should also be doing, imo). I used non-backtracking combinators for the JSON parser for POPL that I never finished: https://gist.github.com/Kha/9a9a19ee1dd3331829f182555dddaf0b#file-parser_combinator-lean-L53

Edit: Now I'm not sure any more whether non-backtracking changes any laws. I still like it better in practice :) .

gebner: > Note that the alternative laws are upheld for non-backtracking parsers,

AFAICT they are only upheld by the *backtracking* combinators (as in attoparsec and your parser combinators, and as opposed to parsec).

In the non-backtracking combinators you need to keep track of whether input has been consumed (that is, if we should backtrack or not).  Hence you necessarily have multiple kinds of failure that you can distinguish.  In particular `a >> fail` can be different from `fail`, since you can observe whether `a` consumes input or not.

Generalizing this point, I think that all parser combinators *should violate* the `a >> fail = a` property.  This is because we want good error messages:
```lean
run (str "foo" >> fail) "foobar"
-- error at foo[]bar
run fail "foobar"
-- error at []foobar
```
Hence `str "foo" >> fail ≠ fail`.

gebner: > I think we should add a new type class [`exceptional`] that has the catch and fail actions

If we do that, then `alternative` should extend `exceptional`.  For me the main value of the alternative type-class is that we can avoid duplicate definitions, e.g. `optional`, or other abstractions that work in all of parser/parser/tactic/io/solver/smt_tactic/etc.

Having said that, at the moment alternative extends applicative and not monad, I think we need to change that since we can't get the laws for bind from the laws for seq.  AFAICT these are the possible laws:
 1. `<|>` is associative
 2. `(failure <|> a) = a`
 3. `(a <|> failure) = a`
 4. `(a >> failure) = failure`
 5. `(failure >>= a) = failure`
 6. `>>` is left-distributive over `<|>`
 7. `>>=` is right-distributive over `<|>`
 8. `(return x <|> a) = return x`

The monads `list` and `set` satisfy all of these laws.   The `option` and `result` monads satisfies all except 7.

On the other hand, `io`, `parser`, `tactic` should still (or could be made to) satisfy all laws except 4, 6, and 7.

The restriction to monad doesn't seem so bad.  We could have the following:
```lean
class exceptional (f : Type u → Type v) extends monad f :=
(failure : Π {α : Type u}, f α)
(orelse  : Π {α : Type u}, f α → f α → f α)
-- laws 1-3, 5, 8

class alternative (f : Type u → Type v) extends exceptional f :=
-- laws 4,6,7
```

I'm not sure if we'd ever use the additional laws provided by alternative, or what we'd gain at all from adding the laws.

leodemoura: > Having said that, at the moment alternative extends applicative and not monad, I think we need to change that since we can't get the laws for bind from the laws for seq. 

Yes, `alternative` should inherit from `monad`, or we should add an `alternative_monad` type class.

leodemoura: @gebner @kha I'm not sure if this is high priority, but these laws are essential for anybody trying to prove things about code using these type classes. 

BTW, I also tried a quick experiment with the state monad this week. The state monad as-is is quite bad for verification because there is no easy way to state properties about the state encapsulated in the monad. I defined a monad inspired by HTT (Hoare Type Theory). Here is an example: https://gist.github.com/leodemoura/15e4f19f325139b17575522bce80dd80
This approach will only be convenient to use after we add tactics for automating lemmas such as `init_mem_inv`.

PR #1519, opened 2017-04-13
fix(library/type_context): properly undo tmp mvars assignments

Kha:


leodemoura:
@kha In principle, we should not override assigned meta variables. Before merging this PR, I want to understand in which circumstances this is happening, and whether it is a bug or not. My concern is that the "override" may be a bug. Could you send me an example that triggers the assignment override?

leodemoura:
@kha The override can happen when compressing the assignment. Example: `subst[?x] = t`, then we instantiate metavars in `t` and want to update the assignment. This "compression" is an optimization, and it is not clear whether it is useful in tmp mode. So, if this is the only case where we can override the assignment, I suggest we disable "compression" in tmp mode.

Kha:
Would [this](https://github.com/leanprover/lean/blob/8ea2bc08cb03b49f61db90c28115dfa9f4d09410/src/library/type_context.cpp#L1834) be the location of the compression? I tried to skip the assignment when `!in_tmp_mode()`, but it didn't seem to fix my [n-queens example](https://gist.github.com/Kha/2cdd2df8bd318019bea75f1ea87ae4a0).

leodemoura:
It is performed in the following two places:
https://github.com/leanprover/lean/blob/8ea2bc08cb03b49f61db90c28115dfa9f4d09410/src/library/metavar_util.h#L89
https://github.com/leanprover/lean/blob/8ea2bc08cb03b49f61db90c28115dfa9f4d09410/src/library/metavar_util.h#L137

Kha:
Indeed, the following commit also fixes the problem: https://github.com/Kha/lean/commit/36daf0c8660196e94f53a4de761713c2acdfbc2f
I haven't run it through the test suite yet.

leodemoura:
We will merge #1526

PR #1520, opened 2017-04-13
[WIP,RFC,preview] leanpkg

### LINKED ISSUE #1200: RFC: Projects and dependencies
Most modern programming languages have designated build tools that allow the easy inclusion of third-party code. (I'm thinking of haskell/cabal, rust/cargo, js/npm, coq/opam, ...)
Lean doesn't (to the best of my knowledge), and this significantly limits its extensibility (e.g. you cannot use the resolution prover from the `library_dev` repo anywhere else..).

Practically speaking, we only have two usable ways to reuse code right now: 1) put it into the standard library, or 2) copy&paste it into the same directory.  (This is because we set the LEAN_PATH to the current directory plus the standard library.)

I think we should have an easy and standard way to include external dependencies.  Has anyone worked on this?  I kind of expect this question to come up at POPL as well, so it would be good to have at least a rough plan on what we want to do.

gebner:
Five second intro:
```
git clone https://github.com/gebner/demopkg
cd demopkg
leanpkg configure
```
Or just open `demopkg/test.lean` in emacs and follow the on-screen instructions. :slightly_smiling_face:

I'm looking forward to any and all kinds of comments of feedback, as well as questions if anything is unclear.  Now is the time to speak up, before we're stuck with the design!

The design almost literally matches the proposal in #1200.  Here is the usage message from `leanpkg`:
```
[gebner:~] $ leanpkg

Usage: leanpkg <command>

configure       download dependencies
build           download dependencies and build *.olean files

new <dir>       creates a lean package in the specified directory
init <name>     adds a leanpkg.toml file to the current directory, and sets up .gitignore

add --git <name> <url> <rev>
                adds a dependency from a git repository
add --local <name> <directory>
                adds a local dependency

install --git <name> <url> <rev>
                installs a user-wide package from git
install --local <name> <url>
                installs a user-wide package from a local directory

dump            prints the parsed leanpkg.toml file (for debugging)
```

Misc points:
 * Interestingly, `data.dlist` can be used as a lazy list, with a lazy append operation.
 * Abusing the equation compiler as a command-line parser doesn't work that well.  The `leanpkg.main` function takes three seconds to elaborate on my laptop.

Fixes #1200.

### To do

 * Documentation
 * Add `leanpkg.toml` files to existing packages.
 * Decide on backwards compatibility (existing packages need a `leanpkg.toml` file added)
 * Windows support (I would really welcome help here.)
   * Exit status for processes
   * Set working directory for child processes
   * Functions that resolve an relative or absolute path against a given base path
   * Getting the home directory
 * `io.path`: is a path an existing directory/file?, create directories (recursively), resolve path against base
 * Move the `result` type out of the native compiler, add enough type-class instances, and use it for error reporting instead of `option`.
 * Somewhere in `io`: get and set working directory for current process (if we want to have these functions, e.g. java doesn't)
 * vscode integration (it works, but you need to run `leanpkg configure` yourself)
 * Check whether the `leanpkg` script works in the release tarballs.
 * Check whether it works on macOS.
 * Improve startup performance by using the native compiler.
 * Tests (at least add a cmake target to compile it)


jroesch:
I'll build it and try on MacOS, as well as try to compile it with the native compiler (I just rebased my months old patch, and it is working for the test cases). 

As for other todos, I think we should follow the convention I proposed last week and move functions for getting/setting the current working directory to be under `io.env` (or whatever we are going to call it).

I will follow up with another issue but I think we should move things from the standard library into Lean packages, for example `super`, `smt2`, and complex math developments.

leodemoura:
> Abusing the equation compiler as a command-line parser doesn't work that well. The leanpkg.main function takes three seconds to elaborate on my laptop.

Do you know where is the bottleneck?

leodemoura:
@gebner I'm wondering if you have a plan for testing our own packages. I'm considering the following scenario:
- We have multiple packages at leanprover (e.g., `super`, `z3`, `mini-crush`, etc).
- We update leanprover/lean, and we want to find out which packages are affected by the change.

avigad:
This looks really nice. Do I understand correctly that with the current design, there is one package per github repo? For example, if we want to maintain a dozen "standard" packages at leanprover, we will keep them in a dozen separate repos, rather than in one repo?

gebner:
> Do you know where is the bottleneck?

I did not look closely yet, but a lot of time is spent generating lemmas for the main function.  It's much better with `set_option eqn_compiler.lemmas false`.  The main functions gets 26 lemmas, such as the following:
```lean
leanpkg.main._main.equations._eqn_22 :
∀ [_inst_1 : io.interface] (a : string) (a_1 : list string),
  ¬a = "--local" → ¬a = "--git" → main._main ("add" :: a :: a_1) = io.fail usage
```
I have disabled them for now since I don't need them.

> I'm wondering if you have a plan for testing our own packages.

The "industry-standard" solution is something like [cargobomb](https://github.com/brson/cargobomb), [hackager](https://github.com/dterei/hackager), or [coq-bench](http://coq-bench.github.io/), i.e. build all known packages against a nightly version of the compiler.  I believe Isabelle's AFP is used in a similar way, although I couldn't confirm it with a quick google search.

In our case, I would start by adding a separate repository that contains a bash script that builds the latest lean master and tries to build the git versions of all known packages.  We can make travis execute this daily/weekly (or manually if we want the results for a given PR).

gebner:
> Do I understand correctly that with the current design, there is one package per github repo? For example, if we want to maintain a dozen "standard" packages at leanprover, we will keep them in a dozen separate repos, rather than in one repo?

This is (mostly) how it's done in rust/javascript/haskell/ruby/vim/etc., and it seems to work well.  I think this structure will work very well for tooling (e.g. the `z3` and `super` packages that everybody talks about) and specialized developments (cumulative hierarchy, etc.).  From what I see, Isabelle's AFP operates on a similar granularity.

I don't have enough experience to know the right granularity for a general-purpose analysis/linear-algebra/measure-theory/topology library.  Either we split them into many small packages, or make one large package.  Having separate `analysis`, `linear_algebra`, etc. packages might not work because of cyclic dependencies between them.  For now I'd suggest to make one large package (say `math`), this would still decouple core development from theory development without adding too much complications.  We can still decide whether and how to split later on.

From a git repo organization point of view, it would be easy enough enough to support multiple packages per git repo, (for now) you could depend on them as follows (proposed syntax, not yet implemented):
```toml
basis_theorem = { git = "https://github.com/leanprover/math", path = "basis_theorem", rev = "6a0ba5870db86ada6d53934cb30707e2a2352725" }
```

avigad:
That sounds good. I don't know whether any finer granularity is needed, and it shouldn't be a problem to bundle related libraries and tools together, when that seems to be the right thing for someone to manage and maintain. We'll figure it out. I am excited about this! I think it will work really well.

johoelzl:
Regarding the AFP: Lars Hupel installed a Jenkins instance (https://ci.isabelle.systems/status/) which runs nightly builds of the Isabelle repository and the AFP. In addition to this there is a testboard for Isabelle which is run against the current AFP and vice versa. There is no solution to say that a test version of Isabelle should be tested against a specific version of the AFP.

spl:
Just as a bit of added evidence, it's not uncommon to have multiple related Haskell packages in a single repository. One example that comes to mind is [Yesod](https://github.com/yesodweb/yesod).

Kha:
Any plans on how `leanpkg upgrade` should behave? Should we make sure to include a mandatory `branch` field that we can use to implement updates in the future, or should we just fall back to `master` in its absence?

Many packages will probably have no need for multiple branches, but there may be some (e.g. `library_dev`) that want to track the git version of Lean on a separate (non-default) branch. Or they use tags for releases for stable Lean, in which case we somehow want updates to be tag-based...?

Kha:
And of course the most dreaded question: What should we do with dependency diamonds? I don't think we can really do anything other than reject them on a version mismatch. That is certainly one disadvantage compared to a centrally managed system like the AFP.

gebner:
> Many packages will probably have no need for multiple branches

This is a good question.  My first instinct was that most people would continue targeting Lean master for the time.  That is, the `master` branch of the packages would build with Lean `master`.  Released versions (tags) would work with the Lean release at that point.

But maybe it is better for packages to target the latest Lean release.  Then we can break Lean master as much as we want.  We would just have to speed up the release cycle a bit (to get the bugfixes out of the door), and maybe provide better tooling in the editors to switch between Lean versions.

> Any plans on how leanpkg upgrade should behave?

For now I think the only reasonable action is to bump the `rev` attribute to the rev that `master` refers to.  (Or maybe the latest tag, but I don't think that's well-defined.)  If we ever make a registry, it should IMHO bump to the latest released version.

> And of course the most dreaded question: What should we do with dependency diamonds?

The current state is that the first version that leanpkg encounters "wins".  You can "override" any transitive dependency by just declaring it again in the current package.  I believe this should solve most of the diamond problems without too much complexity.

From a policy point of view, packages should probably always use the latest released version of their dependencies.  This is what effectively happens in package ecosystems anyhow: for example in nixpkgs, we build all Hackage packages and only have the latest version of everything.  Pretty much all of the breakage is due to overly restrictive version bounds, it is rare that a package actually doesn't build with the latest version of its dependencies.

jroesch:
@avigad @gebner re: multiple packages its possible with Cargo and Cabal to have a single repo containing many packages see https://github.com/nikomatsakis/lalrpop for example. 

I think some of the versioning upgrade issues can be resolved if/when we have a package registry. 

We could use Github as a stop gap in the style of Atom or Homebrew, designing a registry + website + server is a fair amount of work, and requires hosting.

jroesch:
I also started factoring out result to make it generally useful, just a heads up.

I moved `result.lean` to `system/except.lean` it looks like the `result` constructor from `interactive_monad` is leaking into the namespace, which I seem to remember a the reason we ended up with`native.result`. 

This brings the name in line with the monad transformer which I was calling `except_t` which is more appropriate (to me) then `result_t` given it adds exceptional behavior to a monad.

I did this work on the native compiler branch, since doing it on master was likely to make rebasing even more painful. 

The compiler is currently behind an alpha flag, so I would like to merge it into master soon just to prevent bit rot in the C++ portions, and remove `native` from `init` its been in there way too long, even though it still has a few missing features. 

Given the path overlap it seems that the easiest order is for you to merge `leanpkg`, then I'll resolve the path behavior on my branch. 

Hopefully I can then compile `leanpkg` with the native compiler which I think would be cool to try ASAP. Let me know if any of this sounds wrong to you 👍 

Also are there other type classes we want besides `monad` and friends? 

Maybe `alternative`?

jroesch:
We should ensure that the build infrastructure invokes `lean --make leanpkg` it appears that just running `ninja` didn't generate `.olean` files for me. I noticed this because it was taking between `1.75-3s` to run leanpkg and display the help message. If we run `--make` the time reduces to roughly 1s, which is slightly above the time it takes to load `init` and print `Hello World`. 

cc @leodemoura 

No pressure but should we just merge this? it appears to work and would be easier to start experimenting with if merged into master.

gebner:
@jroesch I haven't had much time to work on this PR lately.  The main blocker for merging is windows support, the other todos are either 1) ugly and slow but works, or 2) needs to be polished before the next release.

gebner:
So, I spent a bit of time on leanpkg today:
 * Windows support is no longer broken.  At least `leanpkg configure` works.
 * Nicer syntax for `leanpkg add`.
 * `[package] path = "src"` for @jroesch 

The following now works out of the box, if you want to install the SMT2 interface user-wide to play around:
```
leanpkg install https://github.com/leanprover/smt2_interface
```

Remaining todos (can be done after merging):
 * `leanpkg upgrade` (as requested in this thread)
 * Windows support:
    * spawning `lean --make` does not work
    * stdout from child processes is invisible
    * emacs can't spawn `leanpkg configure`
 * vscode integration for `leanpkg configure`
 * `io.path`, `io.env` additions (waiting for @jroesch's new path class to be merged)
 * Before the next release:
   * Add documentation to tutorials/website/etc.
   * Verify that leanpkg works in release tarball.

cipher1024:
I understand that something like Hackage is planned for collecting packages. In the mean time, would it make sense to have a simple page where people can post a link to their git repository that we can copy-paste in leanpkg?

avigad:
We'll get there, but it is too early now. The system is still changing and as far as I can tell most user developments are experimental. If you have a package that you think others might want to use, please advertise it on lean-user or gitter. When there are enough of them out there, it will be easy to set up a list of user-maintained repositories.

PR #1526, opened 2017-04-16
fix(library/metavar_util): do not compress mvar assignments in tmp mode

Kha:
Alternative fix for #1519


leodemoura:
@kha I'll use this one because it is easier to integrate with the new match procedure

PR #1529, opened 2017-04-21
Add C++ functions required by smt2 bindings and remove partial bindings.

jroesch:
This is the final cleanup from the z3 bindings I mentioned earlier today @leodemoura.


library/init/meta/format.lean
```diff
@@ -23,6 +23,7 @@ meta constant format.of_string       : string → format
 meta constant format.of_nat          : nat → format
 meta constant format.flatten         : format → format
 meta constant format.to_string       : format → options → string
+meta constant format.to_buffer       : format → options → string
```
gebner: This type does not look right.

jroesch: It does not. I must of done something weird when I carved up my old commits. Good catch.

PR #1532, opened 2017-04-21
Various --profile fixes

gebner:


gebner:
Weird, this builds for me locally.  I'll take a look at it tomorrow.

gebner:
Works now.

gebner:
You can now enable the profiler (and set the profiling threshold) via `set_option` as well:
```lean
set_option profiler true
set_option profiler.threshold 0.01 -- 10ms, defaults to 0

-- profiling is enabled here
```

PR #1536, opened 2017-04-24
[WIP, RFC, Preview] Linting Framework

jroesch:
This a prototype for enabling user scriptable linting/warnings, there are a couple issues to work out.

What is the right way to query position information, given the decl hasn't been added to the environment yet, it seems we should use `save_info_thunk` for attaching warnings.

I'm also not sure what the success/failure semantics should be, it seems that in a `-Werror` mode we would want any generated diagnostics to cause an error in Lean.

Below is an example of using the attribute:

```lean
set_option linter true

def name_str_is_accepted_style (s : string) : bool :=
  if (64 < s.head.to_nat) && (s.head.to_nat < 91)
  then bool.tt
  else bool.ff

def name_is_accepted_style : name → bool
| (name.mk_numeral i n) := name_is_accepted_style n
| (name.mk_string str n) :=
name_str_is_accepted_style str.reverse &&
name_is_accepted_style n
| (name.anonymous) := bool.tt

open tactic

private meta def warn_declaration_name : declaration → tactic unit :=
fun d,
  let name := d.to_name in
  if name_is_accepted_style name
  then constructor
  else trace $ "warning: `" ++ to_string name ++ "` violates style guidelines"

    -- we need to pass in position or commit it to env first?
    -- do env ← get_env,
    --       pos ← env.decl_pos name,
    --       tactic.save_info_thunk pos (fun u, to_fmt "hello world"),
    --       constructor,
    --       trace "done"

@[linter] meta def warn_names : linter :=
⟨ warn_declaration_name ⟩

def Foo : nat := 1

def b : nat := 2

def baz : nat := 3
```

linting

### DIFF
--- a/src/frontends/lean/definition_cmds.cpp
+++ b/src/frontends/lean/definition_cmds.cpp
@@ -32,6 +32,7 @@ Author: Leonardo de Moura
 #include "library/compiler/vm_compiler.h"
 #include "library/compiler/rec_fn_macro.h"
 #include "library/tactic/eqn_lemmas.h"
+#include "library/linter.h"
 #include "frontends/lean/parser.h"
 #include "frontends/lean/tokens.h"
 #include "frontends/lean/elaborator.h"
@@ -308,9 +309,18 @@ static certified_declaration check(parser & p, environment const & env, name con
                 msg.get_text_stream().get_stream()
                     << "type checking time of " << c_name << " took " << display_profiling_time{duration} << "\n";
                 msg.report();
-            });
+        });
+
+        if (linting_enabled(p.get_options())) {
+            lint_declaration(env, d);
+        }
+
         return ::lean::check(env, d);
     } else {
+        if (linting_enabled(p.get_options())) {
+            lint_declaration(env, d);
+        }
+
         return ::lean::check(env, d);
     }
 }
--- a/src/frontends/lean/definition_cmds.cpp
+++ b/src/frontends/lean/definition_cmds.cpp
@@ -394,7 +394,7 @@ declare_definition(parser & p, environment const & env, def_cmd_kind kind, buffe
     }
 
     if (linting_enabled(p.get_options())) {
-        lint_declaration(env, p, def);
+        lint_declaration(new_env, p, def);
     }
 
     new_env = compile_decl(p, new_env, c_name, c_real_name, pos);
--- a/src/frontends/lean/definition_cmds.cpp
+++ b/src/frontends/lean/definition_cmds.cpp
@@ -394,7 +394,7 @@ declare_definition(parser & p, environment const & env, def_cmd_kind kind, buffe
     }
 
     if (linting_enabled(p.get_options())) {
-        lint_declaration(new_env, p, def);
+        lint_declaration(new_env, def);
     }
 
     new_env = compile_decl(p, new_env, c_name, c_real_name, pos);
--- a/src/frontends/lean/definition_cmds.cpp
+++ b/src/frontends/lean/definition_cmds.cpp
@@ -311,16 +311,8 @@ static certified_declaration check(parser & p, environment const & env, name con
                 msg.report();
         });
 
-        if (linting_enabled(p.get_options())) {
-            lint_declaration(env, d);
-        }
-
         return ::lean::check(env, d);
     } else {
-        if (linting_enabled(p.get_options())) {
-            lint_declaration(env, d);
-        }
-
         return ::lean::check(env, d);
     }
 }
@@ -399,6 +391,10 @@ declare_definition(parser & p, environment const & env, def_cmd_kind kind, buffe
 
     if (!modifiers.m_is_private) {
         new_env = ensure_decl_namespaces(new_env, c_real_name);
+    }
+
+    if (linting_enabled(p.get_options())) {
+        lint_declaration(env, def);
     }
 
     new_env = compile_decl(p, new_env, c_name, c_real_name, pos);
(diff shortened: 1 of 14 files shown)


src/frontends/lean/definition_cmds.cpp
```diff
@@ -314,9 +315,18 @@ static certified_declaration check(parser & p, environment const & env, name con
...
+
+        if (linting_enabled(p.get_options())) {
+            lint_declaration(env, d);
+        }
+
         return ::lean::check(env, d);
     } else {
+        if (linting_enabled(p.get_options())) {
+            lint_declaration(env, d);
+        }
```
Kha: You probably want to move this into `declare_definition` as the last step so that the declaration and its attributes are already registered in the env.

gebner:
With this approach, we need to add imports for every linter that we want to use, right?  So if you import a "typical analysis library" you will now also import a linter, a spellchecker, a counterexample generator, (and slightly generalizing) a search engine, a debugger, a fancy REPL, etc.  Do you think this is a problem?

> What is the right way to query position information, given the decl hasn't been added to the environment yet, it seems we should use save_info_thunk for attaching warnings.

We could probably add a `get_pos_info : expr -> tactic pos` function that queries the position of an expression.  Then you could even get the position of subterms.

> I'm also not sure what the success/failure semantics should be, it seems that in a -Werror mode we would want any generated diagnostics to cause an error in Lean.

I'm not a big fan of `-Werror` in general, but this would be the way to do it.  Right now, Lean fails if there is at least one message with ERROR severity.  In `-Werror` it would just fail if there was at least one message with severity ERROR or WARNING.

src/library/linter.cpp
```diff
@@ -0,0 +1,72 @@
...
+}
+
+void initialize_linter() {
+    g_linter = new name{"linter"};
+
+    register_system_attribute(basic_attribute::with_check(
+            "linter", "a declaration linter",
+            [](environment const & env, name const & decl_name, bool) {
+                auto decl = env.find(decl_name);
+                std::cout << "in linter check" << decl->get_type() << std::endl;
```
gebner: This line should probably be removed before merging.

src/frontends/lean/definition_cmds.cpp
gebner: I agree.  Aside from that, the linter function should also be invoked on lemmas, and probably on examples as well.

jroesch: Sounds good, one of my questions was where this code should be.

src/library/linter.cpp
jroesch: Yeah I mean to assert the type of the linter here, just haven't implemented it yet.

src/library/vm/vm_pos_info_provider.h
```diff
@@ -0,0 +1,17 @@
...
+Copyright (c) 2017 Microsoft Corporation. All rights reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+
+Author: Jared Roesch
+*/
+#pragma once
+#include "library/vm/vm.h"
+
+namespace lean {
+pos_info to_pos_info_provider(vm_obj const & o);
```
gebner: The return type of this declaration is wrong.

src/library/vm/vm_pos_info_provider.cpp
```diff
@@ -0,0 +1,61 @@
...
+#include "library/vm/vm_pos_info.h"
+#include "kernel/pos_info_provider.h"
+#include "library/vm/vm_nat.h"
+#include "library/vm/vm_expr.h"
+#include "library/vm/vm_option.h"
+
+namespace lean {
+
+struct vm_pos_info_provider : public vm_external {
+    pos_info_provider const & m_val;
```
gebner: This is terribly unsafe if this `vm_obj` outlives the parser, e.g. if its stored in an info_manager, or used asynchronously in a different task (all easily possible since ts_clone is implemented).

I think a much simpler and safer design would be to add a single new (impure) function `tactic.expr_pos : expr → tactic expr`, which just calls `get_pos_info_provider().get_pos_info_or_some(to_expr(...))`.  We can imagine that the pos_info_provider is passed via the tactic_state, and we don't need to pass it around.

Another alternative option is to store a parser_pos_info_provider here (maybe in a `std::shared_ptr<pos_info_provider const>` if you like an extra sprinkle of genericity).

src/library/linter.cpp
```diff
@@ -0,0 +1,75 @@
...
+
+    local_context lctx;
+    tactic_state tac_st = mk_tactic_state_for(env, opts, {"Linter"}, lctx, mk_constant("true"));
+
+    auto linter_names = get_linters(env);
+    for (auto linter_name : linter_names) {
+        // NB: because we currently have 1 field inductive, this requires no unboxing
+        auto linter_fn = S.get_constant(linter_name);
+        vm_obj linter_result = S.invoke(linter_fn, to_obj(prov), to_obj(decl), to_obj(tac_st));
+        if (is_constructor(linter_result) && cidx(linter_result) == 0) {
```
gebner: Minor nitpick: this should be `tactic::is_result_success(linter_result)`.

Kha:
> I think a much simpler and safer design would be to add a single new (impure) function tactic.expr_pos : expr → tactic expr, which just calls get_pos_info_provider().get_pos_info_or_some(to_expr(...)).

I agree, we should always have a global PIP at tactic execution time.

src/library/vm/vm_pos_info_provider.cpp
jroesch: Yeah I agree. I was worried about this yesterday, but no one was awake/around to ask about the right way to do this. I'll refactor it to use this approach instead.

src/library/linter.cpp
jroesch: 👍

jroesch:
I think we should just shelve this until we make progress on #1674. It seems like the design would be much cleaner, and easier to implement/maintain. Thoughts?

PR #1537, opened 2017-04-24
Refactor structure compilation

Kha:
This PR introduces a new default strategy for compiling structure inheritance by nesting instead of flattening. See https://github.com/leanprover/lean/wiki/Refactoring-structures for the reasoning.
Theories that depend on diamond inheritance (`init/algebra`) have been switched to `set_option old_structure_cmd true` for now.

Minor to-dos:
* Should `x.f` also work with a definition `f` in some base namespace of `x`? The code for translating that is already there, it's just looking up `f` that's missing.
* Should we remove `fmap` too? Replacing it with `<$>` is easy now that we have operator sections.


library/init/data/option_t.lean
```diff
@@ -36,16 +36,15 @@ instance {m : Type u → Type v} [monad m] : monad (option_t m) :=
...
  end,
  bind_assoc := begin
    intros,
-   dsimp [option_t_bind, option_t_return, bind],
+   dsimp [option_t_bind, option_t_return],
    rw [monad.bind_assoc],
    apply congr_arg, apply funext, intro x,
    cases x,
-   { dsimp [option_t_bind._match_1, return, pure],
+   { dsimp [option_t_bind],
```
gebner, line 47: There are a lot of simp calls where this PR removes the `._match_1`.  Is this a change that became possible through the structure refactoring, or is this just general cleanup?

leodemoura: The `._match_1` suffixes were not necessary even before this PR. It is great they are gone :)

tests/lean/cyclic_default_fields.lean
```diff
@@ -1,7 +1,6 @@
 class Eq (α : Type) :=
-(eq : α → α → Prop)
+(eq : α → α → Prop := λ a b, ¬ne a b)
 (ne : α → α → Prop := λ a b, ¬eq a b)
```
leodemoura, line 3: Wow, I'm impressed this stuff works.

src/frontends/lean/structure_cmd.cpp
```diff
@@ -499,55 +639,101 @@ struct structure_cmd_fn {
...
-                if (auto fidx = merge(parent, fname, ftype)) {
-                    fmap.push_back(*fidx);
-                    field = m_fields[*fidx].local;
-                    if (local_info(field) != binding_info(intro_type)) {
-                        throw elaborator_exception(parent,
-                                                   sstream() << "invalid 'structure' header, field '" << fname <<
-                                                   "' has already been declared with a different binder annotation");
+            if (m_subobjects) {
+                name fname;
+                if (auto const & ref = m_parent_refs[i])
```
jroesch, line 644: this is just a minor nit, but it seems good to always use curly braces on `if`/`else`

src/frontends/lean/structure_cmd.cpp
```diff
@@ -680,56 +864,71 @@ struct structure_cmd_fn {
...
+
+    /* let-elaborate new default values of fields with explicit types so that they can refer to any field, even cyclically.
+     *
+     *   class Eq (α : Type) :=
+     *   (eq : α → α → Prop := λ a b, ¬ne a b)
+     *   (ne : α → α → Prop := λ a b, ¬eq a b)
+     *
+     * will be elaborated as
+     *
+     *   Π (α : Type) (eq : α → α → Prop) (ne : α → α → Prop),
```
jroesch, line 895: It isn't clear to me how this works, not essential to the PR but could be good to clarify for future readers of the code.

src/frontends/lean/structure_cmd.cpp
```diff
@@ -1019,11 +1222,21 @@ struct structure_cmd_fn {
...
         for (unsigned i = 0; i < m_parents.size(); i++) {
             expr const & parent            = m_parents[i];
-            field_map const & fmap         = m_field_maps[i];
             buffer<expr> parent_params;
             expr const & parent_fn         = get_app_args(parent, parent_params);
             levels const & parent_ls       = const_levels(parent_fn);
             name const & parent_name       = const_name(parent_fn);
+            if (m_subobjects) {
+                if (!m_private_parents[i]) {
+                    if (m_attrs.has_class() && is_class(m_env, parent_name)) {
```
jroesch, line 1231: is it possible to collapse these into a single `if`?

jroesch (commented):
Looks good, excited to use this in the new algebraic stuff.

tests/lean/cyclic_default_fields.lean
Kha: Realized this only while working on the PR, even though they are completely independent :) .

src/frontends/lean/structure_cmd.cpp
Kha: I usually just copy the style of the surrounding code, so no braces if it's Leo's and always braces if it's Gabriel's ;) . But I'm also happy to adopt a consistent style in the future.

Issue #1535, opened 2017-04-23
Transfer does not work with universe metavariables


gebner:
The following snippet fails at `dlist.transfer`, but works as soon as I change `Type _` to `Type u`.
```lean
import data.dlist
universe u

example (α : Type _) (a b c : dlist α) : (a ++ b) ++ c = (a ++ b) ++ c :=
begin
revert a b c,
dlist.transfer,
intros,
simp,
end
```

leodemoura:
@gebner Note that it works if we use `lemma` instead of `example`.
It works because we fully elaborate the lemma/theorem type before we process the body.
So, all universe meta-variables have been already resolved and assigned to universe parameters.

One easy solution is for `transfer` to sign an error message if the context contains universe meta-variables. I don't think this is a big limitation. 

@johoelzl: If the limitation above is easy to remove, then, of course, it is a better solution.

Kha:
@leodemoura
> It works because we fully elaborate the lemma/theorem type before we process the body.

Should we do that for every declaration that has an explicit type? Or is there code that depends on this?
If we change it, would it be a good idea to make even elaborating definitions async?

leodemoura:
> Should we do that for every declaration that has an explicit type?

@kha I think this is a good idea.

> Or is there code that depends on this?
> If we change it, would it be a good idea to make even elaborating definitions async?

I think this is feasible, but we would have to use `task`s for definitions too.
@gebner Do you see problems?

johoelzl:
Might be a simple bug. I guess I just need to handle the meta variables like a local variable. I will investigate it tomorrow.

gebner:
> I think this is feasible, but we would have to use tasks for definitions too.
> @gebner Do you see problems?

I see two issues:
1) Is it clear how this would interact with `variables`?
2) Making definitions async would require us to make lots of other changes:
  * We would need to make VM compilation async as well.  Creating a `vm_state` must not wait for any definition that is not executed.
  * How we would we declare equations for async definitions?
  * Quite a few declarations would need to wait for some definition to finish elaborating.  (Just type-checking the type can require delta-reduction.)  I have no empirical data on how bad this would get in practice.  Remember, every time we wait for something we need an extra thread.
  * Noncomputability annotations would also need to be computed async.

leodemoura:
@gebner Thanks for the feedback. I retract my comment: "I think this is feasible".

johoelzl:
[...]

PR #1538, opened 2017-04-24
fix(library/init/meta/transfer): add check if target contains (univer…

johoelzl:
…se) meta variables (see #1535)


johoelzl:
Rebasing resulted in closing and reopening the PR.

PR #1541, opened 2017-04-25
fix(frontends/lean/inductive_cmds): do not whnf pre-exprs

Kha:
@leodemoura Does this change look sound to you?


leodemoura:
@kha Thanks for fixing this bug. Just to double check: the problem was that    `get_datatype_result_level` was being invoked with pre-terms, right?

I didn't find any problems with your fix.

@dselsam Do you agree?

Kha:
Yes, as well as `update_result_sort`, which also uses whnf.

dselsam:
The fix looks fine to me.

PR #1550, opened 2017-04-27
fix(frontent/lean/elaborator): check if field is found in structure update

### LINKED ISSUE #1549: Segfault on structure update in bad definition
```lean
structure st :=
(i : ℕ)

def g : st → unit × st :=
λ s, {s with i:=0}
```

johoelzl:
Fixes #1549

@Kha is this the right fix? A couple of lines later, at https://github.com/johoelzl/lean/blob/fef58aa304b9de6208eab960eac38d6e4bf42b3a/src/frontends/lean/elaborator.cpp#L2615 is the same exception again. What are the reasons that `visit(f, ...)` could fail?


Kha:
Ahh I see. In the old code, `visit` would have complained about not being able to resolve the non-existing projection, but now `find_field` complains instead. So I think you can remove the try-catch.

johoelzl:
Okay, I removed the following try-catch block.

PR #1554, opened 2017-04-28
fix(frontends/lean/pp): insert line breaks in notations

gebner:
The pretty-printer now allows line breaks in notations.  For example for infix operators we insert a line break after the operator, it looks like this:

```lean
def foo : ℕ → ℕ :=
λ (xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx : ℕ),
  xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx +
    xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx
```

(Right now the addition would be printed on a single line.)

It does not always produce the prettiest output, but it's good enough to make long equations (somewhat) readable.


gebner:
Oh, I just noticed that this triggers an assertion violation in the format code.  I'll look at it.

PR #1555, opened 2017-04-29
fix(library/vm/vm_environment) use is_structure rather than try to catch an assertion

kim-em:
`get_structure_fields` throws an assertion if its argument is not a structure, so there's no point trying to catch an exception. Instead we should just use `is_structure`.

(Sorry, my git-fu is weak, and there are more commits here than necessary.)


kim-em:
This fix is necessary for @Kha's new `congr_struct` tactic; without it we crash lean whenever attempting to apply congr_struct to equalities that are not between structures.

Kha:
Thanks for the fix! I've cherry-picked it onto master.

PR #1556, opened 2017-04-29
feat(library/init/meta/simp_tactic,library/init/meta/interactive) int…

kim-em:
…eractive tactics for unfold_projection

### DIFF
--- a/library/init/meta/interactive.lean
+++ b/library/init/meta/interactive.lean
@@ -782,6 +782,17 @@ meta def delta : parse ident* → parse location → tactic unit
 | cs [] := do new_cs ← to_qualified_names cs, tactic.delta new_cs
 | cs hs := do new_cs ← to_qualified_names cs, delta_hyps new_cs hs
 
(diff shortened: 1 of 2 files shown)


kim-em:
I've also added a commit that fixes the behaviour of all tactics that work by reverting hypotheses. Previously, we had missed the case where reverting the hypotheses creates an `elet` rather than `pi`. This in particular meant that hypotheses created via `pose` could not be simplified.

To achieve this, I wrote a helper function `revert_and_transform` which handles the reversion, matching, and reintroduction, and refactored all the tactics that work with hypotheses to use this.

leodemoura:
We are currently working on the algebraic normalizer and simplifier. I will review/merge this PR after we are done with them.

semorrison:
That's fine, but for me this pull request has nothing to do with abstract algebra as developed in there Lean standard library, or algebraic normalisation. I just want to be able to easily unfold projections on structures I define.

PR #1565, opened 2017-05-03
chore(library/inductive_compiler/nested.cpp): prove all theorems in C++

dselsam:
Don't call tactics in the inductive compiler so that we can use it when defining `pexpr`.


dselsam:
hold on, fixing build issue...

dselsam:
Fixing broken test...

jroesch:
👍

leodemoura:
@dselsam It passed all tests. So, I am assuming it is ready to be merged, right?

dselsam:
Yes.

PR #1567, opened 2017-05-04
feat(library/data/list, library/data/array): theorems needed for new hash_map

digama0:
This commit adds a verified `hash_map` implementation using the same algorithm as before. Several facts were imported from `library_dev` to support it (but no more than necessary).

This commit is a bit of a grab-bag of theorem additions needed by my projects, and in particular `nat/lemmas.lean` also has several new theorems ported from lean 2 which are relevant for an unrelated project on analysis of the pell equation.


library/init/data/list/basic.lean
```diff
@@ -133,6 +138,11 @@ def find [decidable_eq α] : α → list α → nat
 | a []       := 0
 | a (b :: l) := if a = b then 0 else succ (find a l)
 
+def drop : ℕ → list α → list α
```
gebner: This is the same as `remove_nth`, right?

library/data/hash_map.lean
```diff
@@ -65,34 +124,597 @@ def erase_aux (a : α) : list (Σ a, β a) → list (Σ a, β a)
...
-            let nz' : nbuckets' > 0 := mul_pos nz dec_trivial in
-            ⟨hash_fn, size', nbuckets', nz',
-            fold_buckets buckets' (mk_array nbuckets' []) $
-               λ r a b, reinsert_aux hash_fn nz' r a b⟩
+inductive valid_aux {α : Type u} {β : α → Type v} (idx : α → nat) : Π (l : list (list Σ a, β a)) (sz : nat), Prop
+| nil {} : valid_aux [] 0
+| cons : Π (c : list Σ a, β a) {l sz}, valid_aux l sz → (c.map sigma.fst).nodup →
+  (∀ a ∈ c, idx (sigma.fst a) = l.length) → valid_aux (c::l) (sz + c.length)
+
+def valid_aux.unfold_cons {idx : α → nat} : Π {c l sz}, valid_aux idx (c::l) sz →
```
gebner: All of these `def`s should probably be `lemma`s.

library/init/data/set.lean
```diff
@@ -95,4 +91,22 @@ instance : functor set :=
...
+notation `⋃` binders `, ` r:(scoped f, Union f) := r
+notation `⋂` binders `, ` r:(scoped f, Inter f) := r
+
+@[reducible]
+def sUnion (s : set (set α)) : set α := {t | ∃ a ∈ s, t ∈ a}
+
+@[reducible]
+def sInter (s : set (set α)) : set α := {t | ∀ a ∈ s, t ∈ a}
+
+prefix `⋃₀`:110 := sUnion
```
johoelzl: Uh why? This will interfere with library_dev and you do not need it!

library/init/data/list/basic.lean
digama0: oops, you're right - I'll take this out.

library/init/data/set.lean
digama0: At the very least, `Union` needs to be renamed to `sUnion` for consistency. The rest is just for uniformity. I already have the repair to library_dev prepared, but since notation can't be shadowed (it instead introduces ambiguous translations, which cause errors), I needed to prime the notation as well:

     def Union' {α : Sort u} (s : α → set β) : set β := supr s
     
     @[reducible]
    -def Inter {α : Sort u} (s : α → set β) : set β := infi s
    +def Inter' {α : Sort u} (s : α → set β) : set β := infi s
     
    -notation `⋃` binders `, ` r:(scoped f, Union' f) := r
    -notation `⋂` binders `, ` r:(scoped f, Inter f) := r
    +notation `⋃'` binders `, ` r:(scoped f, Union' f) := r
    +notation `⋂'` binders `, ` r:(scoped f, Inter' f) := r

To avoid confusion, I'll remove this and put it in a separate PR, though.

library/data/hash_map.lean
digama0: I changed them to `theorem` and then noticed that you recommend `lemma` instead. Is there an actual difference between `lemma` and `theorem`?

leodemoura:
I will merge this PR after we finish the algebraic normalizer and modifications to the simplifier.

library/data/hash_map.lean
gebner: The name "theorem" conveys a sense of significance, that's it.  Implementation-wise, they are virtually the same since they are token aliases: the only component of Lean that ever sees the difference is the scanner, for both lemma and theorem it returns the `theorem` token.

leodemoura:
@digama0 I'm merging this PR today. The algebraic normalizer is far from done, but we have a customer for the hash_maps.

leodemoura:
@digama0 I merged the PR  #1570 first because of missing dependencies (e.g., `nth_le`). Now, they are a bunch of conflicts. Do you have time to take a look, and rebase this PR with respect to the current `master` branch?

digama0:
ok

On Tue, May 16, 2017 at 5:56 PM, Leonardo de Moura <notifications@github.com

digama0:
@leodemoura Wait, now I'm confused. It looks like you already merged both branches `list_ops` and `hash_map` into master. I tried the merge myself, but it looks like you did all the same things, and the result turned out to be empty. So, I think you can safely close these two.

leodemoura:
@digama0 Sorry, I probably got confused when merging. I tried first this one, but it needed `nth_le` that was in the other PR. Then, when I switched to the other PR, I may have put it on top of this one.

PR #1569, opened 2017-05-04
Preserve type ascriptions during elaboration

gebner:
It is a bit surprising that type ascription does not affect the inferred type of an expression at the moment.  This usually shows up in the context of type class inference (are there other places where we care about the inferred type?):
```lean
def looks_like_git_url (dep : string) : bool :=
':' ∈ (dep : list char)
-- does not work since we infer the type `string` for `dep` here
```
I would have expected this to work.  Mario was also independently surprised about this behavior.

This PR preserves the type ascription during elaboration.  That is, they are present in the elaborated expressions, but they are removed during finalization.

The main downside is that we now need to unwrap type ascriptions in many places.  A particularly egregious example is in `comp_val.cpp`: here we need to unwrap all arguments in every function, since we might want to prove `(10:nat) < 20`...

Alternatives:
 * Use the `show list char, from dep` workaround.

### DIFF
--- a/src/frontends/lean/elaborator.cpp
+++ b/src/frontends/lean/elaborator.cpp
@@ -825,7 +825,7 @@ expr elaborator::visit_typed_expr(expr const & e) {
     expr new_val      = visit(val, some_expr(new_type));
     expr new_val_type = infer_type(new_val);
     if (auto r = ensure_has_type(new_val, new_val_type, new_type, ref))
-        return *r;
+        return mk_typed_expr(new_type, *r);
 
     auto pp_data = pp_until_different(new_val_type, new_type);
     auto pp_fn   = std::get<0>(pp_data);
@@ -3526,6 +3526,14 @@ struct sanitize_param_names_fn : public replace_visitor {
 
     virtual expr visit_sort(expr const & e) override {
         return update_sort(e, sanitize(sort_level(e)));
+    }
+
+    virtual expr visit_macro(expr const & e) override {
+        if (is_typed_expr(e)) {
+            return visit(get_typed_expr_expr(e));
+        } else {
+            return replace_visitor::visit_macro(e);
+        }
     }
 
     void collect_params(expr const & e) {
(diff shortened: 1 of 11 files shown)


avigad:
@gebner I agree that it is counterintuitive that the term `(t : T)` does not infer type `T`. IIRC, in Lean 2, it was originally notation for the identity function (which had this property), but we switched it because of an issue raised by @fpvandoorn. But now I cannot remember the issue, and I cannot seem to find the discussion on github.

@fpvandoorn Do you remember this? (Am I just imagining it?)

leodemoura:
> @gebner I agree that it is counterintuitive that the term (t : T) does not infer type T

@gebner @avigad Clarification: type ascription guarantees that `t` has a type that is definitionally equal to `T`. However, it does not guarantee that it is syntactically equal to `T`. See:
https://github.com/leanprover/lean/blob/master/src/frontends/lean/elaborator.cpp#L827

This may have a negative impact on type class resolution because it will only unfold `reducible` constants.

leodemoura:
@gebner I will wait a little bit before we merge this PR. I want to think about its consequences.
The downside you have described will affect other modules and automation in general.

fpvandoorn:
> @fpvandoorn Do you remember this? (Am I just imagining it?)

I believe the issue I had was that in Lean 2, `(t : T)` was sugar for `@id T t`, which means that the following two definitions are not syntactically equal:
```lean
definition foo1 := succ (0 : nat)
definition foo2 := succ 0
```
This was annoying in cases where you cared about the exact form of the definition. But if I understand @gebner correctly, they will be syntactically equal with this PR.

As a user, I like @gebner behavior a lot. I've sometimes tried to use `(t : T)` to help typeclass inference (like in the tests @gebner added), but that doesn't work if the type of `t` is not forced to be `T`

leodemoura:
I thought about this PR, and I think we should not merge it.
We can use `@id T t` or `show T, from t` to achieve the desired behavior when needed.
If needed we can define a nicer notation for `@id T t`.

As I said above, `(t : T)` guarantees that `t` has a type that is definitionally equal to `T`. 

So, I'm closing this PR

PR #1574, opened 2017-05-08
feat(library/type_context): add option to unfold lemmas

gebner:
As reported by @digama0 on slack, is it currently impossible to prove some lemmas by `rfl`, even though the kernel would accept `rfl` as proof.  In contrast to irreducible definitions, there is (almost) no way to work around this issue, such as by setting `[reducible]` attributes.

This PR adds a `type_context.unfold_lemmas` option as a way to locally override this behavior.
```lean
open nat well_founded

def gcd.F : Π (y : ℕ), (Π (y' : ℕ), y' < y → nat → nat) → nat → nat
| 0        f x := x
| (succ y) f x := f (x % succ y) (mod_lt _ $ succ_pos _) (succ y)

def gcd (x y : nat) := fix lt_wf gcd.F y x

set_option type_context.unfold_lemmas true
@[simp] lemma gcd_zero_right (x : nat) : gcd x 0 = x := rfl
```

Alternatives:
 * Allow `local attribute [reducible] fix._proof_1`.  This was my first approach, but given the error message `expression has type ?m_2 = ?m_2 but is expected to have type gcd x 0 = x`, how on earth do you figure out that `fix._proof_1` needs to be made reducible?
 * Wait for well-founded recursion support in the equation compiler.  Users should not unfold lemmas.

### DIFF
--- a/src/library/type_context.cpp
+++ b/src/library/type_context.cpp
@@ -41,9 +41,14 @@ Author: Leonardo de Moura
 #define LEAN_DEFAULT_NAT_OFFSET_CNSTR_THRESHOLD 1024
 #endif
 
+#ifndef LEAN_DEFAULT_UNFOLD_LEMMAS
+#define LEAN_DEFAULT_UNFOLD_LEMMAS false
+#endif
+
 namespace lean {
 static name * g_class_instance_max_depth = nullptr;
 static name * g_nat_offset_threshold     = nullptr;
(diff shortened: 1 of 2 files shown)


leodemoura:
> This PR adds a type_context.unfold_lemmas option as a way to locally override this behavior.

@gebner I think this is a reasonable solution. 
We don't unfold lemmas by default during elaboration because of performance problems.a

PR #1576, opened 2017-05-09
Locals in expr quotations

Kha:
This PR
* changes the type of \`\`\`(e) from `expr` to `reflected e` (if it does not contain anti-quotations) and installs a coercion `reflected e -> expr` for backwards compatibility
* allows the use of locals in expr quotations as long as it can find `reflected` instances for them, by substituting them with the instance's `expr` value; thus allowing [polymorphic uses](https://github.com/leanprover/lean/commit/ba56dba8fe00259f6011eae937c89229981d972e#diff-2789db2892ce7f5156d12ec6b98c7205R2) of `eval_expr`
* replaces the `has_quote` class with a `has_reflect` class that produces terms instead of pre-terms, [simplifying](https://github.com/leanprover/lean/commit/397ed9ca75818766f2ec1fcbabe820b60d00b0c2#diff-8f2daa0063a7f644ec24982b860eb2f5) expr synthesis and reducing the need for `to_expr` run-time elaboration

The last feature counts as a happy little accident. I originally assumed that if we wanted to rewrite code like

```lean
def f (n : name) : tactic expr := to_expr ``(foo %%(quote n))
```

to

```lean
def f (n : name) : expr := ```(foo n)
```

we'd have to include a new parameter `[reflected n]`, which may make the definition more cumbersome to use. But then I realized that by combining the new `reflected` literals with the power of dependent pattern matching, we can indeed give an instance of `reflected n` [for any `n : name`](https://github.com/leanprover/lean/commit/397ed9ca75818766f2ec1fcbabe820b60d00b0c2#diff-33bf903e3c0fc2fb366520ca7ecad2deR23)! Though I still had to [cheat](https://github.com/leanprover/lean/commit/397ed9ca75818766f2ec1fcbabe820b60d00b0c2#diff-33bf903e3c0fc2fb366520ca7ecad2deR15) in two cases.

Note that, to me, this PR still feels incomplete - it seems natural to next introduce a type `typed_expr` that is indexed by the type of the quoted term and that falls strictly between `reflected` and `expr`. Like `reflected`, its wellformedness would be certified (unless you do unchecked casts, of course), but unlike `reflected`, it would also be able to represent open terms. An expr quotation taking `typed_expr` antiquotations would then use the types for elaboration and itself return an appropriate `typed_expr`. I haven't thought about it in detail yet though.


src/frontends/lean/tactic_notation.cpp
```diff
@@ -141,7 +142,12 @@ static expr parse_interactive_param(parser & p, expr const & ty, expr const & qu
     auto env = eval.compile(n, quote_inst);
     vm_state S(env, p.get_options());
     auto vm_res = S.invoke(n, vm_parsed);
-    return to_expr(vm_res);
+    expr r = to_expr(vm_res);
+    if (is_app_of(r, get_pexpr_subst_name())) {
+        return r; // HACK
```
Kha, line 147: This hack is due to the fact that we cannot actually reflect all `pexpr`s - to do that we would have to know which `to_pexpr` instance to use for each antiquotation - but these may not be elaborated yet. I propose to have `pexpr.reflect` only work with `pexpr` antiquotations, which in practice just means that you cannot use the \`[...] tactic quotations with `expr` antiquotations.

library/init/meta/has_reflect.lean
```diff
@@ -0,0 +1,44 @@
...
+| (name.mk_numeral i n) := ```(λ n, name.mk_numeral i n).subst (name.reflect n)
+
+meta instance list.reflect {α : Type} [has_reflect α] [reflected α] : has_reflect (list α)
+| []     := ```([])
+| (h::t) := ```(λ t, h :: t).subst (list.reflect t)
+
+meta instance option.reflect {α : Type} [has_reflect α] [reflected α] : has_reflect (option α)
+| (some x) := ```(_)
+| none     := ```(_)
+
```
leodemoura, line 35: How does
```lean
```(_)
```
work?

leodemoura:
Is the following supported
```lean
```(%%a -> %%b)
```
?
Or, do we have to use `pexpr`s in this case? That is,
```lean
``(%%a -> %%b)
```
I'm just trying to understand.

library/init/meta/has_reflect.lean
Kha: \`\`\`(_) is elaborated to `[expr_quote ?m_1]` with inferred type `reflected ?m_1`, which is then unified with the expected type `reflected none`. Another unplanned feature :) .

Kha: Note that the only reason we cannot use `_` everywhere (well, apart from the "special" instances using `unchecked_cast`) is that recursive instances like `list.reflect` are excluded from class inference. I'm not complaining though, it still sounds like a reasonable restriction.

leodemoura: > is elaborated to [expr_quote ?m_1] with inferred type reflected ?m_1, which is then unified with the expected type reflected none. Another unplanned feature :) .

Cool!

Kha:
@leodemoura That [used](https://github.com/leanprover/lean/pull/1576/files#diff-30ea0d29fe88463c9d644a707691a52aR84) to be supported, but mostly by accident. Previously, quotations would be elaborated during parsing (which worked because they were supposed to be closed), which created expressions like
```
expr.subst (expr.subst ```(λ (a : Type u_1) (b : Type u_2), a → b) a) b
```
with bad free universe parameters - it just so happened that they were gone after substitution.

Now that expr quotes are not elaborated separately anymore, we don't get this behavior for free anymore. We could still simulate it, but because it stops working as soon as we quote a constant that depends on the universe parameter anyway, I just replaced it with `expr.imp` everywhere.

leodemoura:
> Now that expr quotes are not elaborated separately anymore, we don't get this behavior for free anymore. We could still simulate it, but because it stops working as soon as we quote a constant that depends on the universe parameter anyway, I just replaced it with expr.imp everywhere.

Good for me. I was just curious.

Kha:
Another observation: This change basically gives us `expr` autoquotation. Indeed, \`\`\`(a) (outside of patterns) can now be interpreted as syntax sugar for `reflect a`, where `reflect` is defined as simply
```lean
meta def reflect {α : Type u} (a : α) [h : reflected a] : reflected a := h
```

PR #1578, opened 2017-05-11
Fixes

Kha:


src/frontends/lean/scanner.cpp
```diff
@@ -704,7 +704,7 @@ void scanner::skip_to_pos(pos_info const & pos) {
     for (unsigned line_no = 1; line_no < pos.first; line_no++)
         fetch_line();
     m_line = m_sline;
-    for (unsigned col_idx = 0; col_idx < pos.second; col_idx++)
+    while (static_cast<unsigned>(m_upos) < pos.second)
```
Kha, line 707: @gebner This triggered the assertion below, but curiously only when navigating to the last line of a file. Couldn't reproduce it in a test case either.

src/frontends/lean/decl_util.cpp
```diff
@@ -273,8 +273,6 @@ void collect_implicit_locals(parser & p, buffer<name> & lp_names, buffer<expr> &
     for (expr const & e : all_exprs) {
         collect_locals_ignoring_tactics(e, locals);
         lp_found = collect_univ_params_ignoring_tactics(e, lp_found);
-        if (is_local(e))
-            given_params.insert(mlocal_name(e));
```
Kha, line 277: @leodemoura Does this look correct to you?

leodemoura: @kha Yes, it does. I will merge this PR today.

PR #1579, opened 2017-05-11
Wildcard locations for tactics

jroesch:
@Kha will you review this PR? 

I added basic support for locations with wild cards, and updated the tactics which use `location` with the correct semantics.

For example you can now do:
```
rewrite [foo, bar] at *
```

or

```
simp [foo, bar] at *
```

I have a few more test cases incoming as well.

### DIFF
--- a/library/init/meta/interactive.lean
+++ b/library/init/meta/interactive.lean
@@ -20,6 +20,18 @@ namespace interactive
     to the tactic. -/
 @[reducible] meta def parse {α : Type} [has_reflect α] (p : parser α) : Type := α
 
+inductive loc : Type
+| wildcard : loc
+| ns : list name → loc
+
+def loc.empty := loc.ns []
+
+meta instance : has_reflect loc :=
+fun l, match l with
+| loc.wildcard := ```(_)
+| loc.ns xs := ```(_)
+end
+
 namespace types
 variables {α β : Type}
 
@@ -37,7 +49,7 @@ meta def ident_ : parser name := ident <|> tk "_" *> return `_
 meta def using_ident := (tk "using" *> ident)?
 meta def with_ident_list := (tk "with" *> ident_*) <|> return []
 meta def without_ident_list := (tk "without" *> ident*) <|> return []
-meta def location := (tk "at" *> ident*) <|> return []
+meta def location := (tk "at" *> (tk "*" *> pure loc.wildcard <|> (loc.ns <$> ident*))) <|> return loc.empty
 meta def qexpr_list := list_of (qexpr 0)
 meta def opt_qexpr_list := qexpr_list <|> return []
 meta def qexpr_list_or_texpr := qexpr_list <|> list.ret <$> texpr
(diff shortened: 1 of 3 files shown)


jroesch:
I need to push a couple more commits, took a break to work on artifact eval.

Kha:
Note: There is some prior work by @avigad at https://github.com/leanprover/library_dev/blob/master/tools/tactic/simp_tactic.lean, so this would probably be a good opportunity to merge those changes back into this repo.

leodemoura:
@jroesch Are you going to revise this PR?

avigad:
@jroesch My implementation of `simp_all` is almost the same as yours, except there is a bug in yours: If you try this example,
```lean
example (x y z w : ℕ) (h₁ : x + 0 = y) (h₂ : y + 0 = z) (h₃ : z + 0 = w) : x = w :=
begin
  simp at *
end
```
the (unfinished) goal reads as follows:
```
tactic failed, there are unsolved goals
state:
x y z w : ℕ,
h₁_1 : x = y,
h₂_1 : y = z,
h₃_1 : z = w
⊢ x = w
```
In other words, it renames `h₁` to `h₁_1`, etc., which is annoying. I fixed that in my version. It involved two small changes to Leo's code, found in the comments here: https://github.com/leanprover/library_dev/blob/master/tools/tactic/simp_tactic.lean#L62-L93

There are some other things in https://github.com/leanprover/library_dev/tree/master/tools/tactic, that might be useful:

- `simph_all`, which iteratively simplifies hypotheses and then the goal, using all previous hypotheses to simplify the next one
- `dunfold_all`, `delta_all`, `dsimp_all` useful for situations where you want to unfold a definition everywhere it appears
- versions of these for use in ordinary metaprogramming (not in interactcive mode)
- tactics `simp_only` and `simp_only_at` which do not use the default set simp rules, but rather simplify only with the list of rules you give it. 

The last two duplicate code from `simp` and `simp_at`. Leo said that he wanted to refactor the simplifier tactic library to add a boolean flag to control the inclusion of defaults. There is also the question as to what kind of syntax to use for this in interactive mode.

In addition, there is a subtle issue with tactics like `simp_at` (_not_ the interactive version), which change a hypothesis. Suppose `h` is a local constant, and in a metaprogram we write `simp_at h`. The problem is that `h` no longer refers to the corresponding local constant; `simp_at` creates a _new_ local constant with the same name and clears the old one. AFAICT, there is no nice way of getting that local constant, other than retrieving one with the same name and hoping there was no duplication. This is bad. Any tactic that creates a new hypothesis should return the corresponding local constant, the way that `e ←intro 'h` returns the local constant that was introduced. So the files in `library_dev` have alternative versions `simp_at'`, `assertv'` etc. that do this.

You are welcome to incorporate any of these, or let me know if I should do it.

jroesch:
@avigad thanks for the review, I agree that bug is annoying. I will try to incorporate your versions and update this PR!

leodemoura:
Subsumed by #1615

PR #1580, opened 2017-05-12
Unify expr and pexpr

Kha:
...but not in the C++ way :) . Still gets rid of all `to/of_raw_expr` (one could always use `unchecked_cast` if they really need them).

Also contains the commits of #1578 because I was too lazy to rebase and fix.

There are still a few broken tests - should we try and print `expr tt` as `expr`?

### DIFF
--- a/src/frontends/lean/pp.cpp
+++ b/src/frontends/lean/pp.cpp
@@ -447,6 +447,27 @@ bool pretty_fn::is_implicit(expr const & f) {
     }
 }
 
+bool pretty_fn::is_default_arg_app(expr const & e) {
+    if (m_implict || m_preterm)
+        return false; // showing default arguments
+    if (!closed(app_fn(e))) {
+        // the Lean type checker assumes expressions are closed.
+        return false;
+    }
+    try {
+        expr t = m_ctx.relaxed_whnf(m_ctx.infer(app_fn(e)));
+        if (is_pi(t)) {
+            expr arg_type = binding_domain(t);
+            t = binding_body(t);
+            if (!is_pi(t) && !is_var(t) && is_app_of(arg_type, get_opt_param_name(), 2)) {
+                expr defval = app_arg(arg_type);
+                return closed(defval) && defval == app_arg(e);
+            }
+        }
+    } catch (exception &) { }
+    return false;
+}
+
 bool pretty_fn::is_prop(expr const & e) {
     try {
         expr t = m_ctx.relaxed_whnf(m_ctx.infer(e));
@@ -839,6 +860,8 @@ auto pretty_fn::pp_field_notation(expr const & e) -> result {
 auto pretty_fn::pp_app(expr const & e) -> result {
     if (auto r = pp_local_ref(e))
         return *r;
+    if (is_default_arg_app(e))
+        return pp_child(app_fn(e), max_bp());
     expr const & fn = app_fn(e);
     if (m_structure_instances && is_structure_instance(m_env, e, m_implict))
         return pp_structure_instance(e);
--- a/library/init/meta/expr.lean
+++ b/library/init/meta/expr.lean
@@ -34,10 +34,10 @@ end⟩
 meta constant macro_def : Type
 
 /- Reflect a C++ expr object. The VM replaces it with the C++ implementation. -/
-meta inductive expr
-| var         : nat → expr
-| sort        : level → expr
-| const       : name → list level → expr
+meta inductive expr (elaborated : bool := tt)
+| var      {} : nat → expr
+| sort     {} : level → expr
+| const    {} : name → list level → expr
 | mvar        : name → expr → expr
 | local_const : name → name → binder_info → expr → expr
 | app         : expr → expr → expr
@@ -45,6 +45,8 @@ meta inductive expr
 | pi          : name → binder_info → expr → expr → expr
 | elet        : name → expr → expr → expr → expr
 | macro       : macro_def → list expr → expr
+
+variable {elab : bool}
 
 universes u v
 /-- (reflected a) is a special opaque container for a closed `expr` representing `a`.
@@ -58,10 +60,11 @@ meta constant reflected {α : Type u} : α → Type u
 meta constant reflected.to_expr {α : Type u} {a : α} : reflected a → expr
 meta constant reflected.subst {α : Type u} {β : α → Type v} {f : Π a : α, β a} {a : α} :
   reflected f → reflected a → reflected (f a)
+meta constant expr.reflect (e : expr elab) : reflected e
 meta constant string.reflect (s : string) : reflected s
 
 attribute [class] reflected
-attribute [instance] string.reflect
+attribute [instance] expr.reflect string.reflect
 
 meta instance {α : Type u} (a : α) : has_coe (reflected a) expr :=
 ⟨reflected.to_expr⟩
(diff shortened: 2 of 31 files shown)


gebner:
I was always a supporter of having `expr = pexpr`, and I like this change.  I think we should make almost all of `expr.lean` elaboration-polymorphic.

> should we try and print `expr tt` as `expr`?

We could actually do this for all defaulted arguments.

leodemoura:
> ...but not in the C++ way :) . Still gets rid of all to/of_raw_expr (one could always use unchecked_cast if they really need them).

@kha I think this is a nice change, but it conflicts with https://github.com/jroesch/lean/commit/bb339d55a9876fe1c6af6eea619d6dcc74fe96b0

One option is to add APIs for "decoding" specific macro applications.
Example: we add
```lean
is_equations   : expr -> optional (list (expr × expr))
```
Then, @jroesch would use them to implement his linter.

Kha:
Right, I forgot to state that I've only "upgraded" definitions where necessary for now. And yes, even for `opt_param`s dependent on previous arguments, deciding when to elide them shouldn't be too hard (so it would actually be more accurate than pp handling of implicit parameters).

gebner:
> One option is to add APIs for "decoding" specific macro applications.

I see another option: make `expr` an indexed family, and add @jroesch's `equations` cases, etc., as additional cases to `expr`.
```lean
| equations : list (expr ff × expr ff) -> expr ff
```

leodemoura:
> I see another option: make expr an indexed family, and add @jroesch's equations cases, etc., as additional cases to expr.
> | equations : list (expr ff × expr ff) -> expr ff

@gebner We have considered this option. It is not feasible.

1- The pattern matching for expr is already super expensive. This extra constructor would add a lot of extra complexity. Note that it requires the inductive compiler for compiling the nested inductive datatype. 

2- There is a bootstrapping issue. The inductive compiler depends on the tactic framework that has not been defined when we define `expr`.

Kha:
@leodemoura 
> @Kha I think this is a nice change, but it conflicts with jroesch/lean@bb339d5

Oh, I didn't know that was already implemented.

Ideally, I would hope that we could extend the existing `macro_def` type so that we have to export just one declaration per macro for both constructing and destructing it - something like

```lean

meta constant macro_kind (α : Type) : Type

meta structure macro_def : Type :=
{α    : Type}
(kind : macro_kind α)
(data : α)

meta constant macro_kind.has_decidable_eq (α : Type) : decidable_eq (macro_kind α)
attribute [instance] macro_kind.has_decidable_eq

meta inductive expr
...
| macro       : macro_def → list expr → expr

meta constant choice_macro : macro_kind unit
meta constant pexpr_quote_macro : macro_kind expr
attribute [pattern] pexpr_quote_macro

meta def is_pexpr_quote_macro : expr → option expr
| (expr.macro ⟨pexpr_quote_macro, e⟩ []) := some e
| _ := none
```

Now, there are just two missing pieces with this:
* We would have to teach the equation compiler to accept and use value transitions on `macro_kind`. We should probably always use value transitions on closed, dec_eq expressions in meta code?
* `macro_def` is actually in `Type 2` now... I don't feel like suggesting foregoing both positivity and universe checks in meta code on the same day :smile: 

So yes, I'm completely fine with an API like

> `is_equations   : expr -> optional (list (expr × expr))`

leodemoura:
> macro_def is actually in Type 2 now... I don't feel like suggesting foregoing both positivity and universe checks in meta code on the same day 😄

@kha The check is not just in the kernel. We would have to change `type_context` too, and any other place that creates/solves universe constraints.

Kha:
Okay, but removing only the consistency check on the inductive result
universe shouldn't impact other systems, should it? But again, I'm not
really advocating any such change.

leodemoura:
>  We should probably always use value transitions on closed, dec_eq expressions in meta code?

There are issues here too. We can avoid them by doing this only for meta definitions.
For regular definitions, we have the following issues:
1- The automatically generated equational lemmas are not refl anymore.
2- Dependency. The value-transition can only be used if the type of other arguments do not depend on the value. Sometimes, the dependency is indirect. For example, when we are using structural recursion (`brec_on`), the inductive hypothesis may depend on the value.

leodemoura:
> Okay, but removing only the consistency check on the inductive result
universe shouldn't impact other systems, should it? But again, I'm not
really advocating any such change.

I see. You want to keep `expr` at `Type 0`.

leodemoura:
Let me double check one thing. The main motivation for this last change is: we would be able to pattern match on the `macro_kind`, right?
I think this is a very small improvement. 
Very few parts of the code base need to pattern match on macros. Moreover, for each builtin meta constant such as 
```lean
meta constant choice_macro : macro_kind unit
```
We can have a predicate.
```lean
meta constant : is_choice_macro : macro_def -> bool
```

leodemoura:
Just to make it clear my position. My counter proposal is:
- Merge this commit
- Add predicates/destructors for macros
- Rewrite jroesch/lean@bb339d5 to use these predicates

Kha:
Sounds completely reasonable. I was mostly just speculating about how in theory we could implement it with just one meta constant per macro.

I will change the pretty printer to omit default parameters, though I probably won't find time for it tomorrow.

Kha:
@leodemoura It's not clear to me yet where to insert eliding default parameters in the `pp, pp_app, pp_child, ...` cycle. I assume we'd have to break up that cycle somewhere since we want to elide trailing arguments only.

leodemoura:
> @leodemoura It's not clear to me yet where to insert eliding default parameters in the pp, pp_app, pp_child, ... cycle. I assume we'd have to break up that cycle somewhere since we want to elide trailing arguments only.

I think we have to handle it at `pp_app`. We should add a case for `has_opt_param`.
We will need to use `get_app_args` to extract all arguments, and use a for-loop instead of recursion.

Kha:
@leodemoura I was completely overwhelmed by all the ways `pp_app` can be reentrant, so for now I've restricted myself to a subset of default parameters (that includes the one in `expr`):
* only the last parameter, by ensuring that the binding body type cannot possibly be another Pi (there could still be `coe_fn`s, but I don't think one should mix these with `opt_param` in any case),
* only closed, structurally equal default values.

src/frontends/lean/pp.cpp
```diff
@@ -447,6 +447,27 @@ bool pretty_fn::is_implicit(expr const & f) {
     }
 }
 
+bool pretty_fn::is_default_arg_app(expr const & e) {
+    if (m_implict || m_preterm)
+        return false; // showing default arguments
+    if (!closed(app_fn(e))) {
+        // the Lean type checker assumes expressions are closed.
+        return false;
```
leodemoura, line 455: This is fine. We rarely try pretty print expressions containing de-Bruijn indices.

src/frontends/lean/pp.cpp
```diff
@@ -447,6 +447,27 @@ bool pretty_fn::is_implicit(expr const & f) {
...
+    }
+    try {
+        expr t = m_ctx.relaxed_whnf(m_ctx.infer(app_fn(e)));
+        if (is_pi(t)) {
+            expr arg_type = binding_domain(t);
+            t = binding_body(t);
+            if (!is_pi(t) && !is_var(t) && is_app_of(arg_type, get_opt_param_name(), 2)) {
+                expr defval = app_arg(arg_type);
+                return closed(defval) && defval == app_arg(e);
+            }
```
leodemoura, line 465: We can use `m_ctx.is_def_eq(defval, app_arg(e))` instead of `defval == app_arg(e)`.

leodemoura:
@Kha If the last two arguments are `opt_param`, then only the last one is going to be omitted, right? 
```lean
def f (a : nat) (b : bool := tt) (c : nat := 10) := a + c
#check f 10
-- (f 10 tt) : nat
```
One possible way to address this problem is to have an extra `optional<expr>parent` argument in the `pp*` methods. Then, we can avoid the test: binding body type cannot possibly be another Pi.

> only closed, structurally equal default values.

closed is not a big deal. We rarely pretty print expressions containing unbounded de-Bruijn variables.

If the expressions are close, we can use `is_def_eq` instead of structural equality.

leodemoura:
@kha I merged this PR as-is. We can address pending issues using another PR.

Kha:
@leodemoura 
> If the last two arguments are opt_param, then only the last one is going to be omitted, right?

Yes. The `parent` parameter would be a better solution, but I shied at having to transport it through about half a dozen methods.

PR #1586, opened 2017-05-17
Fixes

Kha:


src/frontends/lean/pp.cpp
```diff
@@ -665,30 +665,36 @@ auto pretty_fn::pp_const(expr const & e, optional<unsigned> const & num_ref_univ
...
+        if (auto n1 = hidden_to_user_name(m_env, short_n))
+            short_n = *n1;
+    }
+    if (m_ctx.has_local_pp_name(short_n.get_root())) {
+        if (m_ctx.has_local_pp_name(n.get_root())) {
+            n = get_root_tk() + n;
+        }
+    } else {
+        n = short_n;
     }
```
Kha, line 697: @leodemoura This whole logic looks a bit fragile, but I _think_ it's correct now.

src/frontends/lean/structure_cmd.cpp
```diff
@@ -654,7 +654,8 @@ struct structure_cmd_fn {
                     expr base_obj = *mk_base_projections(m_env, parent_name, base_S_name, field);
                     level_param_names lparams; unsigned nparams; inductive::intro_rule intro;
                     std::tie(lparams, nparams, intro) = get_parent_info(base_S_name);
-                    expr type = instantiate_univ_params(m_p.env().get(full_fname).get_type(), lparams, const_levels(parent_fn));
+                    levels meta_ls = map2<level>(lparams, [&](name const &) { return m_ctx.mk_univ_metavar_decl(); });
+                    expr type = instantiate_univ_params(m_p.env().get(full_fname).get_type(), lparams, meta_ls);
```
Kha, line 658: @leodemoura I should probably ask whether this whole hack (see also comment above it) is the correct way to go. Basically, all I need is the type `a` to use in a telescope like `Pi (p : applicative), let map : a := functor.map (applicative.to_functor p) in ...`. If I just use a placeholder instead, elaboration fails with some delayed abstractions.

leodemoura: @kha We should discuss this issue next week during our dev meeting. It is not clear to me what the problem is.
Before you refactored this module, I would create a fake `Pi` (or `lambda`, I don't remember), and elaborate everything in as a single call to the elaborator. I'm assuming this approach does not work (i.e., it fails with some delayed abstractions as your reported above). Is this assumption correct?

PR #1588, opened 2017-05-18
Various segfaults

gebner:
I stumbled upon these segfaults when playing around with `ext_simplify_core`.  @leodemoura I hope the fix for the temporary metavariables is at the wrong place.  I can try to reproduce the issue if we should fix it somewhere else.


gebner:
Our CI infrastructure is hosed.  Everything fails, but they're all unrelated failures.

PR #1591, opened 2017-05-19
Recover from all the errors

gebner:
Short demo:
```lean
/- unknown identifiers -/

def f1 : ℕ → ℕ
| 42 := 9000
| arg := ag

#eval f1 42 -- OK (prints 9000)

/- incomplete structure instances -/

def f2 : ℕ × ℕ := { fst := 9000, sn}

#reduce f2.fst -- OK (prints 9000)

/- incomplete if-then-else -/

def f3 (x : ℕ) : ℕ :=
(if x ≥ 42 then 9000)
                 -- ^ missing else reported here

#eval f3 42 -- OK (prints 9000)
```

Particularly the recovery from unknown identifiers and incomplete structure instances is important to me.  Right now unknown identifiers prevent all other errors from being shown, so you spend lots of time fixing the unknown identifiers only to find out that there are lots of other problems as well--I'd like to see all errors as soon as possible.
At the moment we only report elaboration errors for structure instances if we can completely and correctly parse them.  When I type a monad instance like `{ pure := ... }`, I don't want to guess all the required fields (the error message for a missing field is only shown if the syntax is correct..).

While this is just very basic recovery, it was much less work than I'd expected.  Interestingly, the most difficult part is `parser::parse_notation`.  And the hard part is just a corner-case: when `parse_notation` did not consume any token, and did not find an accepting state either.
 * If it just returns with `sorry`, then `parse_expr` goes into an endless loop, since the termination criterion is `rbp < curr_lbp()` (and doesn't depend on the success of `parse_notation`).
 * If it consumes a token to ensure termination, then the interactive `qexpr` parser behaves badly.  In the parser monad that we expose to the VM, we expect `qexpr?` to work.  Think of for example `by simp []`.  If the scanner is at the `]` token, and invoke the `qexpr` parser, then it should fail, and not recover by consuming `]`.

My current solution is to add in extra termination checks in all while loops that check whether we consumed any input.  This works pretty well, only `t12.lean` shows unexpected errors.

~One particular application does not work yet: `by { simp, simmp }`~
~While we can "successfully" elaborate this tactic, there is still an unassigned metavariable left at the end. 
 Executing the tactic then fails as soon as we add it to the environment (which we need to do in order to compile it to VM code).~

In preparation of the upcoming error message decimation, I added an extra boolean flag to sorrys.  This flag distinguishes "synthetic" sorrys introduced during error recovery from the ones the user entered.  I had hoped that you could just suppress all error messages where such a sorry is in the target of the goal, but unfortunately this approach does not help much since this special situation occurs much more rarely than I'd expected.

I think I fixed all endless loops, but there might still be some hidden somewhere.


gebner:
We can now recover from syntax errors in tactics as well:
```lean
lemma f5 (x : ℕ) : x+1 = 1+x :=
by {
    simp,
    trace_state, -- OK (no goals)
    simmp,***,
    simp a,,,
```

tests/lean/interactive/field_info.lean.expected.out
```diff
@@ -1,6 +1,7 @@
 {"msgs":[{"caption":"","file_name":"f","pos_col":40,"pos_line":13,"severity":"error","text":"invalid expression, `)` expected"}],"response":"all_messages"}
+{"msgs":[{"caption":"","file_name":"f","pos_col":40,"pos_line":13,"severity":"error","text":"invalid expression, `)` expected"},{"caption":"","file_name":"f","pos_col":0,"pos_line":12,"severity":"error","text":"don't know how to synthesize placeholder\ncontext:\n⊢ Type ?"}],"response":"all_messages"}
 {"message":"file invalidated","response":"ok","seq_num":0}
 {"record":{"full-id":"nat.to_string","source":,"type":"ℕ → string"},"response":"ok","seq_num":2}
 {"record":{"full-id":"list.all","source":,"type":"Π {α : Type}, list α → (α → bool) → bool"},"response":"ok","seq_num":4}
 {"record":{"full-id":"list.all","source":,"type":"Π {α : Type}, list α → (α → bool) → bool"},"response":"ok","seq_num":9}
-{"record":{"full-id":"list.all","source":,"type":"Π {α : Type u}, list α → (α → bool) → bool"},"response":"ok","seq_num":13}
+{"record":{"full-id":"list.all","source":,"type":"Π {α : Type _elab_u.16.225}, list α → (α → bool) → bool"},"response":"ok","seq_num":13}
```
leodemoura: This looks bad. Why is the auxiliary universe used by the elaborator leaking here?

gebner: They're already leaking in quite a few places before this PR.  For example in proofs:
```lean
lemma foo (α) (xs : list α) : xs ++ [] = xs :=
begin trace_state, simp end
       -- ^ leak visible here
```
And also here when you hover over α:
```lean
def bar (α) (xs : list α) : list α := xs
```

As for why, I believe we use the auxiliary universes for implicit levels during elaboration.  Anything you print (or save in the info_manager) during elaboration will mention the auxiliary names.  We don't even substitute the universe metavariables at the moment.  I'll take a closer look at what we can do tomorrow.

gebner: The problem was that finalization replaced the _elab_u parameters with nicer ones, but we never propagated this information anywhere else.

I removed the source of the leaks.  We no longer use _elab_u parameters, but directly assign the nice names to the universe meta-variables.  This approach is not only much simpler, but also transparently works with the info_manager and anything else that looks at the universe levels during elaboration.

leodemoura: @gebner This is a good change. To be on the safe side, I tried to investigate why we had `g_level_prefix` and the `_elab_u` parameters. I think I found the reason. In an older version of the elaborator, it would convert unassigned universe metavariables into parameters even for nested elaboration invocations. To make sure there were no collisions between different nested calls, the temporary `_elab_u` names were used, and a final cleaning step performed at the top level:  https://github.com/leanprover/lean/blob/2c8e484aa340b46ef3581796d9ea766e5a5b19fd/src/frontends/lean/elaborator.cpp#L1549-L1569
We now only convert universe metavars into parameters at the top level. So, you are right, all this extra logic is not needed anymore.

PR #1599, opened 2017-05-23
feat(leanpkg): add initial support for test command

jroesch:
Making a few tweaks to `leanpkg` added a simple test command which reuses the LEAN_PATH and invokes `--make` in the test directory. You can see an example on the `smt2_interface` repository.

### DIFF
--- a/leanpkg/leanpkg/main.lean
+++ b/leanpkg/leanpkg/main.lean
@@ -45,10 +45,22 @@ assg ← solve_deps d,
 path_file_cnts ← mk_path_file <$> construct_path assg,
 write_file "leanpkg.path" path_file_cnts
 
-def make : io unit :=
-exec_cmd { cmd := "lean", args := ["--make"], env := [("LEAN_PATH", none)] }
+
+def make : io unit := do
+manifest ← read_manifest,
+let src_path := manifest.path.get_or_else "src",
+exec_cmd {
+  cmd := "lean",
+  args := ["--make", src_path],
+  env := [("LEAN_PATH", none)]
+}
 
 def build := configure >> make
+
+def make_test : io unit :=
+exec_cmd { cmd := "lean", args := ["--make", "test"], env := [("LEAN_PATH", none)] }
+
+def test := configure >> make_test
 
 def init_gitignore_contents :=
 "*.olean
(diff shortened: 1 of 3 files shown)


jroesch:
Leo and I talked about a simple hack to implement a testing framework which reports the number of failures and successes. We can simply generate `@[test] def _test_framework.fresh_name_XXX : bool := (tt | ff)` each indicating a success or failure of the test, allowing us to generate a nice report summary. Going to work up a demo of that this afternoon.

jroesch:
@Kha @gebner do you guys have more thoughts on how we should do this? trying to move things out of the standard library and being able to test packages are important to this.

gebner:
I think the general approach is good.  It is simple and pretty future-proof in that we do not commit to support any testing framework, but only promise to execute all files in a given directory.  IIRC, Haskell has the same convention.

I don't have strong opinions on directory layout, but if we start having `test` directories etc., maybe we should really push to have `src` directories (and enable them by default).

> We can simply generate @[test] def _test_framework.fresh_name_XXX : bool := (tt | ff)

Another solution that is pretty popular is to annotate your tests with `@[test]` and then have the test runner look up all definitions in the environment that have the `test` attribute and execute them.  Then we don't need to store anything in the environment.

library/init/meta/tactic.lean
```diff
@@ -86,6 +86,13 @@ meta def fail_if_success {α : Type u} (t : tactic α) : tactic unit :=
  (λ a s, mk_exception "fail_if_success combinator failed, given tactic succeeded" none s)
  (λ e ref s', success () s)
 
+meta def success_if_fail {α : Type u} (t : tactic α) : tactic unit :=
```
gebner, line 89: This change is completely unrelated, right?

leanpkg/leanpkg/main.lean
```diff
@@ -43,13 +43,26 @@ d ← read_manifest,
 io.put_str_ln $ "configuring " ++ d.name ++ " " ++ d.version,
 assg ← solve_deps d,
 path_file_cnts ← mk_path_file <$> construct_path assg,
+io.put_str_ln path_file_cnts,
```
gebner: This is just left-over debugging code, right?  If you actually want to print it, I think we should output it with some kind of description.

library/init/meta/tactic.lean
jroesch: Yes, this is useful for testing tactic failures, and I am using this branch to update & test smt2, it seemed easy to add while making testing related changes.

leanpkg/leanpkg/main.lean
jroesch: I removed this, probably just forgot to push the cleanup commit.

Issue #1402, opened 2017-02-27
RFC: coercion resolution


leodemoura:
We currently use type class resolution to implement coercion resolution.
The type classes are defined in the following file
https://github.com/leanprover/lean/blob/master/library/init/coe.lean
The resolution process is described in Section 10.6 at
https://leanprover.github.io/theorem_proving_in_lean/

On top of that, we have the following two hacks/extensions.

1- A coercion from `Prop` to `bool` for decidable propositions.
This is harded coded in the elaborator. So, it doesn't participate
in the type class resolution process.

We cannot define this coercion as an instance of `has_coe`
because the coercion from `Prop` to `bool` depends on the argument `p : Prop`.
In the current approach, we cannot redefine the has_coe type classes, and allow dependencies
on the input argument. Reason: the input argument usually contains several metavariables
at coercion resolution time.

The current hack addresses this limitation by just checking whether the types are `Prop` and `bool`,
and putting the decidable instance on the todo set for type class resolution.

2- By default coercion resolution is only fired if the types are know
when the type mismatch was detected. However, there is an exception for
monads. Given, an unification failure of the form
```
              M a =?= N ?m
```
where `M` and `N` are monads, we assign `?m := a` and then try to find
a coercion from `M a` to `N a`.
This hack is used all the time when we are lifting tactics (from `tactic` to `smt_tactic`, and `tactic` to `solver`).

We can try to avoid this hack by adding `has_coe (M a) (N ?m)` to the type class resolution todo set,
and assume the instance can eventually be synthesized. One potential drawback is weird error messages
about failures to synthesize `has_coe` instances instead of a clear "type mismatch error".

We also have a list of pending features requested by users

1- Coercions `C` from `T` to function space are *only* used when the term of type
   `T` is used as a function. Example:
```
           f : T
           (f a)  -- Coercion C will be considered here
```
   Users also want us to consider this kind of coercion when elaborating higher-order functions.
   Example:
```
           g : (nat -> nat) -> nat
           f : T
           C : has_coe_to_fun T

           g f   -- Coercion C is used here
```

   This one is relatively easy to implement.

2- Backtrack when considering coercions from a type `T` to the function space.
   The user may have defined multiple coercions from `T` to the function space.
   They want the lean elaborator to trie all of them. Example.
```
           C1 : has_coe_to_fun T -- to nat -> nat
           C2 : has_coe_to_fun T -- to bool -> bool -> bool
           f  : T

           check f 10    -- uses C1
[...]

PR #1612, opened 2017-05-28
feat(data/pnat): positive natural numbers

digama0:
This is just extracting the definition from `hash_map` (which is not the best place for it) and adding a few basic definitions. One question: is it safe to have cyclic coercions as in this case `ℕ -> ℕ+`, `ℕ+ -> ℕ`? Does typeclass inference notice and avoid the cycle?

### DIFF
--- a/library/data/hash_map.lean
+++ b/library/data/hash_map.lean
@@ -3,12 +3,9 @@ Copyright (c) 2017 Microsoft Corporation. All rights reserved.
 Released under Apache 2.0 license as described in the file LICENSE.
 Authors: Leonardo de Moura, Mario Carneiro
 -/
-import data.list.set data.list.comb init.data.array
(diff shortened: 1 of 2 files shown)


leodemoura:
> One question: is it safe to have cyclic coercions as in this case ℕ -> ℕ+, ℕ+ -> ℕ? Does typeclass inference notice and avoid the cycle?

Cyclic dependencies may produce performance problems for problems that do not have a solution.
In principle, this kind of problem may loop if there is cyclic dependency, but we use a threshold on the search depth to prevent it. For example, suppose we are trying to find a coercion from `nat` to `foo`. Then type class resolution will keep trying the coercions `ℕ -> ℕ+, ℕ+ -> ℕ` until the threshold is reached.

We may add better cycle detection after we implement:
https://github.com/leanprover/lean/issues/1402

Issue #1175, opened 2016-11-02
RFC : Strings


jroesch:
This is a placeholder issue which I will expand will a full RFC post-PLDI. 

A 100 mile view: getting strings right is *very* hard, we should try not repeat the sins of the past and build a good story around interoperability for both byte strings (array of `u8`s) and unicode strings.

Specifically we should look at how the String ecosystem has developed in other languages, good examples to learn from are Haskell, and Rust. In both communities there has developed a family of String types, we should make doing the right thing easy, and support *all* common use cases. It is hard to build *real* systems without good support for Strings. 

In general I want to have a story for *all* native types, but this RFC will focus on just Strings.

excerpt from a Slack conversation:

I think its technically an extension

[3:54]  
but gets desugared to a method `fromString : String -> R`

[3:54]  
where `R` is the string type

[3:54]  
https://hackage.haskell.org/package/base-4.9.0.0/docs/Data-String.html

[3:54]  
this allows us to keep simple strings while we experiment with different alternatives

[3:55]  
I like Rust’s structuring better

[3:55]  
there is just `String`, `CString`, `[u8]` and `OsString`

[3:55]  
with appropriate conversions for each

[3:55]  
String is like Haskell’s text and *must* be valid utf-8

[3:56]  
`CString` is a null terminated version of `[u8]`

[3:56]  
and `OsString` is for interfacing with native code, that has platform specific encodings

[3:56]  
they internally use a string encoding called `wtf-8`

[3:56]  
https://simonsapin.github.io/wtf-8/

[3:57]  
to deal with Windows & Unix compatibility

gebner:
> In general I want to have a story for all native types, but this RFC will focus on just Strings.

You haven't defined "native", but I strongly hope that `string` does not become an uninterpreted meta-constant.  It would be a shame if you could not reason about anything involving strings.

From this point of view, Haskell-like strings are probably easier to accomodate, as they are just 

``` lean
@[irreducible] def string = list code_point
def code_point = fin 0x10ffff
```

We can and probably should hide the fact that it is implemented as a list, so that people don't try to use `list.length` or other incorrect operations.  This also makes it easier to replace the list by an alternative implementation later on.

> byte strings (array of u8s)

Should we store byte strings as byte arrays in memory?  It would make cons expensive, but at least Haskell's Data.Text gets away with having O(n) cons.

leodemoura:
The commit above tries to "protect" the string implementation.
[...]

PR #1613, opened 2017-05-28
feat(leanpkg): support (deterministic) timeout value in leanpkg.toml

johoelzl:
Add an option the the `leanpkg.toml` file to set a value of the timeout. When compiling the current value is `0` (i.e. infinity). For compiling `library_dev` it makes more sense to have a smaller value.

### DIFF
--- a/leanpkg/leanpkg/main.lean
+++ b/leanpkg/leanpkg/main.lean
@@ -50,7 +50,8 @@ def make : io unit := do
 manifest ← read_manifest,
 exec_cmd {
   cmd := "lean",
-  args := ["--make"] ++ manifest.effective_path,
+  args := (match manifest.timeout with some t := ["-T", t.to_string] | none := [] end) ++
+    ["--make"] ++ manifest.effective_path,
   env := [("LEAN_PATH", none)]
 }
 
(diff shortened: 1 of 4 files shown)


leanpkg/leanpkg/manifest.lean
```diff
@@ -44,8 +44,9 @@ end dependency
 
 structure manifest :=
 (name : string) (version : string)
-(path : option string)
-(dependencies : list dependency)
+(timeout : option string := none)
```
gebner: `timeout` should probably have type `option nat`

leanpkg/leanpkg/main.lean
```diff
@@ -50,7 +50,8 @@ def make : io unit := do
 manifest ← read_manifest,
 exec_cmd {
   cmd := "lean",
-  args := ["--make"] ++ manifest.effective_path,
+  args := (match manifest.timeout with some t := ["-T", t] | none := [] end) ++
```
gebner: Note: the `timeout` setting is not propagated from dependencies.  I don't think it should, but it is something to keep in mind.

leanpkg/leanpkg/manifest.lean
johoelzl: Okay, I will change this.

leanpkg/leanpkg/main.lean
johoelzl: I also think it should **not** be propagated. It will depend on how proofs are structured and which kind of automation is used. Also packages have multiple dependencies, how should we merge two different timeout values?

johoelzl:
Now timeout is a natural number. I did **not** implement the `_` of TOML. Is it actually used?

johoelzl:
Sorry, don.t merge. There is a problem with the parser...

johoelzl:
It is fixed now. But why is the order of lists and strings is reversed?

gebner:
Yes, strings are reverse lists of characters at the moment.  This confuses me to no end as well, but I suspect the rationale for this representation is that reverse character lists have a fast append operation.

Refactoring strings is on a to-do list, somewhere: #1175

PR #1615, opened 2017-05-29
Wildcards

avigad:
This fixes, extends, and closes #1579.

- adds wildcards for `simp`, `dsimp`, `dunfold`, and `delta` in interactive mode
- adds an `only` flag to suppress inclusion of default theorems in `simp`, `simph`, `simp_intros`, and `dsimp` in interactive mode.
- `mk_simp_set` is no longer private to `interactive` (it is useful for automation in `library_dev`)
- tactics that create or modify a local hypothesis return the corresponding expression: `assert`, `assertv`, `simp_at`, etc.

There are some noninteractive versions of `simp_only`, `simp_all`, etc., in `library_dev`, but let's leave them there for now, because they are mainly useful for writing automation in Lean.


library/init/meta/interactive.lean
```diff
@@ -20,6 +20,18 @@ namespace interactive
...
 @[reducible] meta def parse {α : Type} [has_reflect α] (p : parser α) : Type := α
 
+inductive loc : Type
+| wildcard : loc
+| ns       : list name → loc
+
+def loc.empty := loc.ns []
+
+meta instance : has_reflect loc :=
+fun l, match l with
```
Kha: nit: could also use direct function-level matching

library/init/meta/interactive.lean
```diff
@@ -66,6 +81,7 @@ private meta def parser_desc_aux : expr → tactic (list format)
 | `(ident)  := return ["id"]
 | `(ident_) := return ["id"]
 | `(qexpr) := return ["expr"]
+| `(location) := return ["loc?"]
```
Kha: Did you try the default description? I can see that it may not be quite readable.

library/init/meta/interactive.lean
```diff
@@ -629,12 +662,18 @@ private meta def simp_hyps (cfg : simp_config) : simp_lemmas → list name → t
...
+private meta def simp_core (cfg : simp_config) (no_dflt : bool) (ctx : list expr) (hs : list pexpr) (attr_names : list name) (ids : list name) (locat : loc) : tactic unit :=
+do s ← mk_simp_set no_dflt attr_names hs ids,
    s ← s.append ctx,
-   match loc : _ → tactic unit with
-   | [] := simp_goal cfg s
-   | _  := simp_hyps cfg s loc
+   match locat : _ → tactic unit with
+   | loc.wildcard :=
+     do ls ← local_context,
+        let loc_names := ls.map (fun l, l.local_pp_name),
```
Kha: nit: This feels like taking field notation too far :) . But I don't really mind.

library/init/meta/interactive.lean
```diff
@@ -37,12 +49,15 @@ meta def ident_ : parser name := ident <|> tk "_" *> return `_
 meta def using_ident := (tk "using" *> ident)?
 meta def with_ident_list := (tk "with" *> ident_*) <|> return []
 meta def without_ident_list := (tk "without" *> ident*) <|> return []
-meta def location := (tk "at" *> ident*) <|> return []
+meta def location := (tk "at" *> (tk "*" *> pure loc.wildcard <|> (loc.ns <$> ident*))) <|> return loc.empty
```
Kha: Reminder to myself: We should definitely introduce a `+` combinator and replace most uses of `*` with it in this file.

library/init/meta/interactive.lean
```diff
@@ -655,48 +694,53 @@ It has many variants.
 
 - `simp with attr` simplifies the main goal target using the lemmas tagged with the attribute `[attr]`.
 -/
```
Kha: Doc string should be updated

Kha (approved):

Kha:
Now that I think about it, one may reasonably assume that `simp with,` does the same as `simp only,`, since `simp with a b` does not include the default simpset, either. It doesn't look pretty, but _would_ get us around having to reserve another token.

library/init/meta/interactive.lean
leodemoura: @kha Yes

library/init/meta/interactive.lean
leodemoura: The PR also uses `fun` in a few places instead of the equivalent unicode notation that is used in the rest of the stdlib.

leodemoura:
> Now that I think about it, one may reasonably assume that simp with, does the same as simp only,, since simp with a b does not include the default simpset, either. It doesn't look pretty, but would get us around having to reserve another token.

@kha I agree. We already have too many keywords.  
BTW, some of them are really annoying (e.g., `take`).

library/init/meta/interactive.lean
avigad: The default description complained that it did not know how to display `loc.wildcard`. Can I get that with a `to_string` method? If so, I'll try it and see how it looks.

avigad:
In the previous implementation `simp with a b` added lemmas from `a` and `b` to the default `simp _lemmas`. It sounds like you want to base it on the empty `simp_lemmas` and get rid of `only`. I will do that.

library/init/meta/interactive.lean
Kha: Oh I see, it handles `return` but not `pure`. I pushed a fix, it should work now out of the box.

avigad:
I made all the changes discussed above, and got rid of the 'only' flag. Now we can type
```lean
  simp [h_1, ..., h_n]
```
to simplify using the default simp set and h_1, ..., h_n, and
```lean
  simp [h_1, ..., h_n] with
```
to use *only* `h_1, ..., h_n`.

There is a problem, though. We can use
```lean
  simp [h_1, ..., h_n] with foo
```
to simplify using `h_1, ..., h_n` and the user attribute foo, or
```lean
  simp [h_1, ..., h_n] with foo bar
```
to simplify with user attributes `foo` and `bar`. But these do not include the default `simp` lemmas. Currently `simp` is not recognized as a user attribute, so we cannot write
```lean
  simp [h_1, ..., h_n] with simp foo
```
to simplify with the default simp set + foo. On the other hand, if we include `simp` lemmas by default, there is no way to exclude them.

Options:

- Make `simp` a possible attribute, so we can write
```lean
  simp [h_1, ..., h_n] with simp foo
```
  We would have to declare `simp` as an identifier at the top level or hack the parser. This seems bad.
  
- Go back to the `only` option, so `with` includes the default simp rules unless `only` is present.

- Use alterantive versions, `simp_only`, `simph_only`, `simp_intros_only`, etc., rather
  the a flag.

- Add a `use_dflt := tt` field to `simp_config`.

leodemoura:
> We would have to declare simp as an identifier at the top level or hack the parser. This seems bad.

Yes, it would be bad.

> Go back to the only option, so with includes the default simp rules unless only is present.

This is reasonable. I think it is the best option.

Kha:
> Currently simp is not recognized as a user attribute

We have `simp with foo default` for that. Undocumented, of course :) .

leodemoura:
> We have simp with foo default for that. Undocumented, of course :) .

@Kha Yes, you are correct.
So, are you suggesting we should not add  the `only` keyword?
@avigad wrote a nice summary of the pros and cons. To be honest, I found the syntax `simp with` a little bit awkward. I think it would be ok if we had brackets (i.e., list notation).
That is, we would write `simp with []`, `simp with [default, foo]`, etc.
What do you think?

avigad:
I think the solution with `default` is pretty good. We can say:
```lean
  simp [h1, ..., hn] with default, foo
  simp [h1, ..., hn] with foo
  simp [h1, ..., hn]
  simp [h1, ..., hn] with
```
and the semantics is easy to document and explain. 

With brackets:
```lean
  simp [h1, ..., hn] with [default, foo]
  simp [h1, ..., hn] with [foo]
  simp [h1, ..., hn]
  simp [h1, ..., hn] with []
```

With `only`:
```lean
  simp [h1, ..., hn] with foo
  simp only [h1, ..., hn] with foo
  simp [h1, ..., hn]
  simp only [h1, ..., hn]
```
All are o.k. by me, and I have no strong preference.

leodemoura:
> All are o.k. by me, and I have no strong preference.

@avigad For me, the only weird one is `simp [h1, ..., hn] with`

digama0:
I vote for "simp only [h1,...,hn] with foo". It seems like a reasonable way
to recognize that "simp" is a special attribute set, and also reads fine.

Kha:
I don't have a strong preference towards either of `only` or `with []`.

@avigad 
> We would have to declare simp as an identifier at the top level or hack the parser. This seems bad.

`simp` is not a keyword. All we would have to change is the literal in [this line](https://github.com/leanprover/lean/blob/81e97de463073c7511f01875075651259eb8c339/library/init/meta/simp_tactic.lean#L402).
I believe that would be a reasonable change independent of the surrounding issue.

avigad:
O.k., I restored the `only` flag, added the default `[simp]` lemmas for 'with' by default, and updated the docstring. I rebased on the current version of Lean and tested, so we should be good to go.

In response to @Kha's observation, I tried getting rid of `default` in `get_user_simp_lemmas`, but this broke a test (`run/converter.lean`). It seems that at the low level, there are situations where we still need `default` (or something) to refer to the standard `[simp]` attribute. It doesn't interfere with the `with` option at all, so I left it alone.

PR #1617, opened 2017-05-30
User-defined notation parsers

Kha:
Allow user-defined notation that uses `lean.parser` for parsing its parameters and computes a pre-term in the tactic monad. The notation is activated by a single `tk` parser, which may optionally be preceded by a `qexpr` parser for postfix/infix notations.

[Examples](https://github.com/leanprover/lean/commit/674cec026989f87c7ff5f4fbd7297c0603f1d52c#diff-4b4ca2d079ab4a76b323a473ec3f6eaf): the mandatory `unquote!` macro, a `format!` macro, and a non-creative infix notation example. More suggestions welcome.

### DIFF
--- a/library/init/meta/interactive.lean
+++ b/library/init/meta/interactive.lean
@@ -6,133 +6,13 @@ Authors: Leonardo de Moura
 prelude
 import init.meta.tactic init.meta.rewrite_tactic init.meta.simp_tactic
 import init.meta.smt.congruence_closure init.category.combinators
-import init.meta.lean.parser init.meta.has_reflect
+import init.meta.interactive_base
 
 open lean
 open lean.parser
 
 local postfix `?`:9001 := optional
 local postfix *:9001 := many
-
-namespace interactive
-/-- (parse p) as the parameter type of an interactive tactic will instruct the Lean parser
-    to run `p` when parsing the parameter and to pass the parsed value as an argument
-    to the tactic. -/
-@[reducible] meta def parse {α : Type} [has_reflect α] (p : parser α) : Type := α
-
-inductive loc : Type
-| wildcard : loc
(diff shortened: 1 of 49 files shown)


Kha:
I fear I might give @leodemoura a heart attack with [this](https://github.com/leanprover/lean/pull/1617/commits/8778e9fffbe51d07e5564df706ae384cd7a648ed) commit, but I hope the [results](https://github.com/leanprover/lean/pull/1617/commits/b835e6628432c2a93db5bee366a008cbe0c4a5dd#diff-082c6edbb0e7c52dfc61ce16dabc9bcbR50) justify the means :) .

library/init/meta/interactive.lean
```diff
@@ -133,813 +132,3 @@ meta def param_desc : expr → tactic format
   then return $ to_fmt "{ tactic }"
   else paren <$> pp e
 end interactive
-
```
Kha: Is everyone fine with this refactoring to reduce dependencies? If yes, I'd recommend merging the PR as-is to reduce merge conflicts. I will improve error handling in a later PR.

leodemoura: I am fine with reducing dependencies, but a new directory `tactic` with a single `interactive.lean` file looks strange. I think `interactive_base.lean` and `interactive.lean` in the same directory (i.e., `meta`) looks better. 

BTW, I'm not against mirroring the namespace hierarchy in the directory structure. However, if you do it, we should be consistent. We should also do it in a different PR. What do you think?

Kha: We could move the `*_tactic(s).lean` files into the directory. However, your suggestion is probably better for the time being since it avoids conflicts with other open PRs. I will do the change.

PR #1625, opened 2017-05-31
Bitwise ops

digama0:
Properties of the bitwise operations

### DIFF
--- a/library/init/data/nat/bitwise.lean
+++ b/library/init/data/nat/bitwise.lean
@@ -23,30 +23,187 @@ namespace nat
 
   def test_bit (m n : ℕ) : bool := bodd (shiftr m n)
 
-  def binary_rec {α : Type u} (f : bool → ℕ → α → α) (z : α) : ℕ → α
-  | n := if n0 : n = 0 then z else let n' := shiftr n 1 in
+  def bit (b : bool) : ℕ → ℕ := cond b bit1 bit0
+
+  lemma bit0_val (n : nat) : bit0 n = 2 * n := (two_mul _).symm
+
+  lemma bit1_val (n : nat) : bit1 n = 2 * n + 1 := congr_arg succ (bit0_val _)
+
+  lemma bit_val (b n) : bit b n = 2 * n + cond b 1 0 :=
+  by { cases b, apply bit0_val, apply bit1_val }
+
+  lemma mod_two_of_bodd (n : nat) : n % 2 = cond (bodd n) 1 0 :=
+  match by apply_instance : ∀ d, n % 2 = cond (@to_bool (n % 2 = 1) d) 1 0 with
+  | is_true  h := h
+  | is_false h := (mod_two_eq_zero_or_one _).resolve_right h
+  end
+
+  lemma bit_decomp (n : nat) : bit (bodd n) (shiftr n 1) = n :=
+  (bit_val _ _).trans $ (add_comm _ _).trans $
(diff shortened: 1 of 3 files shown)


library/init/data/nat/bitwise.lean
```diff
@@ -23,30 +23,187 @@ namespace nat
...
   binary_rec
     (λa m Ia, binary_rec
       (λb n _, bit (f a b) (Ia n))
       (cond (f tt ff) (bit a m) 0))
-    (λb, cond (f ff tt) b 0)
+    (λn, cond (f ff tt) n 0)
 
   def lor   : ℕ → ℕ → ℕ := bitwise bor
   def land  : ℕ → ℕ → ℕ := bitwise band
   def ldiff : ℕ → ℕ → ℕ := bitwise (λ a b, a && bnot b)
```
leodemoura, line 116: Should we declare type classes and notation for bitwise operations?
In C, we use `&` (for bitwise and) and `|` (for bitwise or).

library/init/data/nat/bitwise.lean
```diff
@@ -23,30 +23,187 @@ namespace nat
...
+      (λ b n _, bit (f ff b) (cond (f ff tt) n 0))
+      (cond (f tt ff) (bit ff 0) 0) =
+    λ (n : ℕ), cond (f ff tt) n 0 :=
+  begin
+    apply funext, intro n,
+    apply bit_cases_on n, intros b n, rw [binary_rec_eq],
+    { cases b; try {rw h}; ginduction f ff tt with fft; dsimp [cond]; refl },
+    { rw [h, show cond (f ff tt) 0 0 = 0, by cases f ff tt; refl,
+             show cond (f tt ff) (bit ff 0) 0 = 0, by cases f tt ff; refl]; refl }
+  end
```
leodemoura, line 148: It seems you are quite fond of the `rw`/nested `show` idiom ;)
I was shocked the first time I saw you using it. It is growing on me.

library/init/data/nat/bitwise.lean
johoelzl: We could also use the lattice operators inf (⊓) and sup (⊔). They are in the theory `algebra.lattice.basic` in `library_dev`

library/init/data/nat/bitwise.lean
digama0: Personally, I would prefer type ascriptions, but they never seem to work.

library/init/data/nat/bitwise.lean
digama0: We could also use `&&` and `||` as well - unlike C we are well-typed enough to distinguish binary and natural logical ops.

As for using lattice operators, I think it wouldn't be bad to have them in addition, but like the lattice of sets I think the original notation using `&` or `&&` takes precedence over the generic ⊓⊔ symbols.

digama0: I recommend against `|` because this is used in pattern matching and so on - at least I assume that's why subtype changed to `{x // p x}` from lean 2 `{x | p x}`.

library/init/data/nat/bitwise.lean
leodemoura: I think there is a misunderstanding here. They may not do what you want, but they do what they are supposed to do.
The type ascription `(t : T)` makes sure that `t` has a type definitionally equal to `T`.
They do **not** guarantee that `t` will have a type structurally equal to `T`.
So, they are not good for `rw` since it relies on the precise structure of the left hand side of the equation.

PR #1626, opened 2017-05-31
Fix #1623

### LINKED ISSUE #1623: Parameters in structures are duplicated in local context
```
section
parameters {A : Type} (R : list A → Prop)
structure foo (x : list A) : Prop := (bar : R x)
end


type mismatch at application
 R x
term
 x
has type
 list A
but is expected to have type
 list A_1
```
If I replace `x` with `_` to inspect the context, I see:
```
A : Type,
R : list A → Prop,
A : Type,
x : list A
⊢ list A
```
If there are any `variables` floating around, they also show up in the context, even if they are not `include`d. Apparently the whole global context gets pulled in, separate from the context of dependencies of variables like `x`, so that there are two `A` variables coming from one parameter.

Kha:


leodemoura:
The new test (1623.lean) is failing for me.

Kha:
Ah, I didn't know `sort_locals` would reorder only new locals, so now I'm using temporary buffers when elaborating the header. Does this sound right to you?

leodemoura:
@kha I will merge if it passes the tests. I have a few questions regarding `structure_cmd.cpp`. It will be easier to go over the code with you when we meet in Cambridge.

PR #1628, opened 2017-05-31
Caching mechanism

digama0:
This still needs more work on the VM side.

### DIFF
--- a/library/data/hash_map.lean
+++ b/library/data/hash_map.lean
@@ -480,6 +480,15 @@ theorem contains_iff (m : hash_map α β) (a : α) (b : β a) :
   m.contains a ↔ a ∈ m.keys :=
 m.is_valid.contains_aux_iff _ _
 
+theorem entries_empty (hash_fn : α → nat) {n} : entries (@mk_hash_map α _ β hash_fn n) = [] :=
+mk_as_list hash_fn _
+
+theorem find_empty (hash_fn : α → nat) {n} (x) : find (@mk_hash_map α _ β hash_fn n) x = none :=
+by { ginduction find (@mk_hash_map α _ β hash_fn n) x with h y, refl,
+     note := (find_iff _ _ _).1 h,
+     rw entries_empty at this,
+     contradiction }
+
 lemma insert_lemma (hash_fn : α → nat) {n n'}
   {bkts : bucket_array α β n} {sz} (v : valid hash_fn bkts sz) :
   valid hash_fn (bkts.foldl (mk_array _ [] : bucket_array α β n') (reinsert_aux hash_fn)) sz :=
@@ -554,6 +563,15 @@ let n'        : ℕ+ := ⟨n.1 * 2, mul_pos n.2 dec_trivial⟩,
   nbuckets := n',
   buckets  := buckets'',
   is_valid := insert_lemma _ valid' }
+
+theorem insert_hash_fn (m : hash_map α β) (a : α) (b : β a) : (m.insert a b).hash_fn = m.hash_fn :=
+begin
+  cases m, unfold hash_map.insert, simp,
+  cases (by apply_instance : decidable
+    (↑(contains_aux a (bucket_array.read hash_fn buckets a)))) with hc hc; dsimp [dite],
+  cases (by apply_instance : decidable (size + 1 ≤ nbuckets.val)) with hsz hsz; dsimp [ite],
+  all_goals {refl}
+end
 
 theorem mem_insert : Π (m : hash_map α β) (a b a' b'),
   sigma.mk a' b' ∈ (m.insert a b).entries ↔
--- a/library/data/hashed_func.lean
+++ b/library/data/hashed_func.lean
@@ -49,4 +49,7 @@ namespace hashed_func
 
   instance : has_coe_to_fun (hashed_func α β) := ⟨_, hashed_func.eval⟩
 
+  @[simp] theorem eval_val : ∀ hf : hashed_func α β, eval hf = hf.func
+  | ⟨f, hash_fn, c⟩ := funext $ λ a, by rw [c.cache_eq_empty]; refl
+
 end hashed_func
--- a/library/data/hashed_func.lean
+++ b/library/data/hashed_func.lean
@@ -1,0 +1,52 @@
+/-
+Copyright (c) 2017 Microsoft Corporation. All rights reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Author: Mario Carneiro
+-/
+import data.hash_map
+
+universes u v
+
+namespace hashed_func
+  variables {α : Type u} [decidable_eq α] {β : α → Type v}
+
+  def cache_ty (f : Π a, β a) (hash_fn : α → ℕ) :=
+  {m : hash_map α β // ∀ x y, m.find x = some y → f x = y}
+
+  instance (f : Π a, β a) (hash_fn : α → ℕ) : inhabited (cache_ty f hash_fn) :=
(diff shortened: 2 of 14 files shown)


leodemoura:
Is this feature (caching) blocking you? If not, we should postpone it.

digama0:
That's fine, no rush. Just an idea I'm kicking around. I may try my hand at the implementation later, although I'm no expert at the C++ stuff.

digama0:
The implementation went surprisingly without a hitch, although I haven't thought deeply about thread safety and such so perhaps a once-over by a more knowledgeable person wouldn't go awry. Here's a simple test case (not sure how to add):

    #eval let c := cached.mk 2 in c            -- 2
    #eval let c := cached.mk 2 in c.update 3 c -- 3

Note that the printing of cache values uses the meta constant `cached.inspect`, but actually it's not necessary to get "impossible printing" behavior, since the underlying `quot` can also be paradoxically printed. (Paradoxically because a print function does not honor the quotient relation.)

digama0:
The only future feature about this that is not implemented yet is the optional clearing of cache data by the VM whenever it needs to free memory. (Cache objects contain the data necessary for this operation, but I don't know how memory management works for the VM.) The caching itself appears to work as intended.

digama0:
Bump. This PR has been implemented and tested, all that remains is the RFC stuff.

library/data/hash_map.lean
```diff
@@ -555,6 +564,15 @@ let n'        : ℕ+ := ⟨n.1 * 2, mul_pos n.2 dec_trivial⟩,
   buckets  := buckets'',
   is_valid := insert_lemma _ valid' }
 
+theorem insert_hash_fn (m : hash_map α β) (a : α) (b : β a) : (m.insert a b).hash_fn = m.hash_fn :=
+begin
+  cases m, unfold hash_map.insert, simp,
+  cases (by apply_instance : decidable
+    (↑(contains_aux a (bucket_array.read hash_fn buckets a)))) with hc hc; dsimp [dite],
+  cases (by apply_instance : decidable (size + 1 ≤ nbuckets.val)) with hsz hsz; dsimp [ite],
```
Kha, line 572: I think the `by_cases` tactic should work here.

digama0: Yes, it didn't exist when I first wrote that.

library/data/hashed_func.lean
```diff
@@ -0,0 +1,55 @@
...
+      begin
+        by_cases a = x with ax,
+        { revert y, rw -ax, intros y h,
+          rw hash_map.find_insert_eq at h,
+          injection h },
+        { rw hash_map.find_insert_ne _ _ _ _ ax at h,
+          exact H x y h }
+      end⟩ v)
+    (λ⟨m, H⟩, begin
+      unfold default inhabited.default hashed_func.eval._match_1,
```
Kha, line 43: The prefix `hashed_func.eval` should also work

src/library/vm/vm_cached.cpp
```diff
@@ -0,0 +1,67 @@
...
+#define lean_cached_trace(CODE) lean_trace(name({"cached", "update"}), CODE)
+
+namespace lean {
+struct vm_cached : public vm_external {
+    vm_obj m_default;
+    vm_obj m_value;
+    vm_cached(vm_obj const & def, vm_obj const & v):m_default(def), m_value(v) {}
+    virtual ~vm_cached() {}
+    virtual void dealloc() override { this->~vm_cached(); get_vm_allocator().deallocate(sizeof(vm_cached), this); }
+    virtual vm_external * ts_clone(vm_clone_fn const &) override { return new vm_cached(m_default, m_value); }
```
Kha, line 20: This definition is very probably wrong, but I don't know enough about this to fix it either. What is `m_default`, anyway? It looks like you only use `m_value`, in which case you could just ditch the whole class.

gebner: To expand, it is wrong because `m_value` and `m_default` are not cloned.  Both `ts_clone` and `clone` need to do a deep copy.

Kha: On top of that, I believe you need a separate class like with `vm_array_ts_copy` for lifetime management of nested `vm_obj`s.

Kha:
This extension seems to be both relatively simple (apart from the thread safety stuff) and powerful, so I don't have any objections. We should still make sure to add a `@[vm_override]` attribute soon.

src/library/vm/vm_cached.cpp
digama0: `m_default` is the passed in `inhabited A` instance, which may possibly be used in the future if we implement resetting the cache during garbage collection. Note that even if the class is ditched, the pointer indirection is essential for making this work, although a bare pointer could probably do the job just as well.

digama0: Could I ask one of you to jump in and implement the cloning stuff? I'm way out of my depth here.

leodemoura:
> Bump. This PR has been implemented and tested, all that remains is the RFC stuff.

@digama0 I'm not thrilled about this PR. Disclaimer: I didn't review it in detail, but I briefly discussed it with @jroesch which is also not thrilled. First, I'm not sure whether we want this feature in the way it was implemented by this PR. @jroesch and I were planning to add support in the compiler for simulating lazy evaluation. When designing this feature, we have to make sure the design is compatible with the native code generator. We also need to make sure it is compatible with future optimizations we are planning to add. Second, do we need this feature right now? I don't think so. If this feature is blocking a project, then we can prioritize it, and discuss the design. We have many parallel threads, and this feature is not important enough to deserve a new thread right now. We will discuss it in the future, but now it is not a good time. That being said, I'm inclined to close this PR.

digama0:
>  @jroesch and I were planning to add support in the compiler for simulating lazy evaluation.

It's a minor point, but I want to observe that lazy evaluation and caching are two orthogonal issues, although most programming languages implement both or neither. Even with lazy evaluation you can still evaluate a subterm multiple times if it is not memoized.

An explicit control primitive for lazy evaluation already exists: `thunk`. There is no explicit control primitive for caching yet, and if you don't like this implementation then I would like to know how else you would approach the situation.

leodemoura:
> It's a minor point, but I want to observe that lazy evaluation and caching are two orthogonal issues, although most programming languages implement both or neither. Even with lazy evaluation you can still evaluate a subterm multiple times if it is not memoized.
An explicit control primitive for lazy evaluation already exists: thunk. 

We know we have `thunk` (I added it). When I said "lazy evaluation", I was referring to the memoization part. 

> There is no explicit control primitive for caching yet, and if you don't like this implementation then I would like to know how else you would approach the situation.

The main issue is not whether we like it or not.  As I said in the previous comment, we have to make sure the design is compatible with the native code generator. We also need to make sure it is compatible with future optimizations we are planning to add. 

I think this is not a critical feature right now. I don't think anybody needs it right now. As I said in the previous comment, we have many parallel threads, we need to prioritize them.

leodemoura:
I'm going to close this PR. We will discuss this feature in the future.
The plan is to add it when working in the new code generator and native compilation.

digama0:
@leodemoura Sorry to bring up a closed issue, but I wonder if we can merge just the Lean part of this PR. Obviously without the C++ support, `cached` won't actually work (that is, it won't provide the efficiency it promises), but I would still like to have the signatures of `cached` and `memoized` available as an interface from which to design some standard functional data structures. (There is some literature on the necessity of some lazy evaluation construct for achieving lower bounds in some computational tasks, so I don't think we can put this off indefinitely.)

PR #1633, opened 2017-06-02
feat(init/meta/name): make name non-meta

digama0:
Everything in name.lean was implemented except name.cmp, which depends on hashing the pointer to the name, which is not type-safe.

### DIFF
--- a/library/init/data/list/basic.lean
+++ b/library/init/data/list/basic.lean
@@ -318,4 +318,14 @@ infix ` <+: `:50 := is_prefix
 infix ` <:+ `:50 := is_suffix
 infix ` <:+: `:50 := is_infix
 
+instance {α : Type u} [decidable_eq α] : decidable_eq (list α)
+| []     []      := is_true rfl
+| (a::l) []      := is_false (λh, list.no_confusion h)
+| []     (b::l') := is_false (λh, list.no_confusion h)
+| (a::l) (b::l') := if ab : a = b then
+  @dite _ (decidable_eq l l') _
+    (λll, is_true $ congr (congr_arg cons ab) ll)
+    (λll, is_false $ λh, list.no_confusion h (λ_, ll))
+  else is_false $ λh, list.no_confusion h (λn _, ab n)
+
 end list
--- a/library/init/data/list/instances.lean
+++ b/library/init/data/list/instances.lean
@@ -5,7 +5,6 @@ Author: Leonardo de Moura
 -/
 prelude
 import init.data.list.lemmas
-import init.meta.mk_dec_eq_instance
 open list
 
 universes u v
@@ -49,12 +48,6 @@ instance : alternative list :=
 { list.monad with
   failure := @list.nil,
   orelse  := @list.append }
-
-instance {α : Type u} [decidable_eq α] : decidable_eq (list α) :=
-by tactic.mk_dec_eq_instance
-
-instance : decidable_eq string :=
-list.decidable_eq
 
 namespace list
 
--- a/library/init/data/ordering.lean
+++ b/library/init/data/ordering.lean
@@ -6,6 +6,8 @@ Authors: Leonardo de Moura
 prelude
 import init.data.to_string init.data.prod init.data.sum.basic
 
+universes u v
+
 inductive ordering
 | lt | eq | gt
 
@@ -14,6 +16,11 @@ namespace ordering
   | lt := gt
   | eq := eq
   | gt := lt
+
+  def or_else : ordering → thunk ordering → ordering
+  | lt _ := lt
+  | eq f := f ()
+  | gt _ := gt
 
   theorem swap_swap : ∀ (o : ordering), o.swap.swap = o
   | lt := rfl
@@ -26,7 +33,7 @@ open ordering
 instance : has_to_string ordering :=
 has_to_string.mk (λ s, match s with | ordering.lt := "lt" | ordering.eq := "eq" | ordering.gt := "gt" end)
 
-class has_ordering (α : Type) :=
+class has_ordering (α : Type u) :=
 (cmp : α → α → ordering)
 
 def nat.cmp (a b : nat) : ordering :=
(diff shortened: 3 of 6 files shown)


jroesch:
It is not clear to me that this is a good idea *right now*, and in general I am wary of overriding non-meta definitions & constants with implementations in native code. 

For numeric types the case is clear cut as the performance of using model implementations is horrific, and there are clear unchanging semantics for all machine operations. 

For meta programming objects like `name`, we need all the optimizations from the native code, and often tweak the implementation for performance or behavioral issues. Adding these as non-meta definitions, and then giving them definitions in native code increases the maintenance burden, requiring us to faithfully reflect the behavior of C++ when changes are made.

It is also easy to introduce confusing behavior if we update one and not the other, which as the number increases is more and more likely to happen in my opinion, especially since we are lacking any safe guard ensuring the implementations are kept in sync.

I think we should devise a mechanism for making this maintenance easier before heavily using refinement by native code.

Leo and I have also been talking about rewriting the entire VM & native compiler compilation path to enable more optimizations, our current design plan would require an keeping 3 versions in sync, which is even more painful, and we still don't provide a way to ensure they stay in sync. 

There isn't a ton of value of having `name` by itself be non-meta imo, for it to be useful `expr` would have to be non-meta as well, and I don't see that happening any time soon, since we will have many of the same problems I listed above in greater quantity with `expr`.

cc @leodemoura

digama0:
@jroesch 
> For meta programming objects like name, we need all the optimizations from the native code, and often tweak the implementation for performance or behavioral issues. Adding these as non-meta definitions, and then giving them definitions in native code increases the maintenance burden, requiring us to faithfully reflect the behavior of C++ when changes are made.

Is this really the case though? I like to think of the non-meta implementation as a "specification" for the C++ equivalent. In most cases, it is a fairly "canonical" implementation, which should not require any modification unless `name` itself changes. For example, `name.append` surely is not going to change its observable behavior, even if it is reimplemented in C++ to make it more efficient or something. And if it does change, and that change is reflected in a change in the Lean specification, this is simply making it more explicit that your change has consequences on anything that references it, on the Lean side as well as the C++ side.

As a matter of fact, I have been investigating a formalization of non-meta `expr` (not to replace the current one, but a restricted version appropriate for formalizing CIC), but I'm trying to reuse as many components from the lean meta framework as possible, and it is an annoyance to see basic things like `name.append` which have simple non-meta definitions being `meta constant` instead for no good reason. The alternative here is simply that I define my own non-meta `name.append` with a new name, but that seems silly since we use VM overriding in similar circumstances. We can discuss using a `@[vm]` attribute for VM overriding, but that is a separate issue.

digama0:
Also, I only intend to add these sorts of definitions for reflecting bits of the kernel, because I am trying to make theoretical claims about Lean's foundational system, and I assume that the kernel is not the subject of any planned changes.

leodemoura:
> Also, I only intend to add these sorts of definitions for reflecting bits of the kernel, because I am trying to make theoretical claims about Lean's foundational system, and I assume that the kernel is not the subject of any planned changes.

@digama0 Yes, the kernel is not the subject of any planned changes.
However, we are planning to change the VM and the code generator (they are not part of the kernel).
The kernel is also immune to the specific total order implement by `name.lex_cmp`, but the code generator is. The code generator replaces `name.lex_cmp` with the builtin implementation in C++. I didn't check whether your implementation precisely implements the one in C++, nor I found any test to increase our confidence (e.g., `#reduce t` and `#eval t` produce the same result for several `name` primitives). `#reduce` uses the kernel reduction engine, and `#eval` the VM.

It is not clear to me why we need this PR to investigate Lean's foundational system. 
In principle, we can create a namespace `semantics` and define non-meta types: `name`, `expr`, etc.
This development would be completely independent of the `meta` objects we use for meta-programming.

digama0:
@leodemoura You are absolutely correct that we could just make a `semantics.*` version of all these types, which don't have any of this VM overriding stuff, and indeed this is the direction I am heading. I just think that we can reduce the duplication in some places when the real thing doesn't actually need that `meta` label. I actually expanded this PR to `level` locally, but I scaled back due to the earlier reported problems with non-injectivity of the the `level` constructors. The stuff in this PR is all I would reliably be able to claim as non-meta without changing the implementations of anything.

`name.lex_cmp` is in fact a replica of the C++ code, and `name.cmp` cannot be implemented in lean because it uses pointer comparison.

leodemoura:
> @leodemoura You are absolutely correct that we could just make a semantics.* version of all these types, which don't have any of this VM overriding stuff, and indeed this is the direction I am heading. 

Great.

> name.lex_cmp is in fact a replica of the C++ code, and name.cmp cannot be implemented in lean because it uses pointer comparison.

I'm not disputing that, I don't even remember the semantics of the C++ version.
Unfortunately, I don't have time to check. This is why I suggested the `#reduce` vs `#eval` tests.
Another concern is that even if it is correct, we would have to make sure that any change in the C++ version is reflected in the Lean version. I think this is an unlikely change.

This is similar to the other PR. I want to see discussion. I think @jroesch is also not thrilled about this PR. @kha and @gebner didn't manifest themselves.

digama0:
>I'm not disputing that, I don't even remember the semantics of the C++ version.
Unfortunately, I don't have time to check. This is why I suggested the #reduce vs #eval tests.
Another concern is that even if it is correct, we would have to make sure that any change in the C++ version is reflected in the Lean version. I think this is an unlikely change.

This brings up an important and general point: We currently have no mechanism for checking if a lean function matches its VM implementation, and I think this will only become more important as the number of custom VM implemented methods increases. One simple solution is to have test cases, and it would be good to automate this to a degree. I don't want to use `#reduce` because it's not always feasible, but I wonder if we can have `#eval` not use a specific VM override function (just use the lean function instead) for testing purposes? This could be used within the test harness framework.

gebner:
> @gebner didn't manifest themselves.

TLDR: I wouldn't want to formalize Lean's metatheory using `expr`.  The `name` type should be meta to prevent accidental use.

Why am I against using `expr` for formalization?
1. `level` is completely inconsistent
2. `expr.macro` is as meta as it gets
3. `expr.local_const` is unergonomic.  With LN as a binding strategy, I'd use `expr.local_const : nat -> expr` and pass around a local context / upper bound for the local constants instead of using `mk_fresh_name`.

The `name` type is easy enough to re-implement if you want exactly the same syntax.  You can even use tactics to derive decidable equality.  Performance should also be comparable (at least once we have efficient strings).  For most object languages, you probably don't want Lean's `name`.

digama0:
What's LN?

gebner:
Sorry, LN is locally nameless.  That is, the binding strategy used in Lean, where bound variables are identified using de Bruijn indices and free variables (local constants) are identified by names that are invariant under context additions.

gebner:
> One simple solution is to have test cases, and it would be good to automate this to a degree.

I agree, we should definitely have automated tests for this.  Even without a dedicated testing framework, we can still do some basic testing fairly easily right now:
```lean
run_cmd sequence $ do i ← list.range 100, j ← list.range 100, return $ do
unify (let res := i + j in reflect res) `(%%(reflect i) + %%(reflect j) : ℕ)
```

> I don't want to use #reduce because it's not always feasible, but I wonder if we can have #eval not use a specific VM override function

If I see this correctly, you're mainly referring to the fast operations on natural numbers, right?  One hacky way to do this is to copy the definitions onto a new environment.  So if we want to test `nat.div`, we could copy `nat.div` to `_under_test.nat.div`.  The VM doesn't replace the renamed version.

leodemoura:
> The name type is easy enough to re-implement if you want exactly the same syntax.
 
I agree.

The effort to make `name` non meta and making sure it can remain non meta is higher than having two copies.

I'm closing this PR

Issue #1641, opened 2017-06-02
`change at` tactic


digama0:
Currently, `change` only supports definitional rewriting of the goal. An optional `at` param should allow rewriting an element of the local context (without reverting it or causing type correctness issues, even when other hypotheses depend on it).

leodemoura:
Closed by commit above.

PR #1642, opened 2017-06-02
feat(init/meta/interactive): change at tactic

digama0:
addresses #1641

### DIFF
--- a/library/init/meta/interactive.lean
+++ b/library/init/meta/interactive.lean
@@ -267,8 +267,15 @@ This tactic applies to any goal. `change U` replaces the main goal target `T` wi
 providing that `U` is well-formed with respect to the main goal local context,
 and `T` and `U` are definitionally equal.
 -/
-meta def change (q : parse texpr) : tactic unit :=
-i_to_expr q >>= tactic.change
+meta def change (q : parse texpr) : parse location → tactic unit
+| [] := i_to_expr q >>= tactic.change
+| [h] :=
+  do num_reverted : ℕ ← get_local h >>= revert,
+     expr.pi n bi d b ← target,
+     e ← i_to_expr q,
+     tactic.change $ expr.pi n bi e b,
+     intron num_reverted
+| _ := fail "change does not support multiple locations"
 
 /--
 This tactic applies to any goal. It gives directly the exact proof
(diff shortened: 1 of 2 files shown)


leodemoura:
@digama0 Could you please add a test file for the new feature?

leodemoura:
Here is an example:
https://github.com/leanprover/lean/blob/92a72b238b062828afe69debe8df3709c3887b6e/tests/lean/run/tactic_ref.lean

jroesch:
If you are going to take the time to modify change, we might as well make change support multiple hypotheses  since the implementation for a single hypothesis is not much different then doing it for all of them. You should just need to revert/re-intro all of them.

johoelzl:
@jroesch can you give me a example to show the syntax? I guess `change T at a b c` does not make much sense. When I want to change more than one hyptothesis, I want to give different new types for these hyptohesis. If there is not a nice syntax it will be hard to document.

jroesch:
@johoelzl I believed our change tactic worked like Coq's but I see that might not actually be the case. It requires that we replace the "entire" goal? you can get away with `change (1 + 2) with 3 in *` in Coq, and it it would replace all occurrences of the subterm.

```
Lemma foo : 1 + 2 = 3 -> 1 + 2 = 3 -> 3 = 1 + 2.
Proof.
  change (1 + 2) with 3 in *.
  reflexivity.
Qed.
```

digama0:
@jroesch I added an implementation of `change A with B` along the lines you describe. `change with` can be used with any location specification, but `change at` can only work on one hypothesis, for exactly the reason @johoelzl says. In principle, we could just change all the hypotheses for the given target, but that has extremely limited practical application. (Another possible extension is to support multiple `change with` in a list, similar to `rw`.)

digama0:
The error reporting could still be improved. For example:

    example (h : 1 + 2 = 3) : true :=
    by change 3 = 4 at h

    tactic.change failed, given type
      3 = 4 → true
    is not definitionally equal to
      1 + 2 = 3 → true
    state:
    ⊢ 1 + 2 = 3 → true

As you can see, the error shows all the reverted hypotheses as well as the actual failed definitional equality. Is the `is_def_eq` predicate exposed to lean?

Issue #1621, opened 2017-05-30
Coinductive predicates


leodemoura:
Coinductive predicates are useful in practice and are used in projects such as CompCert. 
We can define coinductive predicates using an impredicative fixpoint in `Prop` without any kernel extension.
@gebner has manually defined a few examples using the approach: https://gist.github.com/66bab9997d4b37fdb0884f7bcab95418#file-coindpred-lean
He used an auxiliary inductive predicate to define the functional and minimize the amount of boilerplate code.
However, it would still be useful to automate the process and automatically generate the construction in the example above.

Kha:
In case someone wants to implement this purely in Lean, I'd be happy to implement user-defined commands for that. I didn't yet because I didn't see any need for them, but it shouldn't be very different from #1617.

johoelzl:
I want to see coinductive predicates in Lean. I will take a look at it.

leodemoura:
@kha @johoelzl The parser for `coinductive` command is similar to the one for the
`inductive` command. The drawback is that you will need to do it in C++.
The advantage is that we will make sure the syntax for the two commands is similar.
One possibility is to do the parsing in C++, and invoke tactics written in Lean.
This is not feasible for the inductive command since we need it before the tactic framework is defined, but this should not be a problem for `coinductive`.

leodemoura:
@johoelzl Can we close this issue? Or, is there something missing in the implementation?

johoelzl:
We can close it. I think it make sense to open separate issues when there are problems with `coinductive` or the `coinduction` tactic.

PR #1644, opened 2017-06-02
RFC: Implement coinductive predicates #1621

johoelzl:
This implements the construction of coinductive predicates. Currently only a simple form is supported, i.e. no mutual or nested corecursion.

I'm not sure what to do with the `corec` rule, it currently shows directly the internal inductive predicate.


leodemoura:
@johoelzl This is great. Are you going to try to reuse the C++ inductive parser?

johoelzl:
Yes I want to reuse the inductive parser.

Currently I'm working on mutual & nested definitions. When this works I will take a look into the parser.

johoelzl:
@leodemoura I added support for mutual coinductive predicates and the expected coinductive syntax. I'm not 100% sure if `shared_coinductive_cmd` is correctly implemented:
* do I need to create a `vm_state`, even when using `tactic_evaluator`? When not I get an assertion error that `g_vm_state` is not set.
* The `tactic_evaluator` either returns success or transforms the error into an c++ exception. Only when a silent exception happens nothing is done. What are the cases for a silent exception? 

What I want to implement in the next days:
- [x] syntax!
- [x] tests!
- [X]  nested recursion
- [X]  coinduction rule
- [X]  coinduction method
- [X]  producing a case rule
- [X]  setting up attributes

jroesch:
@johoelzl You probably need to scope a VM state, things which require a global VM must have one scoped by using the class `scope_vm_state`. For example:

```
src/library/inductive_compiler/nested.cpp
2255:        scope_vm_state vms_scope(vms);
```

johoelzl:
@jroesch I currently use `scope_vm_state`. But I was not sure if this is the right approach. My usage looks like the one in `nested.cpp`. So I guess I'm fine. Thanks.

leodemoura:
@johoelzl I can take a look tonight. Please post here any information you think is relevant.
I need one example. Should I use `tests/lean/run/coinductive.lean`?

johoelzl:
@leodemoura `coinduction` and `mutual coinduction` both work. I used the examples I added to `tests/lean/run/coinductive.lean`. My question w.r.t `scope_vm_state` is mostly if this is the right approach or not.

I didn't try any error cases yet. There are no error messages yet.

Essentially every predicate one can write down with `inductive` (without nested induction) should work now with `coinductive`.

johoelzl:
Sorry accidentally closed the PR.

src/frontends/lean/inductive_cmds.cpp
```diff
@@ -693,6 +697,61 @@ class inductive_cmd_fn {
...
+        vm_obj r = tactic_evaluator(m_ctx, opts, mk_true())
+            (mk_constant(get_tactic_add_coinductive_predicate_name()), args, s);
+        if (auto new_s = tactic::is_success(r)) {
+            m_env = new_s->env();
+            post_process(new_params, new_inds, new_intro_rules);
+            return m_env;
+        }
+
+        throw generic_exception(first_ind, "silent exception"); // TODO(): what should we report here?
+    }
```
leodemoura: The `tactic_evaluator` already reported the tactic failure.
We can simply say that the `coinductive` command failed. Example:
```cpp
    throw generic_exception(first_ind, "coinductive command failed");
```

src/frontends/lean/inductive_cmds.cpp
```diff
@@ -693,6 +697,61 @@ class inductive_cmd_fn {
...
+        buffer<vm_obj> inds_arg;
+        for (unsigned i = 0; i < new_inds.size(); i ++) {
+            inds_arg.push_back(mk_vm_pair(to_obj(new_inds[i]), to_obj(new_intro_rules[i])));
+        }
+        args.push_back(to_obj(m_lp_names));
+        args.push_back(to_obj(new_params));
+        args.push_back(to_obj(inds_arg));
+
+        vm_obj r = tactic_evaluator(m_ctx, opts, mk_true())
+            (mk_constant(get_tactic_add_coinductive_predicate_name()), args, s);
```
leodemoura: The third argument of `tactic_evaluator` is used to provide position information for error messages.
`mk_true` does not have position information associated with it.

src/frontends/lean/inductive_cmds.cpp
```diff
@@ -693,6 +697,61 @@ class inductive_cmd_fn {
...
+        }
+        elaborate_inductive_decls(params, inds, intro_rules, new_params, new_inds, new_intro_rules);
+
+        expr first_ind = inds[0];
+        metavar_context mctx;
+        options opts = m_p.get_options();
+
+        vm_state vms(m_env, opts);
+        scope_vm_state vms_scope(vms);
+        tactic_state s = mk_tactic_state_for(m_env, opts, mlocal_name(first_ind), mctx, local_context(), mk_true());
```
leodemoura: I just pushed a relevant fix: 
https://github.com/leanprover/lean/commit/544817cf150a726dbc531e7eb816f1921c478868
The lines 
```lean
  vm_state vms(m_env, opts);
  scope_vm_state vms_scope(vms);
```
are not needed anymore. The `tactic_evaluator` creates its own VM.

library/init/meta/tactic.lean
```diff
@@ -997,6 +997,20 @@ match _root_.try_for max (tac s) with
...
 end
 
+meta def updateex_env (f : environment → exceptional environment) : tactic unit :=
+do env ← get_env,
+   env ← returnex $ f env,
+   set_env env
+
+meta def add_decl (d : declaration) : tactic unit :=
+updateex_env $ λe, do e ← e.add d, e.compile d tt
+
```
leodemoura: What is the motivation for this change?
This seems to be a refactoring that has nothing to do with the coinductive command.
We need to discuss this kind of change. For example, I'm not sure I want to expose `compile` to users.

library/init/meta/coinductive_predicates.lean
```diff
@@ -0,0 +1,326 @@
...
+| (mk_numeral _ n) := last_string n
+
+end name
+
+namespace expr
+open expr
+
+meta def replace_with (e : expr) (s : expr) (s' : expr) : expr :=
+e.replace $ λc d, if c = s then some s' else none
+
```
leodemoura: `replace_with` will produce incorrect results if `s'` contains dangling de-Bruijn variables.
This is not an issue in your code since this case never occurs.
However, if the procedure `replace_with` is available, other users may invoke it with terms containing dangling deBruijn variables. I think we have two possible solutions: mark `replace_with` private, or use `lift_vars`.
```lean
meta def replace_with (e : expr) (s : expr) (s' : expr) : expr :=
e.replace $ λc d, if c = s then some (s'.lift_vars 0 d) else none
```

src/frontends/lean/inductive_cmds.cpp
johoelzl: Then I will use `first_ind` to return errors. I guess this will report the error at the first identifier. Or is it possible to encode the beginning of the `coinductive` command as `expr `?

src/frontends/lean/inductive_cmds.cpp
johoelzl: Nice. I will remove them.

library/init/meta/tactic.lean
johoelzl: It felt cleaner to provide more fine grained operations to the Lean side. I will revert `add_decl`, to be implemented in C++ again.

library/init/meta/coinductive_predicates.lean
johoelzl: Okay, I will add `lift_vars`.

src/frontends/lean/inductive_cmds.cpp
leodemoura: The `first_ind` is good. We use the approach in many other commands.

tests/lean/run/coinductive.lean
```diff
@@ -0,0 +1,83 @@
...
+
+#print prefix tt_stream
+#print prefix ff_stream
+
+#check (@tt_stream : stream bool → Prop)
+#check (@tt_stream.corec_functional :
+  ∀ (C_tt_stream C_ff_stream : stream bool → Prop),
+    (∀ (a : stream bool), C_tt_stream a → tt_stream.functional C_tt_stream C_ff_stream a) →
+    (∀ (a : stream bool), C_ff_stream a → ff_stream.functional C_tt_stream C_ff_stream a) →
+    ∀ (a : stream bool), C_tt_stream a → tt_stream a)
```
leodemoura, line 51: Are you planning to hide the auxiliary type `*.functional`?

johoelzl: `*.functional` is not mentioned in the constructor rules or `corec_on` and `cases_on` anymore. I can put anything having `*.functional` in its type into the `*.functional` namespace. I'm not sure about hiding it, sometimes the construction is interesting in itself.

Is there an easy way to hide the definitions?

johoelzl:
All features are implemented. I will add some tests for the `coinduction` tactic.

A couple of infrastructure changes are still missing:
* add an attribute to register coinductive rules, so the user does not need to write down the name of the rule
* add support for `cases` and `case`

leodemoura:
@johoelzl I think we can merge it as is, right?
The missing features can be in different PRs.

johoelzl:
Yes we can merge it.

PR #1645, opened 2017-06-03
chore(.appveyor.yml): separate upgrading and installing new packages

Kha:
Let's see if this changes anything.


Kha:
Yep, that fixed it.

PR #1648, opened 2017-06-06
feat(init/meta/interactive): specify intermediate goal in transitivity

digama0:
As suggested by @Armael on gitter

### DIFF
--- a/library/init/meta/interactive.lean
+++ b/library/init/meta/interactive.lean
@@ -794,10 +794,9 @@ tactic.symmetry
 meta def transitivity (q : parse texpr?) : tactic unit :=
 tactic.transitivity >> match q with
 | none := skip
-| some q := do t ← target, match t with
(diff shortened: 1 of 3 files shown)


library/init/meta/interactive.lean
```diff
@@ -791,8 +791,14 @@ tactic.reflexivity
 meta def symmetry : tactic unit :=
 tactic.symmetry
 
-meta def transitivity : tactic unit :=
-tactic.transitivity
+meta def transitivity (q : parse texpr?) : tactic unit :=
+tactic.transitivity >> match q with
+| none := skip
+| some q := do t ← target, match t with
+  | (app (app r lhs) rhs) := i_to_expr q >>= unify rhs
```
johoelzl: This doesn't work with `heq {α : Sort u} (a : α) : Π {β : Sort u}, β → Prop` where the arguments are the second and last parameter. Using `relation_lhs_rhs` to project out the lhs and rhs side should do it.

PR #1651, opened 2017-06-07
fix(library/quote): use opaque macro for elaborated expr quotations

Kha:


tests/lean/reflect.lean.expected.out
```diff
@@ -0,0 +1 @@
+has_zero.zero.{0} nat nat.has_zero
```
Kha, line 1: @rlewis1988: `nat_value_macro`s should be gone now

PR #1652, opened 2017-06-07
feat(.): add first pass at contributing & issue & pr templates

jroesch:
Added a first stab at templates for issues & PRs as well as a contribution guideline.


jroesch:
@Kha Thoughts if we should add more? @leodemoura and I thought we should add these to encourage only opening bug reports after finding a repro, etc.

Kha:
Sounds good to me

PR #1653, opened 2017-06-07
Format `reflected a`

robertylewis:
This allows you to write e.g. `tactic.trace (0 : ℕ)`. (Pretend that there's a backtick before `(0`. Is there a way to type that here?)

I've also adjusted `nat.reflect 0` to produce the numeral `0` instead of `nat.zero`.

### DIFF
--- a/library/init/meta/has_reflect.lean
+++ b/library/init/meta/has_reflect.lean
@@ -49,9 +49,8 @@ meta instance prod.reflect {α β : Type} [has_reflect α] [reflected α] [has_r
 meta instance pos.reflect : has_reflect pos
 | ⟨l, c⟩ := `(_)
 
-meta def reflect_to_format {α} {a : α} (h : reflected a) : format :=
+meta def reflected.to_format {α} {a : α} (h : reflected a) : format :=
 to_fmt (reflect a).to_expr
 
-meta instance reflected.has_to_format {α} (a : α) : has_to_format (reflected a) :=
-⟨reflect_to_format⟩
-
+meta instance {α} (a : α) : has_to_format (reflected a) :=
+⟨reflected.to_format⟩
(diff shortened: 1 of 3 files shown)


digama0:
To answer your side question, yes there is, just enclose everything in double backticks instead of single. ``tactic.trace `(0 : ℕ)``

library/init/meta/has_reflect.lean
```diff
@@ -48,3 +48,9 @@ meta instance prod.reflect {α β : Type} [has_reflect α] [reflected α] [has_r
 
 meta instance pos.reflect : has_reflect pos
 | ⟨l, c⟩ := `(_)
+
+meta def reflected.to_format {α} {a : α} (h : reflected a) : format :=
+to_fmt (reflect a).to_expr
+
+meta instance {α} (a : α) : has_to_format (reflected a) :=
+⟨reflected.to_format⟩
```
Kha: Logically, these definitions should probably go in `expr.lean` below the definition of `reflected`. That file is already quite big of course, but you can inline the first definition into the second one, which seems to be our standard for (non-recursive) meta instance definitions.

library/init/meta/has_reflect.lean
```diff
@@ -48,3 +48,9 @@ meta instance prod.reflect {α β : Type} [has_reflect α] [reflected α] [has_r
 
 meta instance pos.reflect : has_reflect pos
 | ⟨l, c⟩ := `(_)
+
+meta def reflected.to_format {α} {a : α} (h : reflected a) : format :=
+to_fmt (reflect a).to_expr
```
Kha: Same as `to_fmt h.to_expr`?

library/init/meta/has_reflect.lean
robertylewis: It was (very slightly) more effort to inline them, since the `reflected` instance wasn't available for type class inference in `reflect`. But using `to_fmt h.to_expr` avoids this.

PR #1656, opened 2017-06-08
fix(init/meta/interactive): allow dsimp to simplify several hypothesis

Armael:
`dsimp at H1 H2` would previously only simplify in hypothesis `H1`. It now runs `dsimp` on all hypothesis provided.

### DIFF
--- a/library/init/meta/interactive.lean
+++ b/library/init/meta/interactive.lean
@@ -657,9 +657,8 @@ do s ← mk_simp_set no_dflt attr_names hs wo_ids,
    end,
    try triv >> try (reflexivity reducible)
 
-private meta def dsimp_hyps (s : simp_lemmas) : list name → tactic unit
-| []      := skip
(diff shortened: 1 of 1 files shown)


library/init/meta/interactive.lean
```diff
@@ -657,9 +657,8 @@ do s ← mk_simp_set no_dflt attr_names hs wo_ids,
    end,
    try triv >> try (reflexivity reducible)
 
-private meta def dsimp_hyps (s : simp_lemmas) : list name → tactic unit
-| []      := skip
-| (h::hs) := get_local h >>= dsimp_at_core s
+private meta def dsimp_hyps (s : simp_lemmas) (hs : list name) : tactic unit :=
+  list.mfor' hs (λ h, get_local h >>= dsimp_at_core s)
```
Kha: This line should not be indented. You can also use the `hs.mfor'` shorthand.

Armael:
I implemented the suggested changes

PR #1661, opened 2017-06-09
adapt leanpkg to change in to_string method for char

avigad:
The recent changes to string have broken `leanpkg`: if you do `leanpkg init foo` the `leanpkg.toml` file looks like this:
```
[package]
name = "'f''o''o'"
version = "'0''.''1'"
```
The problem is that there is now a difference between `to_string 'a'` and `'a'.to_string`:
``` lean
#eval to_string 'a'  -- "'a'"
#eval 'a'.to_string  -- "a"
```
We can discuss whether or not this is problematic, but in the meanwhile the easy fix is to make `leanpkg` use the correct one.

### DIFF
--- a/leanpkg/leanpkg/toml.lean
+++ b/leanpkg/leanpkg/toml.lean
@@ -28,7 +28,7 @@ private def escapec : char → string
 | '\t' := "\\t"
 | '\n' := "\\n"
 | '\\' := "\\\\"
-| c    := to_string c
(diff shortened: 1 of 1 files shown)


gebner:
Thanks!  I guess we should add some tests/lemmas for leanpkg..

I don't know whether to_string should quote a character or not.  But I agree, it is very confusing if it is inconsistent.

PR #1665, opened 2017-06-13
Bitwise ops for integers

digama0:
Expansion of the previous commit to include VM implementations of integer bitwise operations.

### DIFF
--- a/library/init/data/int/basic.lean
+++ b/library/init/data/int/basic.lean
@@ -6,7 +6,7 @@ Authors: Jeremy Avigad
 The integers, with addition, multiplication, and subtraction.
 -/
 prelude
-import init.data.nat.lemmas init.meta.transfer
+import init.data.nat.lemmas init.data.nat.gcd init.meta.transfer init.data.list
 open nat
 
 /- the type, coercions, and notation -/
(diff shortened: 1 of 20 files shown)


library/init/data/int/basic.lean
```diff
@@ -382,4 +417,10 @@ by rw -int.sub_nat_nat_eq_coe; exact sub_nat_nat_elim m n
   (λi n, by rw [nat.add_sub_cancel_left]; refl)
   (λi n, by rw [add_assoc, nat.sub_eq_zero_of_le (nat.le_add_right _ _)]; refl)
 
+#eval do m ← list.range 5,
```
gebner: Debugging code should not be in the library files.

library/init/data/int/bitwise.lean
```diff
@@ -0,0 +1,63 @@
...
+
+universe u
+
+namespace int
+
+  def test_bit : ℤ → ℕ → bool
+  | (m : ℕ) n := nat.test_bit m n
+  | -[1+ m] n := bnot (nat.test_bit m n)
+
+  def nat_bitwise (f : bool → bool → bool) (m n : ℕ) : ℤ :=
```
gebner: This function (and `bitwise`) doesn't seem to be used anywhere.  Is there a reason we can't use it for `lor`, `land`, etc.?

library/init/data/nat/bitwise.lean
```diff
@@ -11,15 +11,78 @@ universe u
...
+  | 0        := (ff, 0)
+  | (succ n) :=
+    match bodd_div2 n with
+    | (ff, m) := (tt, m)
+    | (tt, m) := (ff, succ m)
+    end
+
+  def div2 (n : ℕ) : ℕ := (bodd_div2 n).2
+
+  def bodd (n : ℕ) : bool := (bodd_div2 n).1
```
gebner: What was wrong with `n % 2 = 1`?

library/init/data/nat/gcd.lean
```diff
@@ -0,0 +1,61 @@
...
+prelude
+import init.data.nat.lemmas init.wf
+
+open well_founded
+
+namespace nat
+
+/- gcd -/
+
+def gcd.F : Π (y : ℕ), (Π (y' : ℕ), y' < y → nat → nat) → nat → nat
```
gebner: Can we define `gcd` using the equation compiler now?

gebner (changes requested):

library/init/data/nat/gcd.lean
digama0: Yes. We should probably also define `div` and `mod` using the equation compiler.

library/init/data/nat/bitwise.lean
digama0: Dependence on WF definitions makes it more complicated for the kernel to reduce the definition. I didn't think it would be important, but it kept breaking proofs in weird ways.

gebner:
BTW, this PR doesn't build.

library/init/data/int/bitwise.lean
digama0: I'm torn on whether the definitional equality should be the reduction into positive/negative cases as I've done here, or the theorems `bitwise or = lor` etc. Either way we will have both, but definitional reduction can be important here.

digama0:
Errors in the last commit were fixed, and it has been build-tested this time.

library/init/data/nat/bitwise.lean
gebner: > I didn't think it would be important, but it kept breaking proofs in weird ways.

Do you still remember what it broke?  As far as I can tell, `bodd_div2 n` only reduces if `n` is a numeral.  But in this case I would have expected the WF definition to reduce as well.

library/init/data/nat/bitwise.lean
```diff
@@ -11,15 +11,78 @@ universe u
...
+
+  @[simp] lemma bodd_zero : bodd 0 = ff := rfl
+  @[simp] lemma bodd_one : bodd 1 = tt := rfl
+  @[simp] lemma bodd_two : bodd 2 = ff := rfl
+
+  @[simp] def bodd_succ (n : ℕ) : bodd (succ n) = bnot (bodd n) :=
```
gebner: `s/def/lemma/g`

library/init/data/int/bitwise.lean
gebner: I see, they are used in the theorems at the end of the file.

gebner:
@digama0 This PR still has build errors.

library/init/data/nat/bitwise.lean
digama0: It's not that it can't reduce; it reduces but only when `unfold_lemmas` is enabled. This causes a problem for  `binary_rec_eq`, because the type of the theorem includes `f ff 0 z = z` (where the left side has type `C (bit ff 0)` and the right side has type `C 0`), so that any users of this theorem must also have `unfold_lemmas` enabled in order to reduce `bit ff 0 := 0`. If they don't, we are in the strange circumstance of having equalities embedded in constants whose type-correctness is not verifiable, and I have no idea what kind of guarantees lean can give in this situation.

gebner: Oooh, I see.  Yes, this is really, really ugly.  And we will probably see this again in the future, since it happens with every WF definition...

I see two potential solutions:
1. Use equational rfl-lemmas for reduction in the type_context.  This is a bit of work, and Leo has mentioned that there are lots of issues here.
2. As a quick hack, we could special-case `acc.rec` (and maybe `eq.rec` and `heq.rec`?) in the type_context by reducing the major premise using `transparency_mode::All`.

BTW, we have exactly the same issue with `attribute [irreducible]`.

digama0:
I think this PR is ready for merge (or at least ready for others to try to find issues with it) now.

digama0:
@gebner Is the "request changes" finished?

gebner (approved):

gebner:
> Is the "request changes" finished?

I still think that bodd_div2 is hideous, but I get why we need it.

We should define `gcd` using the equations compiler, but that can be done later.  I don't think we can use the equations compiler for `div` and `mod` for bootstrapping reasons, since they are required by the tactics framework (more precisely `init.data.repr`) and the well-founded recursion support requires tactics.

digama0:
@gebner Actually the latest version of this PR does define `gcd` using the equation compiler. It did involve changing the definition so that it did induction on the left argument instead of the right, but I suppose `gcd` is commutative anyway so all the proofs are simply left-right mirrored.

The other reason for `bodd_div2`, which would remain even if WF definitions worked perfectly, is that it has different definitional equalities. I wanted `bodd (succ n) = bnot (bodd n)` to be a definitional equality instead of whatever you get with the mod definition.

leodemoura:
Merged based on @gebner approval.

PR #1666, opened 2017-06-14
create change log

avigad:
This wasn't hard to do, so I don't think it will be terribly hard to maintain in the future. You can see how it looks here:

  https://github.com/avigad/lean/blob/change_log/doc/changes.md

Let me know if you notice anything missing. I'll wait a few days, and then incorporate additions and comments.

This isn't a replacement for documentation. In particular, documenting well-founded recursion, coinductive predicates, etc. is high priority. When they are described in TPL, we can update these notes.


jroesch:
We should make a notes that we:
- do not support mutual meta definitions
- commands beginning with # are expected to be transient and should not be committed 

Otherwise looks pretty good, we have done a lot in the past few months.

PR #1667, opened 2017-06-14
Transient message boxes in Emacs

david-christiansen:
# Transient message boxes in Emacs

Adds a message display hook to the Emacs mode, and uses it to display boxes containing the results of various commands in the buffer, in addition to the Flycheck interface. The categories of output to display are customizable, and `C-c C-b` toggles the display of boxes in `lean-mode`.

screenshot from 2017-06-13 17-49-08

### DIFF
--- a/src/emacs/README.md
+++ b/src/emacs/README.md
@@ -20,9 +20,9 @@ Trying It Out
 If things are working correctly, you should see the word ``Lean`` in the
 Emacs mode line when you open a file with extension `.lean`. If you type
 ```lean
-check id
+#check id
 ```
-the word ``check`` will be underlined, and hovering over it will show
+the word ``#check`` will be underlined, and hovering over it will show
 you the type of ``id``. The mode line will show ``FlyC:0/1``, indicating
 that there are no errors and one piece of information displayed.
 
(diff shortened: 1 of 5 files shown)


jonsterling:
Whoa, this is super sweet. Maybe I'll resurrect RedPRL's emacs mode and try this out...

david-christiansen:
@jonsterling The code as written is only somewhat portable to other proof systems, but it could be abstracted into an elisp library if you want it for RedPRL.

Kha:
This is nice! Should there be a `(lean-server-show-messages)` on toggling so that the display is changed immediately?
What do you think of using a more efficient presentation for single-line messages, like what IntelliJ does during debugging?
inline variables view

david-christiansen:
> Should there be a (lean-server-show-messages) on toggling so that the display is changed immediately?

Yes, I think so. Good idea. I can update the PR later today.

> What do you think of using a more efficient presentation for single-line messages, like what IntelliJ does during debugging?

Also a good idea. The only downside that I can think of is that the presentation of the box version and the single-line version would become quite visually different, but that's more of an argument for finding an alternative graphical presentation for the multi-line version. Perhaps instead of a box, it should just be a vertical line on the left combined with putting it in a different face.

david-christiansen:
Now it looks like this:
screenshot from 2017-06-14 08-04-27

Kha:
Very nice!

david-christiansen:
Thanks!

Issue #1483, opened 2017-03-24
Allow nested `case` distinction


Kha:
We can use a custom interactive parser to allow patterns such as
```lean
 -- ex : ∃ a, p₁ a ∧ p₂ a
cases ex with a ⟨h₁, h₂⟩
```

Kha:
It would also be nice to support case distinction in `intro(s)`, in analogy to `λ`.

spl:
Along with these suggestions (which are all great), it would be nice to support types in `intro` as in `∀`/`λ`, e.g. `intro x : A`, `intro (x : A) (px : ∃ y, y > x)`, etc.

spl:
Oh, and it would also be nice to support types in `cases`.

Kha:
@digama0 has implemented a "recursive cases" tactic in mathlib: https://github.com/leanprover/mathlib/commit/c5c5b504b918a6c5e91e372ee29ed754b0513e85
As such, I'm closing this issue. We can open focused issues for the other suggestions noted above.

PR #1670, opened 2017-06-14
refactor(init/meta/interactive): merge assert -> note

digama0:
This merges the interactive mode `assert h : T` into `note h : T`, further extending the variations of `note` so that in `note h : t := p`, all of `h`, `t`, and `p` are optional. Additionally, `t` is no longer entered in "strict expr" mode, so holes in `t` may introduce new metavariables. Finally, the fact that all of `h`,`t`,`p` are optional introduces a new possibility, bare `note`, which acts like `assert this : _`, which is to say, it makes the current goal a metavariable, and adds `this : ?m_1` to the original goal.

All the same comments apply to merging `define h : T` => `pose h : T`.

### DIFF
--- a/library/data/bitvec.lean
+++ b/library/data/bitvec.lean
@@ -192,7 +192,7 @@ section conversion
     revert n,
     induction k with k ; intro n,
     { unfold pow, simp [nat.mod_one], refl },
-    { assert h : 0 < 2, { apply le_succ },
+    { note h : 0 < 2, { apply le_succ },
       rw [ of_nat_succ
          , to_nat_append
          , ih_1
--- a/library/data/bitvec.lean
+++ b/library/data/bitvec.lean
@@ -165,7 +165,7 @@ section conversion
     simp [bits_to_nat_to_list], clear P,
     unfold bits_to_nat list.foldl,
       -- the next 4 lines generalize the accumulator of foldl
-    pose x := 0,
+    define x := 0,
     change _ = add_lsb x b + _,
     generalize 0 y,
     revert x, simp,
--- a/library/data/bitvec.lean
+++ b/library/data/bitvec.lean
@@ -42,9 +42,9 @@ section shift
   bitvec.cong
     begin
       by_cases (i ≤ n),
-      { note h₁ := sub_le n i,
+      { have h₁ := sub_le n i,
         rw [min_eq_right h], rw [min_eq_left h₁, -nat.add_sub_assoc h, add_comm, nat.add_sub_cancel] },
-      { note h₁ := le_of_not_ge h,
+      { have h₁ := le_of_not_ge h,
         rw [min_eq_left h₁, sub_eq_zero_of_le h₁, min_zero_left, add_zero] }
     end $
     repeat fill (min n i) ++ₜ taken (n-i) x
@@ -165,7 +165,7 @@ section conversion
     simp [bits_to_nat_to_list], clear P,
     unfold bits_to_nat list.foldl,
       -- the next 4 lines generalize the accumulator of foldl
-    define x := 0,
+    let x := 0,
     change _ = add_lsb x b + _,
     generalize 0 y,
     revert x, simp,
@@ -192,7 +192,7 @@ section conversion
     revert n,
     induction k with k ; intro n,
     { unfold pow, simp [nat.mod_one], refl },
-    { note h : 0 < 2, { apply le_succ },
+    { have h : 0 < 2, { apply le_succ },
       rw [ of_nat_succ
          , to_nat_append
          , ih_1
--- a/library/data/dlist.lean
+++ b/library/data/dlist.lean
@@ -68,7 +68,7 @@ by cases l; simp
 lemma of_list_to_list (l : dlist α) : of_list (to_list l) = l :=
 begin
    cases l with xs,
-   note h : append (xs []) = xs,
+   have h : append (xs []) = xs,
    {intros, apply funext, intros x, simp [invariant x]},
    simp [h]
 end
--- a/library/data/dlist.lean
+++ b/library/data/dlist.lean
@@ -68,7 +68,7 @@ by cases l; simp
 lemma of_list_to_list (l : dlist α) : of_list (to_list l) = l :=
 begin
    cases l with xs,
-   assert h : append (xs []) = xs,
+   note h : append (xs []) = xs,
    {intros, apply funext, intros x, simp [invariant x]},
    simp [h]
 end
--- a/library/data/hash_map.lean
+++ b/library/data/hash_map.lean
@@ -41,7 +41,7 @@ begin
   intros,
   induction j with j IH,
   exact ⟨false.elim, λ⟨i, h, _⟩, absurd h (nat.not_lt_zero _)⟩,
-  note IH := IH (le_of_lt h),
+  have IH := IH (le_of_lt h),
   simp[array.iterate_aux],
   exact ⟨λo, or.elim o
     (λm, ⟨⟨j, h⟩, nat.le_refl _, m⟩)
@@ -165,10 +165,10 @@ have h1 : list.length (array.to_list bkts) - 1 - i < list.length (list.reverse (
 have sigma.mk a b ∈ list.nth_le (array.to_list bkts) i (by simph[array.to_list_length]), by {rw array.to_list_nth, exact el},
 begin
   rw -list.nth_le_reverse at this,
-  note v : valid_aux (λa, (mk_idx n (hash_fn a)).1) (array.to_list bkts).reverse sz,
+  have v : valid_aux (λa, (mk_idx n (hash_fn a)).1) (array.to_list bkts).reverse sz,
   rw array.to_list_reverse,
   exact v,
-  note mm := @_root_.hash_map.valid_aux.eq _ _ _ _ _ _ v -- TODO (Mario): Why is explicit namespacing needed here?
+  have mm := @_root_.hash_map.valid_aux.eq _ _ _ _ _ _ v -- TODO (Mario): Why is explicit namespacing needed here?
     (list.length (array.to_list bkts) - 1 - i) h1 a b this,
   rwa [list.length_reverse, array.to_list_length, nat.sub_sub_self (show i ≤ n.1 - 1, from nat.pred_le_pred h)] at mm
 end
@@ -190,7 +190,7 @@ begin
     let ⟨⟨a', b⟩, m1, e1⟩ := list.exists_of_mem_map m1 in
     let ⟨⟨a'', b'⟩, m2, e2⟩ := list.exists_of_mem_map m2 in
     match a', a'', b, b', m1, m2, e1, e2 with ._, ._, b, b', m1, m2, rfl, rfl :=
-      by {note hlt := al _ m1, rw v.eq _ m2 at hlt, exact lt_irrefl _ hlt}
+      by {have hlt := al _ m1, rw v.eq _ m2 at hlt, exact lt_irrefl _ hlt}
     end,
   λ⟨a, b⟩ m, or.elim m
     (λm2, by rw v.eq _ m2; exact nat.le_refl _)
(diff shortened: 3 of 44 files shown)


johoelzl:
Can we then rename `pose` to `define`? I think this would make the difference between `note` and `pose`/`define` clearer. What is the general opinion on this?

avigad:
I originally thought combining `assert` and `note` was a good idea, but after writing tutorial documentation yesterday, I am not so sure. It seems useful to have a clear distinction between `assert`, which introduces a new subgoal, and `note`, which "inlines" the proof.

I can't make up my mind which I like better. Both are o.k. with me. 

If we combine `define` and `pose`, I also don't have a strong feeling about which word to use. `define` is a little clearer, `pose` is a little shorter to type.

digama0:
@avigad I think a good way to explain the relation between the two is to say that `note h : t := by tac` and `note h : t, { tac }` are equivalent, which is not precisely true but is good enough for basic documentation.

I also have no opinion on `define` vs. `pose`. Neither one gets much use, so I don't think it matters too much to have a longer name.

avigad:
O.k., on the philosophy that fewer terms are better, I am sold on folding `assert` into `note`. And if @johoelzl  is willing to change `pose` to `define` everywhere,  that sounds good to me too. Is `assert` a better word than `note`?

digama0:
Someone (I think Leo) expressed a preference for `note` over `assert`, which is why I went with that. Also I would like to keep the `assert` name open for other possible uses involving assertions in the C sense. (It's not a huge deal, but I'd prefer to keep those senses separate.)

digama0:
Bump. Is there anything this PR is waiting for? (I would have preferred this ship with v3.2.0, since it noticeably affects the user behavior, but that ship has sailed.)

gebner:
@digama0 I think Leo wasn't very happy about the renaming, but I can't recall any details.

For my part, I really like the change from `pose` to `define`.  I never really remembered what `pose` did, and `define` seems much more intuitive to me.  Having just two tactic for introducing assumptions really cuts down the complexity as well, now there's just `note` for lambda-bindings and `define` for let-bindings and that's it.

gebner:
@digama0 This PR has build errors.

digama0:
@gebner I'm not getting any errors. Are you talking about the test suite, because I usually let Leo handle updating the tests. (I don't know how to run the test suite without getting hundreds of unrelated expected output mismatch errors.)

gebner:
> Are you talking about the test suite

Yes.

> without getting hundreds of unrelated expected output mismatch errors.

Hmm, this weird and shouldn't happen.  Do you have untracked files in the `tests` directory?

Kha:
> If we combine define and pose, I also don't have a strong feeling about which word to use. define is a little clearer, pose is a little shorter to type.

For an even shorter name, how about `let` :smile: ? We already allow `let x := e,` in `do` blocks, we can support the same syntax in tactics. We still need an underlying tactic to desugar to, but we don't have to advertise that one. You can also use it if you really want to leave out the right-hand side.

Similarly, after the recent `show` change, `have` should probably return to tactic mode, i.e. desugar to `note`.

These suggestions are mostly orthogonal to the PR at hand, but may make the naming issues less critical.

gebner:
I also like the name `have` better than `note` or `assert`.  But how would `have` work with missing proof terms?  With this PR both of the following are supported:
```lean
assert x > 0,
assert x > 0 := h x,
```

digama0:
I also think it is a good idea to unify the naming between term proof mode and tactic mode. How about the following: `define` is renamed to `let`, with the same behavior as in this PR, and `note` becomes `have`. (Yes, that means we get `have h := term` instead of `have h : _, from term`.) `change` is merged into `show`. This will make it a lot easier for users to remember all the main constructs.

Kha:
@gebner We already have the syntax `have p, { tac }` which then should be desugared to `assert p, { tac }`. At that point, I feel like we may as well accept just `have p,` so that only `from` and `by` following `have` are special-cased.

digama0:
One small point: This PR allows `note : type,` but not `note type` in order to state a lemma. Both `assert` and `have` seem to not need the colon, but this is required if we want to have the type and identifier both be optional. Unless we parse more context-sensitively, but that will make it harder to understand.

Kha:
```
assert x > 0,
assert x > 0 := h x,
```

Btw, I don't think these will actually work since `x` will be parsed as the hypothesis name. The implementation of `have` does some tricks to parse this without backtracking and which are not exposed to Lean right now, so that is one more reason to utilize it.

leodemoura:
> @digama0 I think Leo wasn't very happy about the renaming, but I can't recall any details.

@gebner Yes, I really dislike this kind of PR. First, I wanted to see more discussion, since the PR is a cosmetic change. For example, in the past, people (I don't remember who) made the point that we should have tactics with names similar to the ones used in Coq (e.g., `assert`, `note`, etc). Now, suppose I blindly merge this commit, then we may be in the same situation again, and  people try to add the removed tactics back. 
Second, I don't like broken PRs. It takes time to fix tests. 
Third, this PR will break existing developments (e.g., KreMLin project). I don't think this is the "end of the world", but we need to discuss and make sure we have a nice solution before we merge, and break backward compatibility.

Now, @kha and you made several comments and provided feedback (this is great). After we reach a consensus, I will be happy to merge the PR.

digama0:
@leodemoura Actually, it's not just a cosmetic change, and there are a few features hidden in here that you can't do currently, like metavariables in the asserted type, which should explain why I want this to be merged. Frankly, I don't care about the name chosen, and am happy to defer to whatever bikeshed people want to get behind.

There are also plenty of ways to do this backwards-compatibly, such as making both `note` and `assert` act like this and just be aliases of each other, but I have gotten the sense from the lean team that aliases are to be avoided, even if it breaks people's code. We haven't shied away from breaking changes in the recent past (such as `check` -> `#check`) with the reasoning that the lean ecosystem is not yet large enough for this to be a large burden, so we should "clean the cruft" as much as we can before it stops being feasible and we are forced to consider backward compatibility more seriously.

Are there any tests that are still broken? I tried to get all the tests I could, but my test suite still gives me a lot of errors in interactive mode even on "working" builds, and I'm still investigating. It looks like the CI likes the latest commit, except for one of the travis machines which crashed on `/home/travis/build/leanprover/lean/build/packages/smt2_interface/src/smt2/except.lean` (which doesn't exist for me). Where is the source for this test?

leodemoura:
@digama0 Yes, we haven't shied away from breaking backward compatibility. However, we usually discuss the change and make sure it is not an arbitrary change. For example, the `check` -> `#check` change was discussed and we had a good motivation for doing it.

I'm not against this PR, but I want to see discussion. As I said above, in the past, we had people arguing we should have tactics named `assert`, `note`, ... because Coq has similar tactics. 
@kha and @gebner provided useful feedback today after you sent the "bump". This is great. 
For me, the ideal scenario is: we converge to a solid proposal and implementation, and if in the future someone raises the issue "why don't you have `assert` like Coq?", we just send the link to this PR.

> Actually, it's not just a cosmetic change, and there are a few features hidden in here that you can't do currently, like metavariables in the asserted type, which should explain why I want this to be merged. 

This is good. 

> Are there any tests that are still broken?

I don't know. I haven't tested it again.

digama0:
It looks like `show` is special cased in the parser, to redirect to `interactive.show_goal`. Would we need to do a similar thing in order to use `have` instead of `note` for this tactic?

leodemoura:
> It looks like show is special cased in the parser, to redirect to interactive.show_goal. Would we need to do a similar thing in order to use have instead of note for this tactic?

Yes :(
`have` is a keyword, and we will not be able to define `tactic.interactive.have`.

Kha:
So, everyone in favor of introducing `let`/`have` as the preferred sugar for `define`/`note`? I can do the change and push it into this PR or a new one.

In any case, `define` and `note` should still be changed to take a non-optional identifier to remove the parsing ambiguity I mentioned above.

Kha:
@digama0 
> Where is the source for this test?

These packages are checked out by `script/test_registry.py`. It's okay to fail that test, we'll just have to fix them after merging the PR.

digama0:
@Kha I'm not sure I want to compromise on the optional identifier thing. I would much rather have the setup in this PR, where the colon is non-optional for stating a goal without an identifier. It is easy enough to parse these as all distinct, and the pattern is easy to understand and document:

    let,
    let h,
    let : type,
    let h : type,
    let := term,
    let h := term,
    let : type := term,
    let h : type := term,

`let type` will thus be either rejected or misinterpreted as an identifier if you try to drop the colon.

avigad:
Here is my two cents. I like the uniform `op h : T := t` notation, with each parameter optional. My preferences for names are `have`/`let`, then `assert`/`define`, then `note`/`pose`, but I am o.k. with any of the names in any combination. I don't think it is all that bad if the syntax for `have` in tactic mode is slightly difference from the syntax in proof term mode.

While we are on the topic of syntax for tactic proofs, note that we could also apply the `:=` for `show`, that is, `show T := t`. If we do that, then `:=` is uniform notation for "inline a proof term."

After that, the main structuring mechanism missing from proof term mode is that ability to type an "assume." When trying to make tactic proofs clearer, I have written things like this:
```lean
intros m n h,  -- m n : ℕ, hw
```
It would be nice if we could say `intro h : m > n` or `assume h : m > n`.

digama0:
And while we're on *that* topic, one thing I really miss in tactic mode is
match expressions and destructuring let / "obtain". The cases tactic does
not allow more than one destructuring at a time, so it is considerably more
verbose than unfolding 6 existentials with `let ⟨a,b,c,d,e,f,hyps⟩ := proof
in ...`. If we are relaxing the indentifier lists in intros and cases to
patterns, then this is something I would like to be able to do, say by

    cases proof with ⟨a,b,c,d,e,f,hyps⟩

Kha:
@digama0 
> I would much rather have the setup in this PR, where the colon is non-optional for stating a goal without an identifier.

Oh I see, so it's not ambiguous after all. I'm fine with the underlying tactics using this notation, but do you think we should add it to let too (if we do, we may as well add it to term let too)? It doesn't strike me as a very useful combination, in contrast to anonymous `have`.

Btw, if we allowed the same syntax in `assume : p` (or even styled as `assume: p`?), we could obsolete `suppose`.

> If we are relaxing the indentifier lists in intros and cases to patterns

This is https://github.com/leanprover/lean/issues/1483. I'd be happy to implement it after we reach some consensus.

Kha:
I have renamed `note`/`defined` to `have_tac`/`let_tac` in https://github.com/Kha/lean/commit/25c7d702d82c598b63eec5b9aa230f4e57522e31, which will automatically be used by the respective keywords.

leodemoura:
> Oh I see, so it's not ambiguous after all. I'm fine with the underlying tactics using this notation, but do you think we should add it to let too (if we do, we may as well add it to term let too)? It doesn't strike me as a very useful combination, in contrast to anonymous have.

I agree, anonymous `let` doesn't sound very useful.

> Btw, if we allowed the same syntax in assume : p (or even styled as assume: p?), we could obsolete suppose.

Good for me.

leodemoura:
> This is #1483. I'd be happy to implement it after we reach some consensus.

A few days ago, I discussed #1483 with @Armael. He is missing it in the KreMLin project.
There are two missing features in the `cases` tactic. 
1- A nice way to group the names for each constructor/minor premise. 
I think this is no so critical anymore since we have the `case` tactic.
2- Nested `cases` (like the one posted by @digama0). It seems we would also have to extend the `case` tactic. BTW, @david-christiansen suggested we add a more robust solution where we tag meta variables for different cases.

Kha:
I finally managed to push my commit onto this PR (you need an access token if you have 2FA activated). I'd propose to merge the PR as is and discuss the other proposals in more appropriate issues. I will open one for reconciling term mode and tactic mode keywords.

leodemoura:
merged

Kha:
@leodemoura 
>> Btw, if we allowed the same syntax in `assume : p` (or even styled as `assume: p`?), we could obsolete suppose.

> Good for me.

I didn't know that `assume` takes arbitrary binders, so having `suppose` as a separate keyword doesn't sound too bad to me any more.

PR #1671, opened 2017-06-15
feat(emacs): Add a right-click menu to the Emacs mode

david-christiansen:
The right-click menu is computed from a list of functions that produce
menu items. Each menu item is a string and a closure to be invoked if
the menu item is selected.

The only menu item right now is "Find definition", but this approach
is extensible, so more can be implemented.

### DIFF
--- a/src/emacs/lean-info.el
+++ b/src/emacs/lean-info.el
@@ -67,6 +67,26 @@
       (lambda (&key record)
         (funcall cont record))))))
 
+(defun lean-info-right-click-find-definition ()
+  "Offer to jump to definition of right-click target."
+  (interactive)
+  (list 'info
+        (list :file_name (buffer-file-name)
+              :line (line-number-at-pos)
+              :column (lean-line-offset))
+        (cl-function
+         (lambda (&key record)
+           (let ((source-record (plist-get record :source)))
+             (if source-record
+                 (let ((full-name (plist-get record :full-id)))
+                   (list
+                    (list :name (if full-name
+                                    (concat "Find definition of " full-name)
+                                  "Find definition")
+                          :action (lambda ()
+                                    (apply #'lean-find-definition-cont source-record)))))
+               (list)))))))
+
 (cl-defun lean-find-definition-cont (&key file line column)
   (when (fboundp 'xref-push-marker-stack) (xref-push-marker-stack))
   (when file
--- a/src/emacs/lean-server.el
+++ b/src/emacs/lean-server.el
@@ -59,7 +59,7 @@ least the following keys:
        (lean-server-notify-tasks-changed sess old-tasks)))
     ("error"
      (message "error: %s" (plist-get res :message))
-     ; TODO(gabriel): maybe even add the error as a message
+     ;; TODO(gabriel): maybe even add the error as a message
      (when (plist-get res :seq_num)
        (let ((cb (lean-server-session-pop-callback sess (plist-get res :seq_num))))
          (when (cdr cb) (funcall (cdr cb) res)))))
@@ -370,13 +370,21 @@ least the following keys:
   "Sends a command to the lean server for the current buffer, waiting for and returning the response"
   ;; inspired by company--force-sync
   (let ((res 'trash)
+        (ok t)
         (start (time-to-seconds)))
-    (lean-server-send-command cmd params (lambda (&rest result) (setq res result)))
+    (lean-server-send-command cmd params
+                              (lambda (&rest result) (setq res result))
+                              (cl-function
+                               (lambda (&key message)
+                                 (setq ok nil)
+                                 (setq res message))))
     (while (eq res 'trash)
       (if (> (- (time-to-seconds) start) company-async-timeout)
           (error "Lean server timed out")
         (sleep-for company-async-wait)))
-    res))
+    (if ok
+        res
+      (error res))))
 
 (defun lean-server-sync (&optional buf)
   "Synchronizes the state of BUF (or the current buffer, if nil) with the lean server"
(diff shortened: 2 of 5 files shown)


david-christiansen:
Rebased on top of #1672 and added support for right-click menus on holes.

src/emacs/lean-info.el
```diff
@@ -67,6 +67,26 @@
...
+  (with-demoted-errors "lean get info: %s"
+    (lean-server-send-synchronous-command
+     'info (list :file_name (buffer-file-name)
+                 :line (line-number-at-pos)
+                 :column (lean-line-offset)))))
+
+(defun lean-info-right-click-find-definition ()
+  "Offer to jump to definition of right-click target."
+  (interactive)
+  (let* ((info-record (plist-get (lean-get-info-record-at-point-sync) :record))
```
gebner: The info command may not return immediately.  For example when dependencies of the current file are out of date, the info command will only return after the dependencies have been processed.

As a result, doing a right click may hang emacs for some time (seconds, minutes) if we make this request synchronously.  I think we should just show the "find definition" menu entry unconditionally, and then do the info request asynchronously in the callback.

(A related problem occurs in auto-completion.  Since company-mode does not support async, auto-completion may hang...  That's one of the reasons why we don't do auto-completion on TAB by default.)

src/emacs/lean-server.el
```diff
@@ -44,8 +47,9 @@
        (setf (lean-server-session-tasks sess) res)
        (lean-server-notify-tasks-changed sess old-tasks)))
     ("error"
-     (message "error: %s" (plist-get res :message))
-     ; TODO(gabriel): maybe even add the error as a message
+     (when lean-server-print-errors-p
+       (message "error: %s" (plist-get res :message)))
```
gebner: This should only happen with protocol errors, i.e., malformed messages or unknown commands.  I don't think we should hide them.  Did you encounter this kind of error in a different situation?

david-christiansen: Yes, the hole commands were reporting errors when run outside of a hole. I can have @leodemoura change it so they don't today.

src/emacs/lean-info.el
david-christiansen: Hmm. An issue with unconditionally showing "Find Definition" is that many regions don't support that. I suppose that this is also going to be an issue for holes, right?

Perhaps the best way forward is to have some regexp-based heuristics for deciding whether the click was in a hole or on an identifier, and then make the appropriate requests. I'll take a look at it today and revise the PR.

david-christiansen: After some more thought, I think that it would instead be best to just have a timeout. The same trick that's used to implement the synchronous Lean commands can be used to fire off a few requests for menu items, and if they don't complete in under .1 seconds, then they get ignored.

I'll try this approach.

gebner: > An issue with unconditionally showing "Find Definition" is that many regions don't support that. I suppose that this is also going to be an issue for holes, right?

That is true, right.  However e.g. vscode and intellij always show all go-to-definition commands, so we wouldn't be in bad company.  I guess you could even make a usability argument in *favor* of always showing go-to-definition: then the rmb-menu is predictable, and doesn't change depending on where you click.  In addition, the functionality also becomes discoverable: you can click anywhere, and get a list of all supported features.

> regexp-based heuristics

This is going to be flaky.  For example, an identifier like `generalizing` could be a constant (go-to-definition is possible), or a special token (where you can't use go-to-definition).  (If you import the standard library, it's a special token.  But it's not built-in.)

> if they don't complete in under .1 seconds, then they get ignored.

If we want to hide the menu items, then I think this is the best option.

src/emacs/lean-server.el
gebner: > Yes, the hole commands were reporting errors when run outside of a hole.

No, that's fine.  But the errors should cause the error callback to be called (the last argument to `lean-server-send-command`).  This line should only be executed for errors that are not related to a request.

gebner: Ok, now I see the problem as well.  We always call `(message)` with the error, even if there is an error callback installed.  I think we should only call `(message)` if there is no error callback.

src/emacs/lean-info.el
david-christiansen: I agree that regexen are a bad solution here, and am working on the timeout version.

The issue with always showing all things in the right-click menu is that many of them are being dynamically computed by Lean metaprograms right now, such as the hole commands. So this means that a solution for asking the server is necessary.

Emacs should do better than VS Code where possible :-) The whole point of a contextual menu is that it gives the commands that are relevant to some particular context. After all, what's the good of knowing that a "Find definition" command exists that will throw an error?

gebner: > Emacs should do better than VS Code where possible :-)

Nothing beats friendly competition.  I hope emacs catches up, in particular with auto-completion and  elisp.  It always feels a bit like going back to the stone age, with no real module system, only manual namespacing, and no type-checker.

> [...] hole commands. [...] what's the good of knowing that a "Find definition" command exists that will throw an error?

Speaking of invalid commands, we currently show all hole commands for every hole--even if they wouldn't work.  Ideally I'd also like to have at most one replacement for every valid hole command.  That is, I think we should change `hole_command` to:
```lean
meta def hole_command :=
list pexpr -> tactic (list (string × tactic (option string)))
```
The nested `tactic string` is there to allow lazy computation of a replacement (in case it's expensive).  When executing a hole command in the editor, we'd first compute the list of possible replacements.  Then the user makes a single choice, and we evaluate the `tactic (option string)` (which could fail, I just want to get rid of the "obviously" impossible actions).

david-christiansen: The current interface is intended to support commands that give a list of options. For instance, in Agda, calling Agsy with the `-l` option will have it give back a collection of type-correct inhabitants.

It does make sense for the hole commands to be able to say "I can't be used here" though.

david-christiansen:
This latest version has the following changes relative to the one reviewed by @gebner:
1. Rebased onto a Lean master that doesn't emit spurious errors, and the error print flag is removed again
2. The "find definition" menu entry also states the name
3. Menu entries are collected synchronously, but on a 150ms timeout to prevent noticable lag

Thanks for the feedback!

PR #1673, opened 2017-06-15
fix(frontends/lean/info_manager): do not store holes in triplicate

gebner:
Just making sure this doesn't escalate into an edit war.


leodemoura:
I just pushed  e99786afcd105921b80350

PR #1677, opened 2017-06-17
Add user-defined commands and use them to define `coinductive` in Lean

Kha:
The move itself is of dubious value, but the newly exposed primitives should help with defining other inductive-like commands and of course user-defined commands in general.

### DIFF
--- a/library/init/meta/coinductive_predicates.lean
+++ b/library/init/meta/coinductive_predicates.lean
@@ -114,8 +114,8 @@ end tactic
 section
 universe u
 
+@[user_attribute]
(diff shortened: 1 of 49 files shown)


Kha:
We could/should? of course move `coinductive` out of `init` now.

leodemoura:
@kha I will merge this PR after we release 3.2.0

Kha:
@leodemoura Sure, there's no rush. Also reminds me to update the changelog.

Kha:
I didn't actively plan for it, but overriding existing commands seems to work. So one could define just the bare command needed for the basic tactic framework in C++ and then implement the full thing in Lean.

PR #1689, opened 2017-06-21
fix(tests/lean/interactive/mk_input): strip \r from input files (win)

digama0:
This causes tests like the following to fail on windows:

    example :=
    true

out:

    {"message":"parse error - unexpected '\"'","response":"error"}

expected:

    {"message":"file invalidated","response":"ok","seq_num":0}


gebner:
To explain: git on windows has two modes.  It can either 1) convert LF line endings to CRLF, and 2) keep the line endings as they are.

The server tests works fine if the line endings are unchanged (i.e. LF).  This PR addresses the case where the line endings are changed on checkout.

PR #1690, opened 2017-06-21
fix(frontends/lean/elaborator): @applications don't make thunks

digama0:
If `f : thunk A -> B` and `g : thunk A`, then `f g` will automatically apply thunk parsing to create the ill-typed term `f (λ_, g)`. With this PR, you can now write `@f g` in order to disable thunk parsing.

### DIFF
--- a/library/data/lazy_list.lean
+++ b/library/data/lazy_list.lean
@@ -29,7 +29,7 @@ def tail : lazy_list α → lazy_list α
 
 def append : lazy_list α → thunk (lazy_list α) → lazy_list α
 | nil        l  := l ()
-| (cons h t) l  := cons h (append (t ()) (l ()))
+| (cons h t) l  := cons h (@append (t ()) l)
 
 def map (f : α → β) : lazy_list α → lazy_list β
 | nil        := nil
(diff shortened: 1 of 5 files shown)


leodemoura:
This PR is broken. Lean failed to compile the standard library. You should be able to reproduce the problem by executing the following commands in your build directory.
```
make clean-olean
make
```

digama0:
Okay, this is baffling. I only get one breakage in the standard library, at `lazy_list.append`, but this doesn't make any sense since this was exactly my test case when I wrote it!

    def append : lazy_list α → thunk (lazy_list α) → lazy_list α 
    | nil        l  := l () 
    | (cons h t) l  := cons h (@append (t ()) l)   -- The error message sounds like it is still making a thunk here

Is this what breaks for you? The test machines are showing breakage across the board, most likely because a different elaboration strategy is chosen, but this doesn't make any sense, since my change is trivial and explicitly only matters when (a) there is a thunk and (b) the `@` explicit parameter is used. Help!

leodemoura:
I didn't try to analyze the problem. The behavior is identical to the one in the cloud build:
https://travis-ci.org/leanprover/lean/jobs/245256729

Kha:
@digama0 https://github.com/Kha/lean/commit/c22e38d fixes the build for me. I can't figure out how to push to this PR right now.

leodemoura:
Merged with @kha's fix.

PR #1699, opened 2017-06-23
Minor changes to hash_map.lean

dselsam:
The first commit removes an unused argument to a lemma that prevents the simplifier from using it, and the second adds a basic lemma that no element is contained in the empty hash map.

### DIFF
--- a/library/data/hash_map.lean
+++ b/library/data/hash_map.lean
@@ -476,7 +476,7 @@ theorem find_iff (m : hash_map α β) (a : α) (b : β a) :
   m.find a = some b ↔ sigma.mk a b ∈ m.entries :=
 m.is_valid.find_aux_iff _ _ _
 
-theorem contains_iff (m : hash_map α β) (a : α) (b : β a) :
+theorem contains_iff (m : hash_map α β) (a : α) :
   m.contains a ↔ a ∈ m.keys :=
 m.is_valid.contains_aux_iff _ _
 


library/data/hash_map.lean
```diff
@@ -476,10 +476,14 @@ theorem find_iff (m : hash_map α β) (a : α) (b : β a) :
...
 
-theorem contains_iff (m : hash_map α β) (a : α) (b : β a) :
+theorem contains_iff (m : hash_map α β) (a : α) :
   m.contains a ↔ a ∈ m.keys :=
 m.is_valid.contains_aux_iff _ _
 
+lemma not_mem_mk_hash_map (hash_fn : α → nat) (nbuckets := 8) (a : α) :
+  a ∉ (mk_hash_map hash_fn nbuckets : hash_map α β) :=
+dec_trivial
+
```
leodemoura: Is the default value 8 intended? It is in the middle of two explicit ones. So, it will only fire in partial apps where only the first explicit argument is provided.

dselsam: No, that was an error. Good catch. I just amended the commit.

dselsam: Actually, perhaps it would be easier to use manually if both `hash_fn` and `nbuckets` were implicit. I'll push the change.

digama0:
These theorems are already in the backlog PR:

https://github.com/leanprover/lean/pull/1696/commits/2cc75c7bdae44d6ab0de9e1457eff301454448cc

dselsam:
@digama0 Thanks for pointing that out. I didn't think to look there. I removed the commit that added the lemma (but kept the commit that removes the unused argument).

Issue #1674, opened 2017-06-16
Parser refactoring + Hygienic macro system


leodemoura:
@kha @gebner: we have been talking about the parser refactoring for a few months.
@david-christiansen has many ideas and suggestions. I'm open this RFC to record his ideas and collect suggestions.

David is suggesting an architecture where we would have the following main steps:
1- `read : string -> syntax`
2- `expand : syntax -> pre-expr`
3- `elaborate : pre-expr -> expr`
The `read` procedure would be much simpler than the current lean parser, and it would generate a s-expression like data structure (`syntax`). It would include precise position information and comments.
Here are some links that David sent me: https://www.cs.indiana.edu/~dyb/pubs/LaSC-5-4-pp295-326.pdf, https://www.cs.utah.edu/plt/publications/popl16-f.pdf
 We would be able to reconstruct the input `string` faithfully from a `syntax` object produced by the `read` procedure. The `syntax` object will also represent faithfully constructs such as `match`-exprs, `do`-blocks, `calc`-blocks, etc which are currently expanded at parsing time. Names would *not* be resolved by the `read` procedure. It is not clear to us whether we should combine the `expand` and `elaborate` steps or not. It seems it would be a good idea to have both of them at least initially. In this way, the elaborator code would not be affected. The downside of not combining them is that expansion cannot take type information into account.

Users would be able to define new syntax using Lean. David also suggested that we could even have different `read` procedures like Racket.  Here is a paper describing how this works in Racket: http://users.eecs.northwestern.edu/~stamourv/papers/languages-the-racket-way.pdf

An important aspect of this proposal is that the expansion step is implemented in Lean, which means expanding user notation become far more flexible, and we can define new syntax based on old syntax. 
In the current implementation, new syntax must be mapped into pre-terms. So, it is non-trivial to add new syntax that extends existing syntax.

This approach is also useful for implementing tools such as formatters and linters using Lean. 
I'm not sure now it is best time for implementing this change, but I believe it would simplify the current implementation and make it much more flexible.

gebner:
I think this is a good approach.  ASTs that are close to the source code are pretty popular these days.  Scala's meta-programming framework also recently switched to it (scala.meta).

> `read : string -> syntax`

What strings does `read` read (i.e, whole files or only parts)? How do we deal with infix operators or user-defined syntax?
```lean
infix ` # ` := list.diff
#eval [1,2,3] # [2,3,4]

/- declare coinductive command (see Sebastian's branch) -/
[...]

PR #1701, opened 2017-06-25
fix(frontends/smt2,frontends/lean): handle EOF as int in scanner

moritayasuaki:
Handle EOF as int in scanner


leodemoura:
@gebner Do we need this PR? You have addressed the EOF issue, right?

gebner:
> @gebner Do we need this PR? You have addressed the EOF issue, right?

Yeah, I addressed the EOF issue already and we can build on ARM now.  Besides, the scanner will probably change quite a bit with #1674 anyhow.

Issue #1523, opened 2017-04-14
[RFC] Minimal Standard Library


jroesch:
I propose that we remove everything but the basics from the standard library. 

Many other languages have taken this approach for a variety of reasons. Most importantly it allows libraries to evolve uncoupled from language release cycles, reduces the maintenance burden on the core team, and keeps us free from standard library bloat that has happened in older languages without easy package distribution.

I am suggesting this change now because @gebner has done an awesome job prototyping a package management tool that we should start to dog food ASAP. 

I feel that these problems are exacerbated in a proof language as there is a desire to have as many useful facts as possible at your fingertips. It seems to lead to incoherence in the standard libraries and inconsistent quality control as evidenced by Coq's library which has many half finished libraries.

I propose that we convert `library_dev` to a contrib style development where we curate awesome developments in the community or incubate standard library extensions. We might also want to ship a batteries/haskell platform/stdx style package which includes a good standard set of packages.

Concrete of examples of what could be removed from the standard library today:
- smt2 bindings
- super
- mini-crush

Since the library is already pretty minimal, my purpose is to argue new additions should go in packages and not in the standard library.

A pointer to Rust's original proposal to do the same: https://github.com/rust-lang/rfcs/blob/master/text/1242-rust-lang-crates.md.

leodemoura:
I strongly support this RFC.

avigad:
I support it too, but we should talk about how we intend to "curate awesome developments in the community or incubate standard library extensions." Right now `library_dev` is an awkward mix of things: a staging ground for the standard library, a place for experimentation, a staging ground for more specialized theories, etc. 

Here are some different kinds of libraries:
- The standard library. This ships with Lean. It should be small, carefully curated, and we want to keep incompatible changes minimal across releases.
- Individual specialized libraries. These are easy. Any user (or developer) can create a library, try to make it useful, and take responsibility for maintaining it in their own space.
- Project-managed specialized libraries. For example, I think the project should support a good analysis library, which does not have to be part of the standard library but should be carefully maintained and kept up to date. Similarly for other branches of mathematics, program verification, etc.  
- Experimental libraries. We may (or may not) want to support spaces for shared experimentation, with bugs, sorry's, parts of the libraries that don't compile, and no promise or expectation of stability at all.

@jroesch How do Rust and other programming languages manage these different aspects of library development? Here is one way we can go about it: 
[...]

PR #1704, opened 2017-06-27
generalize_proofs tactic

digama0:
I thought I PR'd this a while ago, but I can't find it in the history... This tactic will generalize or merge to existing local constants any proofs within the goal. Not much configuration is available, just setting the names of generated hypotheses. I'm thinking about using a special notation where the name would go like `.` or something to indicate that the proof at this position should not be generalized, but even as is this is a very useful tactic for cleaning up the goal and also fixing some dependent rewrite issues.

### DIFF
--- a/library/init/meta/generalize_proofs.lean
+++ b/library/init/meta/generalize_proofs.lean
@@ -10,37 +10,47 @@ import init.meta.tactic
 namespace tactic
 
 private meta def collect_proofs_in :
-  expr → list name × list expr → tactic (list name × list expr) | e (ns, hs) :=
-let go : tactic (list name × list expr) :=
-(do t ← infer_type e,
-    is_prop t >>= guardb,
-    first (hs.map $ λ h, do
-      t' ← infer_type h,
-      is_def_eq t t',
-      g ← target,
-      change $ g.replace (λ a n, if a = e then some h else none),
-      return (ns, hs)) <|>
-    (let (n, ns) := (match ns with
-       | [] := (`_x, [])
-       | (n :: ns) := (n, ns)
-       end : name × list name) in
-     do generalize e n,
-        h ← intro n,
-        return (ns, h::hs))) <|> return (ns, hs) in
+  expr → list expr → list name × list expr → tactic (list name × list expr)
+| e ctx (ns, hs) :=
+let go (tac : list name × list expr → tactic (list name × list expr)) :
+  tactic (list name × list expr) :=
+do t ← infer_type e,
+   mcond (is_prop t) (do
+     first (hs.map $ λ h, do
+       t' ← infer_type h,
+       is_def_eq t t',
+       g ← target,
+       change $ g.replace (λ a n, if a = e then some h else none),
+       return (ns, hs)) <|>
+     (let (n, ns) := (match ns with
+        | [] := (`_x, [])
+        | (n :: ns) := (n, ns)
+        end : name × list name) in
+      do generalize e n,
+         h ← intro n,
+         return (ns, h::hs)) <|> return (ns, hs)) (tac (ns, hs)) in
 match e with
-| (expr.const _ _)   := go
-| (expr.local_const _ _ _ t) := collect_proofs_in t (ns, hs)
-| (expr.mvar _ t)    := collect_proofs_in t (ns, hs)
+| (expr.const _ _)   := go return
+| (expr.local_const _ _ _ t) := collect_proofs_in t ctx (ns, hs)
+| (expr.mvar _ t)    := collect_proofs_in t ctx (ns, hs)
 | (expr.app f x)     :=
-  go >>= collect_proofs_in f >>= collect_proofs_in x
+  go (λ nh, collect_proofs_in f ctx nh >>= collect_proofs_in x ctx)
 | (expr.lam n b d e) :=
-  go >>= collect_proofs_in d >>= collect_proofs_in e
-| (expr.pi n b d e) :=
-  collect_proofs_in d (ns, hs) >>= collect_proofs_in e
+  go (λ nh, do
+    nh ← collect_proofs_in d ctx nh,
+    var ← mk_local' n b d,
+    collect_proofs_in (expr.instantiate_var e var) (var::ctx) nh)
+| (expr.pi n b d e) := do
+  nh ← collect_proofs_in d ctx (ns, hs),
+  var ← mk_local' n b d,
+  collect_proofs_in (expr.instantiate_var e var) (var::ctx) nh
 | (expr.elet n t d e) :=
-  go >>= collect_proofs_in t >>= collect_proofs_in d >>= collect_proofs_in e
+  go (λ nh, do
+    nh ← collect_proofs_in t ctx nh,
+    nh ← collect_proofs_in d ctx nh,
+    collect_proofs_in (expr.instantiate_var e d) ctx nh)
 | (expr.macro m l) :=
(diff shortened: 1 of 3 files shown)


gebner:
This tactic has slightly unexpected behavior for proof terms under binders:
```lean
example : ∃ n (x : vector ℕ n) (y : vector ℕ (1+n)),
  cast (by simp) (5 :: x) = y :=
begin generalize_proofs [] end
```
produces:
```lean
state:
_x : ∀ {α β : Prop}, α = β → β → α,
_x : ∀ (α : Prop), α → α,
_x : ∀ {a b c : Prop}, a = b → b = c → a = c,
_x : ∀ (a a_1 : Type), a = a_1 → ∀ (a_2 a_3 : Type), a_2 = a_3 → a = a_2 = (a_1 = a_3),
_x : ∀ {β : Type 1} {a₁ a₂ : Type} (f : Type → β), a₁ = a₂ → f a₁ = f a₂,
_x : ∀ (a : Type), a = a,
_x : ∀ (n n_1 : ℕ), n = n_1 → vector ℕ n = vector ℕ n_1,
_x : ∀ (b : ℕ), 1 + b = b + 1,
_x : ∀ {a b : Prop}, (a ↔ b) → a = b,
_x : ∀ (a : Type), a = a ↔ true,
_x : true
⊢ ∃ (n : ℕ) (x : vector ℕ n) (y : vector ℕ (1 + n)),
    cast
        (_x
           (_x (vector ℕ (nat.succ n) = vector ℕ (1 + n) = true)
              (_x
                 (_x (vector ℕ (nat.succ n)) (vector ℕ (nat.succ n)) (_x (vector ℕ (nat.succ n)))
                    (vector ℕ (1 + n))
                    (vector ℕ (n + 1))
                    (_x (1 + n) (n + 1) (_x n)))
                 (_x (_x (vector ℕ (nat.succ n))))))
           _x)
        (5 :: x) =
      y
```

digama0:
Yeah, that's a bug. It turns out that most of the C++ typechecking work is very difficult to replicate in Lean, I'm doing a lot by hand that is going to cost performance, I think. The issue I ran into while fixing this is that I can't use `tactic.generalize` to generalize under a binder (where the generalized variable would have a Pi type). Any suggestions, or should I just reimplement this in C++?

digama0:
I pushed a fix, but without dependent generalize all I can do is fail (although the tactic as a whole doesn't fail, it just doesn't do any replacements). Now your example comes back unchanged, and you have to instantiate the quantifiers before you can generalize the proofs.

leodemoura:
> Any suggestions, or should I just reimplement this in C++?

@digama0 As I said in the slack channel (or in another place, I don't remember). We have code in C++ for writing an efficient version of this tactic. I think this is not a super urgent feature. If you need it for a project, I suggest you keep it in the project.

digama0:
@leodemoura This is the sort of tactic that, once you use it, you want to use it everywhere - its use cases are very broad, and there are times when nothing else is nearly as effective. So "I need it for a project" doesn't quite cover it.

I'm not sure I understand what you mean by "We have code in C++ for writing an efficient version of this tactic." Do you mean that this is already implemented somewhere? Or that a feature like this ought to be implemented in C++? Or that there are tools that would make it easier to write these sorts of tactics?

Would you object if I tried my hand at implementing this in C++ myself? I am not looking for you to drop your work and work on this, but if I reimplemented it myself, would it be merged?

digama0:
Maybe the C++ code for writing expr traversal tactics should be exposed as a meta constant. It is bothersome that there are no expr traversal functions that produce closed terms (so that `infer_type` works on the subexpressions).

gebner:
> @leodemoura This is the sort of tactic that, once you use it, you want to use it everywhere 

I agree.  Just now @javra asked about it on Slack as well.

I think the C++ issue is a bit of a red herring.  The tactic works fine as it is.  I'd vote that we merge it, and consider performance optimizations when people complain.  We can discuss better expression traversals in another issue.

> It is bothersome that there are no expr traversal functions that produce closed terms

Actually it's easy (but unergonomic) if you know locally nameless.  Whenever you traverse a binder, you need to produce a fresh local constant, and instantiate the term with it.  See the `trim` function in super for example: https://github.com/leanprover/super/blob/85649de2c6ee42f84fdd147a00d15a094410ab58/src/super/trim.lean#L19-L22

javra:
Am I mistaken or did the lean2 `induction` tactic do bits of generalization automatically, that the one in lean3 doesn't?

leodemoura:
> I agree. Just now @javra asked about it on Slack as well.
I think the C++ issue is a bit of a red herring. The tactic works fine as it is. I'd vote that we merge it, and consider performance optimizations when people complain. We can discuss better expression traversals in another issue.

@digama0 @gebner For me, performance is a big deal. I spend a lot of time profiling and trying to improve performance. We already have code for performing traversals (`dsimplify` and `simplify` modules) and canonizing proofs and type class instances (`defeq_canonizer` module) efficiently in C++. This PR doesn't use them. Moreover, there are clear performance issues with this PR. Here are two of them: it traverses the expression as a tree (potential exponential blowup for goals with a lot of shared terms like Certigrad goals);  it doesn't handle `lambda/pi` as blocks. It might have more issues, but I didn't review the code carefully. 
Another minor issue: the PR does not have any tests.
I don't want to merge the PR, and then have to deal with the performance and maintenance problems myself. We already have a ton of problems, and I spend most of my time fixing bugs and improving performance. I'm trying to minimize what goes in the standard library. 
We have a nice package system now. I suggest we put this tactic at `library_dev` or in another repo. 

> its use cases are very broad, and there are times when nothing else is nearly as effective. So "I need it for a project" doesn't quite cover it.

@digama0  I'm strong believer of #1523. Most new developments should go into `library_dev`. So, we will not need this tactic right now in this repo. I think it is pretty reasonable to put it at `library_dev`.

I'm strongly against a big library in this repo. We used  "the big/fat library" approach in Lean 0.2 and 0.1. It was a nightmare for me. I was spending most of my time fixing proofs (several of them were a big spaghetti, and were not written by me). Lean is changing a lot, and I don't want to go back to the Lean 0.2 days. I was literally going insane.

> Would you object if I tried my hand at implementing this in C++ myself? I am not looking for you to drop your work and work on this, but if I reimplemented it myself, would it be merged?

I can go over the code with you when we meet in Cambridge.

digama0:
@leodemoura (I agree with you about all the performance points, and I'm sure that a proper implementation of this tactic requires a complete rewrite, so just consider this a proof of concept for now.)

> @digama0 I'm strong believer of #1523. Most new developments should go into library_dev. So, we will not need this tactic right now in this repo. I think it is pretty reasonable to put it at library_dev.

As a matter of fact, I am also very much in favor of #1523, but I was under the impression that it hadn't landed yet. There are still huge chunks of this repo that can be cleaved off into another library without breaking the C++ build - I would summarize it as "Lean programming constructions & tools" which includes all the things that make Lean easy to use as a programming language. `hash_map` is an easy one to point to, but I would argue even things like `list` (not `list` itself but everything else in `list.basic`). At minimum, we need all the VM extension points to be defined in this repo, but literally nothing else outside of `init.core` is truly necessary. Most of the tactic framework is already checked for existence first because this is needed anyway in the `prelude`.

If the `lean` repo is split along these lines, then of course my additions will be going into the stdlib part, not the lean core part (unless this tactic is rewritten in C++), but until then these two types of contribution are not distinguished.

gebner:
Please put this in stdlib for now.

PR #1706, opened 2017-06-27
`assume` and `suppose` tactics

Kha:
`assume` works the same as in term mode, except without anonymous constructor notation (would depend on #1483). `suppose` uses the same syntax as `have`/`let` without the value part, `suppose [id] [: type]`.

### DIFF
--- a/library/init/meta/expr.lean
+++ b/library/init/meta/expr.lean
@@ -247,11 +247,11 @@ meta def local_uniq_name : expr → name
 | (local_const n m bi t) := n
 | e                      := name.anonymous
 
-meta def local_pp_name : expr → name
+meta def local_pp_name : expr elab → name
 | (local_const x n bi t) := n
 | e                      := name.anonymous
 
-meta def local_type : expr → expr
+meta def local_type : expr elab → expr elab
 | (local_const _ _ _ t) := t
 | e := e
 
--- a/library/init/meta/interactive.lean
+++ b/library/init/meta/interactive.lean
@@ -450,6 +450,26 @@ tactic.any_goals
 meta def focus (tac : itactic) : tactic unit :=
 tactic.focus [tac]
 
+meta def assume_tac (p : parse parse_binders) : tactic unit :=
+list.mfor' p $ λ b,
+  do t ← target,
+     when (not $ t.is_pi ∨ t.is_let) whnf_target,
+     when (not $ t.is_pi ∨ t.is_let) $
+       fail "assume tactic failed, Pi/let expression expected",
+     ty ← i_to_expr b.local_type,
+     unify ty t.binding_domain,
+     intro_core b.local_pp_name
+
+meta def suppose_tac (h : parse ident?) (q : parse (tk ":" *> texpr)?) : tactic unit :=
+let h := h.get_or_else `this in
+match q with
+| some e := do
+  uniq_name ← mk_fresh_name,
+  let l := expr.local_const uniq_name h binder_info.default e,
+  assume_tac [l]
+| none := intro h
+end
+
 /--
 This tactic applies to any goal.
 - `have h : T := p` adds the hypothesis `h : T` to the current goal if `p` a term of type `T`.
--- a/library/init/meta/interactive_base.lean
+++ b/library/init/meta/interactive_base.lean
@@ -142,6 +142,9 @@ meta def param_desc : expr → tactic format
   else paren <$> pp e
 
 
+private meta constant parse_binders_core (rbp : ℕ) : parser (list pexpr)
+meta def parse_binders (rbp := std.prec.max) := with_desc "<binders>" (parse_binders_core rbp)
+
 meta constant decl_attributes : Type
 
 meta constant decl_attributes.apply : decl_attributes → name → parser unit
--- a/library/init/meta/smt/interactive.lean
+++ b/library/init/meta/smt/interactive.lean
@@ -77,6 +77,26 @@ tactic.interactive.change q none (loc.ns [])
 meta def exact (q : parse texpr) : smt_tactic unit :=
 tactic.interactive.exact q
 
+meta def assume_tac (p : parse parse_binders) : smt_tactic unit :=
+list.mfor' p $ λ b,
+  do t ← tactic.target,
+     when (not $ t.is_pi ∨ t.is_let) tactic.whnf_target,
+     when (not $ t.is_pi ∨ t.is_let) $
+       fail "assume tactic failed, Pi/let expression expected",
+     ty ← tactic.i_to_expr b.local_type,
+     tactic.unify ty t.binding_domain,
+     smt_tactic.intro_lst [b.local_pp_name]
+
+meta def suppose_tac (h : parse ident?) (q : parse (tk ":" *> texpr)?) : smt_tactic unit :=
+let h := h.get_or_else `this in
+match q with
+| some e := do
+  uniq_name ← tactic.mk_fresh_name,
+  let l := expr.local_const uniq_name h binder_info.default e,
+  assume_tac [l]
+| none := smt_tactic.intro_lst [h]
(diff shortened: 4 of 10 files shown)


Kha:
There is the open question of whether to consolidate these tactics and `intro`/`intros`. We'd probably want to keep the parameter-less versions of `intro`/`intros`.

leodemoura:
@Kha In term mode, `fun`, `take` and `assume` are identical, `suppose` provides some extra functionality (anonymous binder). These variants were added in Lean 0.1, and were inspired by Mizar and Isar. In retrospect, I think this was a bad idea. In particular, I think it is really bad that `take` is a keyword. Should we remove all these variants and consolidate?

leodemoura:
BTW, I will merge this PR as is. The comment above is not really about the PR.

avigad:
`take` is nice for teaching students how to write proofs, and for that purpose, lambda, fun, and assume are not very natural as substitutes. So, I will be sorry to see it go. That said, I can live without it.

Kha:
> lambda, fun, and assume are not very natural as substitutes

I agree; I don't mind `take` very much. If we keep it, we should also add it as a tactic (as a synonym for `assume_tac`).

leodemoura:
> I agree; I don't mind take very much. If we keep it, we should also add it as a tactic (as a synonym for assume_tac).

@kha I didn't mind it until you made the "Taken" joke ;)

Kha:
@leodemoura Hey, `list.taken` is a whole nother (non-)issue :smile:

avigad:
@kha, the point is well `taken`.

leodemoura:
> @leodemoura Hey, list.taken is a whole nother (non-)issue 😄
> @Kha, the point is well taken.

@avigad @Kha I'm pretty sure whoever wrote/named this function, first tried to name it `list.take` since `take` is the name used in other programming languages. Moreover, `take` is useful for naming functions in many other modules (e.g., `stream`, `lazy_list`, etc). I think we have no good reason for sacrificing `take` (i.e., marking it as a keyword). `take` is only useful for very novice users. Note that, it is barely used in the standard library. It is mostly used in old files inherited from Lean 0.1. 
For teaching, we can have a separate file that declares the `take` keyword and notation.

avigad:
I agree.

Can user-defined commands currently be used to declare `take` as an alternative to `assume`? Or will it be possible in the future? If so, that's a good solution for me. I can also teach students to use lambda or `assume` -- it is not hard to get used to.

leodemoura:
> Can user-defined commands currently be used to declare take as an alternative to assume? 

Yes, we can.
```lean
notation `Take` binders `, ` r:(scoped P, P) := r
#check Take x : nat, eq.refl x
```

gebner:
> I'm pretty sure whoever wrote/named this function, first tried to name it list.take

That would have been me, and this was indeed the reason.  We also have `list.dropn` for consistency.  If we remove the `take` keyword, we should probably rename both of them.

Using work-around names because the real ones are taken by keywords doesn't bother me too much.  Every language with keywords has this problem, just think of `klass`, `clazz`, `typ`, etc.

Kha:
I think I would be fine with just calling it `list.take` and making it `protected`. Using dot-notation for list functions is more ergonomic than opening the namespace anyway.

leodemoura:
> Using work-around names because the real ones are taken by keywords doesn't bother me too much. Every language with keywords has this problem, just think of klass, clazz, typ, etc.

@gebner I agree with your point when the keyword is useful (e.g., `class`). This is not the case for `take`, it is just an alias that doesn't add any functionality, and it is barely used in the standard library. As I said in the previous comment, it only occurs in files that were originally written for Lean 0.1 (e.g., `set`).

> I think I would be fine with just calling it list.take and making it protected. Using dot-notation for list functions is more ergonomic than opening the namespace anyway.

@Kha This is quite hackish, and we cannot define it in the namespace `list` (like `taken`).
https://github.com/leanprover/lean/blob/master/library/init/data/list/basic.lean#L190-L193
We would have to move it to the top-level and define it using `protected def list.take ...`.

avigad:
The more I think about it, the more I become convinced that it will no problem to teach students to use e.g. `assume n : ℕ`, I am starting to like that even more than `take`.

@Kha  the following proof works:
```lean
example (p q : Prop) : p ∧ q ↔ q ∧ p :=
begin
  apply iff.intro,
  { assume h,
    have hp : p := h.left,
    have hq : q := h.right,
    show q ∧ p,
     exact ⟨hq, hp⟩ },
  assume h,
  have hp : p := h.right,
  have hq : q := h.left,
  show p w,
    exact ⟨hp, hq⟩
end
```
But writing `assume h : p ∧ q` in the first instance gives errors, `unknown identifier 'p'`. Shouldn't that work?

Also, note that if we adopted a similar convention for `show`, we could write `show q∧ p := ⟨hq, hp⟩`, avoiding the need for the explicit `exact` tactic. Would that we hard to do, and do you think it is a good idea?

avigad:
@Kha is would also be nice to be able to use `assume` with negation:
```lean
-- works
example (p q : Prop) (h : p → q) : ¬ q → ¬ p :=
assume hnq hp,
hnq (h hp)

-- works
example (p q : Prop) (h : p → q) : ¬ q → ¬ p :=
begin
  intros hnq hp,
  exact hnq (h hp)
end

-- doesn't
example (p q : Prop) (h : p → q) : ¬ q → ¬ p :=
begin
  assume hnq hp,
  exact hnq (h hp)
end
```

Kha:
So, the proposed removals are:
* `take` ~> `assume`
* `suppose` ~> `assume:` (this would be a new anonymous variant of `assume`)
I'm fine with both of these.

Kha:
@avigad
> Also, note that if we adopted a similar convention for show, we could write show q∧ p := ⟨hq, hp⟩, avoiding the need for the explicit exact tactic. Would that we hard to do, and do you think it is a good idea?

We should definitely be consistent with the `have` tactic, but now I'm torn on the syntax in general - for both `have` and `show`, `, from` reads much more naturally to me. I just adopted the syntax from the `define` tactic, but there really is no reason to break consistency with the term-mode keywords here. What do you think?

digama0:
Here's a crazy idea: rename the `exact` tactic to `from`. Then `have h : t, from p,` and `show t, from p` will parse as two tactics and give exactly the expected behavior (except that `$` can't be used in interactive mode, which is a bit of a pain).

In the other direction, we could modify the term mode keywords to accept `:=`, which I think is a nice way to unify the syntax of `have` and `let`. Unfortunately, it's quite confusing to use with `suffices`, since the terms are switched, and I don't really have a general solution.

Kha:
@digama0 
> rename the exact tactic to from

That would certainly be the easiest solution, I'm warming up to it. Though we should probably keep the old name as well.
Analogously, we could make the interactive `by tac` sugar for `focus { tac }`. Right now, it tries to build a meta meta program, which doesn't strike me as very useful.

digama0:
agreed on the meta meta thing. There is no way for the meta meta program to even see the original tactic state, so there are no gains over simply `my_tac : tactic unit := by meta_meta` where the context of the `by my_tac` application is removed. In other words, `by by tac` is definitely pointless, and making `by = focus` in interactive mode is as good as any.

By the way, is there a way to sequence tactics _without_ also focusing? This comes up in things like `cases h; { tac1, tac2 }` where I want to see the new goal state after application of the tactics, rather than erroring because `tac2` didn't close the goal. I think something like ``(`[ tac1, tac2])`` will work, but dropping into (tactic) term mode often messes up surrounding syntax.

avigad:
The proposed changes sound good to me. I never liked `suppose`. It bothered me to have such a different name for the essentially the same operation. But is `assume: ...` better than `assume : ...`, with an extra space? The latter is intuitive: it is `assume h : ...` without the `h`.

PR #1707, opened 2017-06-28
feat(leanpkg/main.lean): allow passing extra arguments to Lean

dselsam:
Note: there is some extra complexity because I wanted to distinguish `leanpkg-args` and `lean-args`.
If anyone who works on `leanpkg` would prefer not to, I can change it.

### DIFF
--- a/leanpkg/leanpkg/main.lean
+++ b/leanpkg/leanpkg/main.lean
@@ -45,22 +45,21 @@ assg ← solve_deps d,
 path_file_cnts ← mk_path_file <$> construct_path assg,
 write_file "leanpkg.path" path_file_cnts
 
-
-def make : io unit := do
+def make (lean_args : list string) : io unit := do
 manifest ← read_manifest,
 exec_cmd {
   cmd := "lean",
   args := (match manifest.timeout with some t := ["-T", repr t] | none := [] end) ++
-    ["--make"] ++ manifest.effective_path,
+    ["--make"] ++ manifest.effective_path ++ lean_args,
   env := [("LEAN_PATH", none)]
 }
 
-def build := configure >> make
+def build (lean_args : list string) := configure >> make lean_args
 
-def make_test : io unit :=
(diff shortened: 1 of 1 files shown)


leanpkg/leanpkg/main.lean
```diff
@@ -45,22 +45,21 @@ assg ← solve_deps d,
...
 exec_cmd {
   cmd := "lean",
   args := (match manifest.timeout with some t := ["-T", repr t] | none := [] end) ++
-    ["--make"] ++ manifest.effective_path,
+    ["--make"] ++ manifest.effective_path ++ lean_args,
   env := [("LEAN_PATH", none)]
 }
 
-def build := configure >> make
+def build (lean_args :list string) := configure >> make lean_args
```
gebner: The colon in the type ascription should be followed by a space.

leanpkg/leanpkg/main.lean
```diff
@@ -149,32 +148,32 @@ configure
 def usage := "
 Usage: leanpkg <command>
 
-configure       download dependencies
-build           download dependencies and build *.olean files
-test            download dependencies and run test files
+configure            download dependencies
+build -- <lean-args> download dependencies and build *.olean files
```
gebner: I think we should put brackets (the `[]` variety) around `-- <lean-args>` to indicate that the extra arguments are optional.

leanpkg/leanpkg/main.lean
dselsam: Thanks. Sorry I missed the space -- the `S-SPC` key-binding in Emacs eats some of my spaces (I change it to TAB on my branches).

leanpkg/leanpkg/main.lean
dselsam: Agreed.

gebner:
Now `leanpkg` fails without printing the usage message when called with no arguments:
```
$ leanpkg
split_cmdline_args: no args provided
```

dselsam:
> Now `leanpkg` fails without printing the usage message when called with no arguments:

Amended.

PR #1708, opened 2017-06-28
Escaped identifiers

Kha:
Add `«id»` notation that can be used to declare and refer to identifiers containing prohibited characters. For example, `a.«b.c»` is a two-part identifier with parts `a` and `b.c`.

### DIFF
--- a/src/frontends/lean/pp.cpp
+++ b/src/frontends/lean/pp.cpp
@@ -669,6 +669,10 @@ optional<name> pretty_fn::is_aliased(name const & n) const {
     }
 }
 
+static format escape(name const & n) {
+    return format(n.escape());
+}
+
 auto pretty_fn::pp_const(expr const & e, optional<unsigned> const & num_ref_univ_params) -> result {
     if (is_neutral_expr(e) && m_unicode)
         return format("◾");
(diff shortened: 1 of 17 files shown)


gebner:
Re: emacs input.  I didn't see any changes to input.  Am I missing something or is the commit message misleading?

Should we rename `expr.elet` to `expr.«let»` as well?  (I'm not sure I like it.)

Kha:
> Re: emacs input. I didn't see any changes to input. Am I missing something or is the commit message misleading?

Right, the keybindings vanished on rebasing because they were already there. Updated message.

> Should we rename expr.elet to expr.«let» as well? (I'm not sure I like it.)

I don't think I'd like that either for a name that is usually imported from its namespace (so that you indeed need to escape it). It's not even shorter, but many keystrokes longer.

gebner:
> Right, the keybindings vanished on rebasing because they were already there. Updated message.

I see, "Kha added some commits on Aug 25, 2016". :smile:

Let's not forget to port these changes to https://github.com/leanprover/lean-mode as well.

src/frontends/lean/pp.cpp
```diff
@@ -669,6 +669,10 @@ optional<name> pretty_fn::is_aliased(name const & n) const {
     }
 }
 
+static format escape(name const & n) {
+    return format(n.escape());
+}
```
leodemoura, line 674: It would be nice to handle keywords.
```lean
def «let» := 10
#check «let»
-- let : nat
```

PR #1712, opened 2017-06-30
dist: add leanpkg.bat script for invocation on Windows

emberian:


bin/leanpkg.bat
```diff
@@ -0,0 +1,8 @@
+@ECHO OFF
+
+SET LEANDIR=%~dp0%../
+SET LIBDIR=%LEANDIR%\lib\lean
```
gebner, line 4: This does not work when running `leanpkg.bat` directly from the git checkout, right?  The shell scripts all work in the git-checkout as well (without installation).

leodemoura:
@cmr: Should I close this PR? @gebner raised an issue above.

bin/leanpkg.bat
emberian: I guess not. I can fix it.

PR #1713, opened 2017-06-30
add leanpkg.bat

emberian:


src/CMakeLists.txt
```diff
@@ -477,9 +478,15 @@ add_custom_target(clean-std-lib
 add_custom_target(clean-olean
   DEPENDS clean-std-lib)
 
-install(FILES "${CMAKE_SOURCE_DIR}/../bin/leanpkg"
-  DESTINATION bin
-  PERMISSIONS OWNER_READ OWNER_WRITE OWNER_EXECUTE GROUP_READ GROUP_EXECUTE WORLD_READ WORLD_EXECUTE)
+if(${CMAKE_SYSTEM_NAME} MATCHES "Windows")
+  install(FILES "${CMAKE_SOURCE_DIR}/../bin/leanpkg.bat" "${CMAKE_SOURCE_DIR}/../bin/leanpkg.bat"
```
gebner: This installs only the `.bat` file and not the shell script on windows.  Does the `.bat` file work in the msys2 shell as well?

emberian: I just cargo culted the "leanemacs.bat" lines below.

I don't use the msys2 shell though, and the ZIP I download from github doesn't have the shell scripts.

emberian: (I'm still confused about the leanemacs.bat line, I'm pretty sure it's a mistake to specify the same path twice)

gebner: I don't think anybody actively tests the `.bat` files.  If you find problems with them, please fix (or report) them.

You can generate the release zip by running `make package` if you want to test it.

leodemoura: @gebner @cmr Should we close this PR?

emberian: I will fix it.

emberian:
@gebner this is ready for another review (it also combines that other PR, which had some changes leaked into it by accident))

gebner:
Thanks, all issues have been resolved.

PR #1715, opened 2017-06-30
doc(init/core): Document init_quotient in lean

digama0:
This way people can search for "constant quot" and find it in the lean source. Plus the init_quotient command only occurs once, so this way people know what it means.


gebner:
We should also make go-to-definition work with `quot`, `quot.mk`, etc.

gebner:
BTW, this PR has failing tests.

digama0:
How could this possibly fail a test?! Nothing has changed in the code. I suppose it's line numbers, is there a way to just fix the `produced` files to equal the `expected` files?

gebner:
> is there a way to just fix the produced files to equal the expected files?

You can copy the produced file over the expected file.  Just make sure that the line numbers are the only thing changing, and we're not breaking anything else.

Kha:
Using the diff tool:
```
tests/lean/interactive $ ./test_single.sh lean complete_field.lean yes
```
We should probably make these tests self-contained.

PR #1719, opened 2017-07-04
Interactive fixes

Kha:
See https://github.com/leanprover/lean/pull/1706#issuecomment-312423568 and following

### DIFF
--- a/src/emacs/lean-dev.el
+++ b/src/emacs/lean-dev.el
@@ -1,0 +1,16 @@
+;;  -*- lexical-binding: t -*-
+;;
+;; Copyright (c) 2017 Microsoft Corporation. All rights reserved.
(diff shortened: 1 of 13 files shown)


src/emacs/lean-dev.el
```diff
@@ -0,0 +1,14 @@
...
+;;
+;; Author: Sebastian Ullrich
+;;
+
+(defun lean-diff-test-file ()
+  "Use interactive ./test_input.sh on file of current buffer"
+  (interactive)
+  (message (shell-command-to-string (format "./test_single.sh \"%s\" \"%s\" yes"
+                                            (lean-get-executable "lean")
+                                            (buffer-name)))))
```
Kha, line 14: @leodemoura You may find this function helpful :)

leodemoura: Nice!

Issue #1518, opened 2017-04-13
VM error


gebner:
The following fails with `error: vm check failed: i < 2 (possibly due to incorrect axioms, or sorry)`:
```lean
def join (sep : string) : list string → string
| [x]     := x
| []      := ""
| (x::xs) := x ++ sep ++ join xs

namespace toml

inductive value : Type
| str : string → value
| bool : bool → value
| table : list (string × value) → value

namespace value

private def to_string_core : ∀ (n : ℕ), value → string
| _ (value.str s) := "\"" ++ s ++ "\""
| _ (value.bool tt) := "true"
| _ (value.bool ff) := "false"
| (n+1) (value.table cs) :=
  "{" ++ join ", " (do (k, v) ← cs, [k ++ " = " ++ to_string_core n v]) ++ "}"
| 0 _ := "<max recursion depth reached>"

protected def to_string : ∀ (v : value), string
| (table cs) := join "\n" $ do (h, c) ← cs,
  match c with
  | table ds :=
    ["[" ++ h ++ "]\n" ++
     join "" (do (k, v) ← ds,
       [k ++ " = " ++ to_string_core (sizeof v) v ++ "\n"])]
  | _ := ["<error> " ++ to_string_core (sizeof c) c]
  end
| v := to_string_core (sizeof v) v

instance : has_to_string value :=
⟨value.to_string⟩

#eval value.to_string (table [("foo", table [("bar", table [])])])

end value
end toml
```

leodemoura:
@dselsam I believe the problem is the `sizeof` instances of nested inductive datatypes.
They use the internal recursors generated by the inductive compiler.
Example:
```lean
@[irreducible]
protected def _nest_2_2.toml.value._mut_.sizeof : Π (a : psum unit (psum unit unit)), _nest_2_2.toml.value._mut_ a → ℕ :=
λ (a : psum unit (psum unit unit)),
  _nest_2_2.toml.value._mut_.rec (λ (a : string), 1 + sizeof a) (λ (a : bool), 1 + sizeof a)
    (λ (a : _nest_2_2.toml.value._mut_ (psum.inr (psum.inr ()))) (ih : ℕ), 1 + ih)
    (λ (fst : string) (snd : _nest_2_2.toml.value._mut_ (psum.inl ())) (ih : ℕ), 1 + sizeof fst + ih)
    1
    (λ (a : _nest_2_2.toml.value._mut_ (psum.inr (psum.inl ())))
[...]

PR #1720, opened 2017-07-04
Counter-examples for non-exhaustive matches

gebner:
For example:
```lean
def foo (y : ℕ) : ℤ → ℕ → ℕ
| x 0 := 3
| x (n+5) := 42
/-
non-exhaustive set of equations, the following cases are missing:
 (_) (1)
 (_) (2)
 (_) (3)
 (_) (4)
-/

def bar : bool → bool → bool
| ff _ := tt
| _ tt := ff
/-
non-exhaustive set of equations, the following cases are missing:
 (tt) (ff)
-/
```

The implementation works by assigning an *example* to each problem in the equation compiler, and propagating it through the various transitions.  The example is a sequence of terms of the same length as the original match variables, that is supposed to be matched using the current problem.  If we arrive at a leaf without equations, the example is a counter-example to the exhaustiveness of the original match.

This works reasonably well.  The main downside is that the nested inductives are expanded (since they are processed by unfolding + transport).  The other problem is that inexhaustive matches on strings generate sub-problems for every unmatched character, maybe we should just always do if-then-else here.

I originally tried to implement this feature by adding an extra wildcard equation at the end of the existing equations.  However this causes issues all over the place.  In `process_no_equation` we check whether there are no equations, in the if-then-else transition we check whether there are variable matches, etc.


leodemoura:
I will take a look and merge tomorrow.

leodemoura:
@gebner Another problem is that the equation compiler may apply transformations before `elim_match`. When this happens, we get weird error messages. Here are two examples:

```lean
/- Error message contains extra (_) for the brec_on functional
   (introduced by structural_rec.cpp) -/
def foo (y : ℕ) : ℕ → ℕ → ℕ
| 0     a     := a
| (x+1) 0     := 3 + foo x 10
| (x+1) (n+2) := 42


/- Error message:

non-exhaustive set of equations, the following cases are missing:
 (psum.inl ⟨0, nat.succ _⟩)
 (psum.inl ⟨nat.succ _, _⟩)
 (psum.inr ⟨0, nat.succ _⟩)
 (psum.inr ⟨nat.succ _, _⟩)

It reflects the packing steps introduced by pack_domain.cpp and pack_mutual.cpp.
-/
mutual def f, g
with f : nat → nat → nat
| 0 0 := 1
with g : nat → nat → nat
| 0 0 := 1
```

We need to "undo" these transformations.

digama0:
Maybe I'm overestimating the difficulty, but I think that cleaning up the error messages in the case of pack/unpack type transformations is not a major concern compared to the added value of having a counterexample to look at to begin with. I think that it is comparatively low priority, and I'd hate to see it block this PR.

gebner:
> Another problem is that the equation compiler may apply transformations before elim_match.

I see.  So we need to start a bit earlier in `compile_equations` and in some way pass along information about the original match.

With the additional wildcard equation approach, we could just add the equation in `compile_equation` and properly tag it so that we can identify it later in elim_match.  The tag would be a generalization of the current ignore_is_unused flag.

I'm still unhappy with the wildcard solution since it completely changes the behavior of the equations compiler.  There is no way around it, with the wildcard solution we at the very least always need to do a complete transition before every constructor transition.

Thinking about it, the additional complete transition does not seem to be a big problem to me (it's only a bit of overhead but no real change in behavior).  Maybe we can just ignore the additional fake equations in i-t-e and no-equation transitions (pattern matches in do-blocks also generate a wildcard match).  I can try that approach.

I don't see any nice way to pass along the information about the original match in the examples approach.

leodemoura:
> I see. So we need to start a bit earlier in compile_equations and in some way pass along information about the original match.

I was imagining a different solution:
- At `elim_match`, we throw an "non_exhaustive" exception. This exception contains structured information (with `expr`) instead of `format` objects.
- We catch this exception at `elim_match` call sites, and "undo" the transformations. Remark: we have similar code for "undoing" transformations in equational lemmas.
- Then, we convert the exception into a regular exception.

leodemoura:
> Maybe I'm overestimating the difficulty, but I think that cleaning up the error messages in the case of pack/unpack type transformations is not a major concern compared to the added value of having a counterexample to look at to begin with. I think that it is comparatively low priority, and I'd hate to see it block this PR.

@digama0 This is not acceptable. Users would keep reporting it as a bug.
Well, one option is to append the message: "WE KNOW THE ERROR MESSAGE IS CRYPTIC SOMETIMES, PLEASE DO **NOT** REPORT A BUG" :)

digama0:
@leodemoura I realize this. Still, cryptic error messages are not uncommon in Lean. I think it would be fine if we had an issue for this and left it open and low priority.

...But if addressing this is not difficult, don't let me stop you.

gebner:
> At elim_match, we throw an "non_exhaustive" exception. This exception contains structured information (with expr) instead of format objects.

This doesn't solve the problem with expanded pack functions in nested inductives, right?

leodemoura:
> This doesn't solve the problem with expanded pack functions in nested inductives, right?

No, it doesn't. We can fix this one when we fix #1518.
https://github.com/leanprover/lean/issues/1518#issuecomment-305295190
The fix is to use the user-facing `cases_on` at `elim_match`. We would completely avoid the `pack/unpack` madness there. We would only have to handle it when proving the equational lemmas, but this is fine. We already support them there. We would just have to unfold the user-facing `cases_on` recursors.

gebner:
Mutually recursive definitions and brec definitions now work as expected:
```lean
def foo (y : ℕ) : ℕ → ℕ → ℕ
| 0     a     := a
| (x+1) 0     := 3 + foo x 10
| (x+1) (n+2) := 42
/-
non-exhaustive match, the following cases are missing:
foo (nat.succ _) 1
-/

mutual def f, g
with f : nat → nat → nat
| 0 0 := 1
with g : nat → nat → nat
| 0 0 := 1
/-
non-exhaustive match, the following cases are missing:
f 0 (nat.succ _)
f (nat.succ _) _
g 0 (nat.succ _)
g (nat.succ _) _
-/
```

Actual match expressions are shown using the generated `f._match_1` function.  I think this should be ok, since the auxiliary match functions are user-facing anyhow.  We also show the extra arguments introduced by `pull_nested_rec_fn`, again I think this is fine because the user sees them anyhow.

Nested recursives are still shown using the very internal constructors and `pack`.

gebner:
> The fix is to use the user-facing cases_on at elim_match.

Did you expect any problems here?  I just flipped the switch, and everything seems to work.  (Well, at least all the tests pass.)  It was only a three-line change.

leodemoura:
> > The fix is to use the user-facing cases_on at elim_match.
>
> Did you expect any problems here? 

@gebner No, I didn't. Note that, this change is only part of the fix for #1518.

src/library/equations_compiler/elim_match.cpp
```diff
@@ -541,25 +541,36 @@ struct elim_match_fn {
...
+                        }
+                        return true;
                     } else {
                         return false;
                     }
                 }
             });
+        if (has_string)
+            return true;
         if (!r || !has_value || !has_variable)
```
leodemoura: This is incorrect if there are forward dependencies.

leodemoura: By "incorrect", I meant it may produce code that fails to type check.

tests/lean/non_exhaustive_error.lean.expected.out
```diff
@@ -16,3 +16,6 @@ f (nat.succ _) _
 g 0 (nat.succ _)
 g (nat.succ _) _
 non_exhaustive_error.lean:17:0: warning: declaration '_mutual.f.g' uses sorry
+non_exhaustive_error.lean:27:4: error: non-exhaustive match, the following cases are missing:
+const_name (term.mk _ (term.mk _ _ :: _))
+non_exhaustive_error.lean:27:0: warning: declaration 'const_name' uses sorry
```
leodemoura: We need parentheses around `_ :: _`.

gebner: It looks right to me.  This is the same expression with more parens:
```
const_name (term.mk _ ((term.mk _ _) :: _))
```

I'm using the normal pretty-printer here, so it should not be buggier than the usual error messages.

leodemoura: You are right, sorry.

src/library/equations_compiler/elim_match.cpp
```diff
@@ -526,26 +541,35 @@ struct elim_match_fn {
...
+                        return true;
                     } else {
                         return false;
                     }
                 }
             });
-        if (!r || !has_value || !has_variable)
+        if (!has_string && (!r || !has_value || !has_variable))
             return false;
         type_context ctx  = mk_type_context(P);
```
leodemoura, line 574: We have a similar problem with numerical literals.
Example: 
```lean
def f : nat → nat
| 1000 := 1
```
So, I replaced the condition above with 
```lean
if (!r || !has_value)
  return false;
```
This is sufficient for infinite types.
We also use `ite` for `char` literals. It is very unlikely someone will write a function that covers all char literals, but it is possible. So, just to be on the safe side. I also added
```
if (!has_variable && has_finite_value)
    return false;
```

leodemoura: Remark: So, Lean produces a nasty error message for 
```lean
def c : char -> bool
| 'a' := tt
```

leodemoura: Here is a potential issue with the change I described above.
```lean
def f : nat → nat
| 0 := 1
```
Produces an error message saying `f _` is missing which is not incorrect, but some users may think `f (succ _)` is more natural in this case. The same is true if the user uses only small nat values. We can add code for handling this case. What do you think?

PR #1722, opened 2017-07-05
The Great Renaming

Kha:
The one proposal I did not implement (yet) is `show ℕ := n` because visually that simply looks like a type mismatch to me :) .

### DIFF
--- a/library/data/bitvec.lean
+++ b/library/data/bitvec.lean
@@ -36,7 +36,7 @@ section shift
 
   def shl (x : bitvec n) (i : ℕ) : bitvec n :=
   bitvec.cong (by simp) $
-    dropn i x ++ₜ repeat ff (min n i)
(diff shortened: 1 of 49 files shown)


avigad:
:1st_place_medal: Wow, it looks great. I admire your fortitude.

johoelzl:
NIce! One question regarding `by`: 

Currently `by` is ignored. But shouldn't 
` by t ` translate to ` { t } `. So as an alternative for `solve.`

Kha:
@johoelzl Good catch, fixed

leodemoura:
@kha Great!

BTW, you didn't update `doc/changes.md` ;)

avigad:
BTW, I will make sure TPL has all the information for the next release. So the change log can be brief and refer to that.

PR #1726, opened 2017-07-05
add simp rule for 'true implies p'

avigad:
Without this rule, the following fail:
```lean
example (h : true → false) : false :=
by simp * at *

example (p q : Prop) (hp : p) (hnq : ¬ q) (h : p → q) : false :=
by simp * at *
```
With it, they succeed. 

When I ran the tests, 7 failed in `smt2`, and also `complete_field` and `thread_test`. As far as I can tell, these have nothing to do with this pr.


leodemoura:
The failure in `complete_field` is probably due to this PR. Changes in `logic.lean` may affect line numbers reported in this test.

Kha:
I will work on making the `interactive` tests independent from the stdlib tomorrow.

leodemoura:
> I will work on making the interactive tests independent from the stdlib tomorrow.

@Kha This is awesome. I will be so happy. Some of these tests I fixed more than 100 times :)

avigad:
Cool. In the meanwhile, I fixed the test.

PR #1731, opened 2017-07-06
mod whnf

Kha:
@leodemoura Looks good? It does make for some nice [definitions](https://github.com/Kha/semantics-lean/blob/37d5d49cba2050a150772a23d141c227755fcb7c/semantics.lean#L239) and [lemmas](https://github.com/Kha/semantics-lean/blob/37d5d49cba2050a150772a23d141c227755fcb7c/semantics.lean#L247) :) .

### DIFF
--- a/src/library/equations_compiler/util.cpp
+++ b/src/library/equations_compiler/util.cpp
@@ -122,11 +122,11 @@ unpack_eqns::unpack_eqns(type_context & ctx, expr const & e):
             }
         } else {
             /* noequation, guess arity using type of function */
-            expr type = ctx.infer(m_fns[fidx]);
+            expr type = ctx.relaxed_whnf(ctx.infer(m_fns[fidx]));
             unsigned arity = 0;
             while (is_pi(type)) {
                 arity++;
-                type = binding_body(type);
+                type = ctx.relaxed_whnf(binding_body(type));
             }
             if (arity == 0) throw_ill_formed_eqns();
             m_arity.push_back(arity);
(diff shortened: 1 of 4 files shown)


leodemoura:
I found an `ass` constructor in your file ;)
```lean
| ass  {x a σ} : ⟨x ::= a,σ⟩⇓σ[x ↦ A⟦a⟧σ]
```

src/library/equations_compiler/util.cpp
```diff
@@ -122,11 +122,11 @@ unpack_eqns::unpack_eqns(type_context & ctx, expr const & e):
...
         } else {
             /* noequation, guess arity using type of function */
-            expr type = ctx.infer(m_fns[fidx]);
+            expr type = ctx.relaxed_whnf(ctx.infer(m_fns[fidx]));
             unsigned arity = 0;
             while (is_pi(type)) {
                 arity++;
-                type = binding_body(type);
+                type = ctx.relaxed_whnf(binding_body(type));
             }
```
leodemoura, line 130: Type inference methods assume we don't have dangling de-Bruijn variables.
`whnf` depends on type inference (example K-rule in the kernel).

The following artificial example produces an error:
```lean
def f := ∀ (a : nat) (h : a = a), @eq.rec nat a (λ x, Prop) (empty → false) a h

lemma ex : f.
```

Kha:
I see, thanks for fixing it! I don't think I've ever used `push_local_from_binding` before.

> I found an ass constructor in your file ;)

... :P

PR #1734, opened 2017-07-07
fix(library/tactic/unfold_projs)

### LINKED ISSUE #1733: `unfold_projs at ?` failing unexpectedly
### Prerequisites

* [X] Put an X between the brackets on this line if you have done all of the following:
    * Checked that your issue isn't already [filed](https://github.com/leanprover/lean/issues).
        * Specifially, check out the [wishlist](https://github.com/leanprover/lean/issues?q=is%3Aissue+is%3Aopen+label%3AI-wishlist), open [RFCs](https://github.com/leanprover/lean/issues?q=is%3Aissue+is%3Aopen+label%3ARFC),
        or [feature requests](https://github.com/leanprover/lean/issues?q=is%3Aissue+is%3Aopen+label%3AFeature).
    * Reduced the issue to a self-contained, reproducible test case.

### Description

The [code](https://github.com/leanprover/lean/blob/1c389cb9252e8b5aa32e9eed802599915a7d9e85/library/init/meta/interactive.lean#L934)

````
private meta def unfold_projs_hyps (cfg : unfold_proj_config := {}) : list name → tactic unit
| []      := skip
| (h::hs) := get_local h >>= unfold_projs_hyp >> unfold_projs_hyps hs
````

has two problems:

1. it does not pass the argument `cfg` into `unfold_projs_hyp`.
2. it requires `unfold_projs_hyp` to succeed on every argument `h`, rather than at least one of them. 

See https://gist.github.com/semorrison/f7fcf72477e147079e53b1b4c48d5903 for an example of the failure the first issue causes.

### Versions

Lean (version 3.2.1, commit 99515f48f842, DEBUG)

kim-em:
Fixes #1733.

It might make sense to move try_list and at_least_one to more natural locations, or avoid their use.

### DIFF
--- a/library/init/meta/interactive.lean
+++ b/library/init/meta/interactive.lean
@@ -931,9 +931,18 @@ meta def delta : parse ident* → parse location → tactic unit
                           delta_target new_cs,
                           intron n
 
-private meta def unfold_projs_hyps (cfg : unfold_proj_config := {}) : list name → tactic unit
-| []      := skip
-| (h::hs) := get_local h >>= unfold_projs_hyp >> unfold_projs_hyps hs
+-- Applies a list of tactics in turn, always succeeding.
+meta def try_list : list (tactic unit) → tactic unit 
+| list.nil  := skip
+| (t :: ts) := try t >> try_list ts
+
+-- Applies a list of tactics in turn, succeeding if at least one succeeds.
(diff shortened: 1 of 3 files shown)


leodemoura:
ced4b6a54df

PR #1738, opened 2017-07-07
Generalize generalize

Kha:
Just the first step

### DIFF
--- a/library/data/bitvec.lean
+++ b/library/data/bitvec.lean
@@ -164,14 +164,14 @@ section conversion
     cases xs with xs P,
     simp [bits_to_nat_to_list], clear P,
     unfold bits_to_nat list.foldl,
-      -- the next 4 lines generalize the accumulator of foldl
-    let x := 0,
-    change _ = add_lsb x b + _,
-    generalize 0 y,
-    revert x, simp,
-    induction xs with x xs ; intro y,
-    { simp, unfold list.foldl add_lsb, simp [nat.mul_succ] },
-    { simp, apply ih_1 }
+    -- generalize the accumulator of foldl
+    have : ∀ x, list.foldl add_lsb x (xs ++ [b]) = add_lsb 0 b + list.foldl add_lsb x xs * 2, {
+      simp,
+      induction xs with x xs ; intro y,
+      { simp, unfold list.foldl add_lsb, simp [nat.mul_succ] },
+      { simp, apply ih_1 }
+    },
+    apply this
   end
 
   theorem bits_to_nat_to_bool (n : ℕ)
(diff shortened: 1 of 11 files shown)


library/init/meta/expr.lean
```diff
@@ -111,12 +111,16 @@ meta constant expr.instantiate_vars        : expr → list expr → expr
 
 protected meta constant expr.subst : expr elab → expr elab → expr elab
 
+/-- `has_var e` returns true iff e has free variables. -/
 meta constant expr.has_var       : expr → bool
```
Kha: @leodemoura I found this name a bit confusing tbh

library/init/meta/interactive.lean
```diff
@@ -461,16 +461,15 @@ do e ← i_to_expr p,
...
-   let e' := tgt.replace $ λa n, if a = e then some (var n.succ) else none,
-   tgt' ← (do
-     tgt' ← to_expr ``(Π x, %%e = x → %%e'),
-     type_check tgt',
-     return tgt') <|> to_expr ``(Π x, %%e = x → %%tgt),
-   assert h tgt',
+   e ← i_to_expr p,
+   -- if generalizing fails, fall back to not replacing anything
+   tgt' ← do {
+     ⟨tgt', _⟩ ← solve_aux tgt (tactic.generalize e `_ >> target),
```
Kha: @digama0 In the end, emulating `kabstract` by combining `solve_aux` and `generalize` turned out to be the simpler solution.

digama0: What does `solve_aux` do?

Kha: It runs a tactic on a new goal of some type and returns both the tactic result and the proof

Kha:
Okay, I believe I've officially run out of things to further generalize.

library/init/meta/interactive.lean
```diff
@@ -383,10 +383,44 @@ precedence `generalizing` : 0
...
+        intron n,
+        -- now try to clear hypotheses that may have been abstracted away
+        let locals := arg.fold list.nil (λ e _ acc, if e.is_local_constant then e::acc else acc),
+        locals.mfor' (try ∘ clear),
+        pure h
+      },
+      intron (n-1),
+      e ← intro1,
+      pure (e, newvars)
+   },
```
Kha, line 416: This implementation could probably be a bit brittle with just one test case, but at least should not break any `induction` invocation that worked before.

library/data/bitvec.lean
```diff
@@ -164,12 +164,10 @@ section conversion
...
     unfold bits_to_nat list.foldl,
-      -- the next 4 lines generalize the accumulator of foldl
-    let x := 0,
-    change _ = add_lsb x b + _,
-    generalize 0 y,
-    revert x, simp,
-    induction xs with x xs ; intro y,
+    -- generalize the accumulator of foldl
+    generalize h : 0 = x, conv in (add_lsb x b) { rw ←h }, clear h,
+    simp,
```
leodemoura, line 169: The new syntax looks awesome!

PR #1742, opened 2017-07-08
fix(src/library/vm,tests/lean): corrects vm implementation of int_shiftl

### LINKED ISSUE #1723: Correct value for int.shiftl (-1) (-1)?
Posting this here for visibility.  I did some random* testing on several functions that we override in the VM (just nat and int), and found one discrepancy:
```lean
#eval int.shiftl (-1) (-1)
-- 0
#reduce int.shiftl (-1) (-1)
-- -[1+ 0]
```

@digama0 You looked into bitwise operations recently, what *should* the correct result be here?

*I used a uniform distribution on [-40,40], and generated values using the pseudo random number generator known as `list.range`.

joshpoll:
Fixes #1723. Previously, the vm and lean implementations of `shiftl` disagreed when both inputs were negative due to type and rounding bugs in the vm version. This PR fixes that issue.


gebner:
Thanks!

PR #1747, opened 2017-07-16
feat(src/library/tactic): add list_axioms tactic

robertylewis:
This copies and slightly modifies the code used for `#print axioms` to add a tactic `list_axioms : name -> tactic (list name)`. 

Two questions. One, the code for this and for `#print axioms` can be combined pretty easily, but I wasn't sure where it should live. I'm happy to make that change if someone suggests a place. Two, `sorry` won't appear in the list of axioms, since it isn't a name. We could change the type to `name -> tactic (list name x bool)`, but that's slightly less convenient to use. Thoughts? Should we add both variants?

### DIFF
--- a/src/library/tactic/list_axioms_tactic.cpp
+++ b/src/library/tactic/list_axioms_tactic.cpp
@@ -1,0 +1,67 @@
+/*
+Copyright (c) 2017 Robert Y. Lewis. All rights reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+
+Author: Robert Y. Lewis
+*/
+#include "kernel/for_each_fn.h"
+#include "library/util.h"
+#include "library/vm/vm_name.h"
+#include "library/vm/vm_list.h"
+#include "library/tactic/tactic_state.h"
+#include "library/tactic/list_axioms_tactic.h"
+#include "library/sorry.h"
+
+namespace lean {
+
+void list_axioms_deps::visit(name const & n) {
+    if (m_visited.contains(n))
+        return;
+    m_visited.insert(n);
+    declaration const & d = m_env.get(n);
+    if (!d.is_definition() && !m_env.is_builtin(n)) {
+        m_use_axioms = true;
+        m_axioms.insert(d.get_name());
+    }
+    visit(d.get_type());
+    if (d.is_definition())
+        visit(d.get_value());
+}
+
+void list_axioms_deps::visit(expr const & e) {
+    for_each(e, [&](expr const & e, unsigned) {
+        if (is_sorry(e) && !m_used_sorry) {
+            m_used_sorry = true;
+        }
+        if (is_constant(e))
+            visit(const_name(e));
+        return true;
+    });
+}
+
+buffer<name> list_axioms_deps::axiom_list() {
+    buffer<name> b = buffer<name>();
+    m_axioms.to_buffer(b);
+    return b;
+}
+
+vm_obj tactic_list_axioms(vm_obj const & n, vm_obj const & s0) {
+    tactic_state const & s = tactic::to_state(s0);
+    environment const & env = s.env();
+    name nm = to_name(n);
+    try {
+        buffer<name> const & axiom_names = list_axioms_deps(env)(nm);
+        return tactic::mk_success(to_obj(axiom_names), s);
+    } catch (exception e) {
+        return tactic::mk_exception(e, s);
+    }
+}
+    
+void initialize_list_axioms_tactic() {
+    DECLARE_VM_BUILTIN(name({"tactic", "list_axioms"}), tactic_list_axioms);
+}
+
+void finalize_list_axioms_tactic() {
+}
+
+}
--- a/src/library/tactic/list_axioms_tactic.h
+++ b/src/library/tactic/list_axioms_tactic.h
@@ -1,0 +1,34 @@
+/*
+Copyright (c) 2017 Robert Y. Lewis. All rights reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+
+Author: Robert Y. Lewis
+*/
+
+#pragma once
+#include "util/name_set.h"
+
+namespace lean {
+
+struct list_axioms_deps {
+    environment     m_env;
+    name_set        m_visited;
+    name_set        m_axioms;
+    bool            m_use_axioms;
+    bool            m_used_sorry;
+    list_axioms_deps(environment const & env):
+        m_env(env), m_use_axioms(false), m_used_sorry(false) {}
+
+    void visit(name const & n);
+    void visit(expr const & e);
+    buffer<name> axiom_list();
+    buffer<name> operator()(name const & n) {
+        visit(n);
+        return axiom_list();
+   }
+};
+
+void initialize_list_axioms_tactic();
+void finalize_list_axioms_tactic();
+
+}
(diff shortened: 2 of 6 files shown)


digama0:
We could output a name `` `_sorry`` in the tactic and translate it to `[sorry]` in the printing. The initial underscore prevents any name clashes with user axioms.

robertylewis:
I don't want to do that. I should be able to get a list of names with `list_axioms` and map over the list to, e.g., get the type of each axiom. If there's a fake name in the list that will fail.

src/library/tactic/list_axioms_tactic.cpp
```diff
@@ -0,0 +1,67 @@
...
+namespace lean {
+
+void list_axioms_deps::visit(name const & n) {
+    if (m_visited.contains(n))
+        return;
+    m_visited.insert(n);
+    declaration const & d = m_env.get(n);
+    if (!d.is_definition() && !m_env.is_builtin(n)) {
+        m_use_axioms = true;
+        m_axioms.insert(d.get_name());
```
gebner, line 24: Maybe we should visit the type of an axiom as well, just for completeness.  EDIT: nevermind, can't read.

src/library/tactic/list_axioms_tactic.cpp
```diff
@@ -0,0 +1,67 @@
...
+        if (is_sorry(e) && !m_used_sorry) {
+            m_used_sorry = true;
+        }
+        if (is_constant(e))
+            visit(const_name(e));
+        return true;
+    });
+}
+
+buffer<name> list_axioms_deps::axiom_list() {
```
gebner, line 42: You can't safely copy or return buffers (since they're stack-allocated).  Just return the name_set here.

src/library/tactic/list_axioms_tactic.h
```diff
@@ -0,0 +1,34 @@
...
+    name_set        m_axioms;
+    bool            m_use_axioms;
+    bool            m_used_sorry;
+    list_axioms_deps(environment const & env):
+        m_env(env), m_use_axioms(false), m_used_sorry(false) {}
+
+    void visit(name const & n);
+    void visit(expr const & e);
+    buffer<name> axiom_list();
+    buffer<name> operator()(name const & n) {
```
gebner, line 25: Maybe `#print axioms` should use this function as well.

gebner:
> I'm happy to make that change if someone suggests a place.

I think the current place is fine.

> We could change the type to name -> tactic (list name x bool), but that's slightly less convenient to use.

I think we need to report sorry as part of the axioms.  For me, `_sorry` would be fine as well, but that requires special cases as you've noticed.

digama0:
Regardless of how you do it, `sorry` needs to be special-cased somehow, because you can't look `sorry` up in the context. A more principled approach would be `list Axiom` where `inductive Axiom | sorry | defn (n : name)`, but that seems like a lot of bookkeeping. Of course that type is isomorphic to `option name`, so you could use that, or even just use `name.anonymous` for the `sorry` case.

gebner:
I like the inductive, at least it's descriptive.

leodemoura:
This is the kind of tactic that should be implemented in Lean.
We should only expose the missing APIs. Example: `environment.is_builtin : environment -> name -> bool`.

leodemoura:
> Thoughts? Should we add both variants?

I think we should not add any of them in this repository.
We should only make sure that users can write any of these variants in Lean.

src/library/tactic/list_axioms_tactic.h
robertylewis: Yes, this was only a slight modification of the `#print axioms` code at https://github.com/leanprover/lean/blob/d610709a5358de741f5094ccdfed6241434252e2/src/frontends/lean/print_cmd.cpp#L39 . They should share one function, I just wasn't sure where that function should live. It doesn't seem right for `print_cmd.cpp` to import something in the `tactic` folder, no?

gebner: Most of the commands call tactics, so I think it would be fine.

robertylewis:
> I think we should not add any of them in this repository.
We should only make sure that users can write any of these variants in Lean.

Being able to write it in Lean would be great. This does seem like a useful tactic to have available by default though, why don't you think it belongs in the main repo?

digama0:
> I think we should not add any of them in this repository.
We should only make sure that users can write any of these variants in Lean.

@leodemoura I strongly disagree. Even if it is trivial to implement, axiom tracking is a core Lean functionality, and we should not trust users to do it correctly. An incorrect result from `#print axioms` is just as bad as an actual inconsistency, since it is lying to a reviewer about the logical strength of a proof. I would prefer for this to be implemented in C, not for efficiency reasons but rather to make it more complicated to rewrite. The situation is no different than with statement checking in `init_quotient` as far as I am concerned.

leodemoura:
> Being able to write it in Lean would be great. This does seem like a useful tactic to have available by default though, why don't you think it belongs in the main repo?

Because it can be written in Lean, and there is no tactic in this repository that needs it.

leodemoura:
> @leodemoura I strongly disagree. Even if it is trivial to implement, axiom tracking is a core Lean functionality, and we should not trust users to do it correctly. An incorrect result from #print axioms is just as bad as an actual inconsistency, since it is lying to a reviewer about the logical strength of a proof. I would prefer for this to be implemented in C, not for efficiency reasons but rather to make it more complicated to rewrite. The situation is no different than with statement checking in init_quotient as far as I am concerned.

@digama0 
- We already have `#print axioms`. 
- We have reference type checkers, and users that really care about trust should rely on them (Reason: they are much easier to inspect). This functionality can be easily added to them.
- There isn't a single tactic that needs the `list_axioms` tactic.

digama0:
> We already have #print axioms.

I guess I misunderstood - I thought we were planning on reimplementing `#print axioms` using this tactic / C function.

robertylewis:
I guess my point is that `#print axioms` and `list axioms` are the same function. Given that we have and need one of them, why shouldn't we provide the other one as well? It could be implemented in C++ and exposed, or it could be implemented in Lean and `print axioms` could be a user command (or something like that). But implementing `#print axioms` and leaving users on their own to reimplement it for `list_axioms` seems redundant.

leodemoura:
> I guess my point is that #print axioms and list axioms are the same function. Given that we have and need one of them, why shouldn't we provide the other one as well? It could be implemented in C++ and exposed, or it could be implemented in Lean and print axioms could be a user command (or something like that). But implementing #print axioms and leaving users on their own to reimplement it for list_axioms seems redundant.

I understand you are trying to help and submit PRs that you think are useful for the project.
I don't want to be rude, but I will not accept this PR. 
From my point of view, you are proposing a refactoring to provide a functionality that is not needed (the `list_axioms` tactic). If I merge this PR, I will have to maintain it. Yes, I agree `#print axioms` and `list_axioms` are the same in spirit, but I wrote `#print axioms`, and it is trivial for me to maintain it. 
You are not a skilled C++ programmer, and your code requires careful reviewing. 
@gebner pointed out a basic error above, and I found similar bugs in your code before.
I have a lot of things to do, and I don't want to review your code, nor waste time with a functionality that is not needed from my point of view, nor I want to discuss/read a discussion on how to expose this functionality. If someone needs it, they can implement, is this wasteful? Yes, it is, but I don't care since I don't have to maintain the other copy.
Sorry, if I'm being harsh. Please put yourself in my position. I don't have the luxury of ignoring discussions and PRs related to Lean since I have to maintain it. I'm happy to do it when I think the PR is useful and the code is high quality. For example, PRs from @gebner and @Kha are merged very quickly.

PR #1751, opened 2017-07-18
Enhance & refactor some elaboration error messages

Kha:
I fear that the quality of our error reporting (both content and quantity of messages) has somewhat declined, so I'm trying to improve any bad errors I notice. In the process of that, I also made sure to factor out common subparts of error messages.


tests/lean/emptyc_errors.lean.expected.out
```diff
@@ -8,9 +8,3 @@ context:
 A : Type u,
 x : A
 ⊢ Type ?
-emptyc_errors.lean:5:0: error: type mismatch, expression
-  λ (h : x ∈ ∅), h
-has type
-  x ∈ ∅ → x ∈ ∅
-but is expected to have type
-  x ∉ ∅
```
Kha, line 16: note: gone because of sorrys in implicit args

tests/lean/string_imp.lean.expected.out
```diff
@@ -3,7 +3,8 @@ string_imp.lean:1:4: error: equation compiler failed (use 'set_option trace.eqn_
 string_imp.lean:5:3: error: invalid pattern: variable, constructor or constant tagged as pattern expected
 string_imp.lean:5:3: error: function expected at
   string_imp.mk
+term has type
+  _
 string_imp.lean:4:4: error: equation compiler failed (use 'set_option trace.eqn_compiler.elim_match true' for additional details)
 string_imp.lean:8:0: error: unknown identifier 'string_imp.cases_on'
-string_imp.lean:8:0: error: function expected at
-  ⁇
+string_imp.lean:7:0: warning: declaration 'h' uses sorry
```
Kha, line 10: @gebner `error_already_reported` seems to fail here. Could it be that this is the first instance of a definition that has a parser error but no elab error?

tests/lean/refine_error.lean.expected.out
```diff
@@ -1,5 +1,5 @@
-refine_error.lean:5:1: error: type mismatch, expression
-  @pax ℤ int.has_zero
+refine_error.lean:5:1: error: type mismatch, term
+  pax
```
Kha, line 2: Note the small regression -- I don't think it makes sense to use the `until_different` formatter for the types on the term as well in general. After #1674, we may want to just show the original input (if available) instead.

tests/lean/string_imp.lean.expected.out
gebner: > Could it be that this is the first instance of a definition that has a parser error but no elab error?

I believe this is the issue here.

leodemoura:
> I fear that the quality of our error reporting (both content and quantity of messages) has somewhat declined, so I'm trying to improve any bad errors I notice. In the process of that, I also made sure to factor out common subparts of error messages.

Do you know why it declined?

Kha:
@leodemoura 
> Do you know why it declined?

'Declined' was probably too strong, I just noticed users getting confused by these errors recently. The error recovery certainly pushed up the quantity of errors, but it was still worth it and we can slowly work on rectifying that.

PR #1752, opened 2017-07-18
refactor(init/meta/tactic): use cases on decidable argument

digama0:
instead of on decidable.em, so that you can eliminate into Type.

Reverts 9a41f0f

### DIFF
--- a/library/init/data/array/lemmas.lean
+++ b/library/init/data/array/lemmas.lean
@@ -129,10 +129,9 @@ def read_foreach_aux (f : fin n → α → α) (ai : array α n) :
 | (i+1) hi a ⟨j, hj⟩ ji := begin
   dsimp [iterate_aux], dsimp at ji,
   change ite _ _ _ = _,
-  by_cases (⟨i, hi⟩ : fin _) = ⟨j, hj⟩ with e,
-  { simp [e, if_pos] },
-  { simp [if_neg e],
-    rw [read_foreach_aux _ _ _ ⟨j, hj⟩],
+  by_cases (⟨i, hi⟩ : fin _) = ⟨j, hj⟩ with e; dsimp [ite],
+  { simp [e] },
+  { rw [read_foreach_aux _ _ _ ⟨j, hj⟩],
     exact (lt_or_eq_of_le (nat.le_of_lt_succ ji)).resolve_right
       (ne.symm $ mt (@fin.eq_of_veq _ ⟨i, hi⟩ ⟨j, hj⟩) e) }
 end
--- a/library/init/data/nat/bitwise.lean
+++ b/library/init/data/nat/bitwise.lean
@@ -5,7 +5,7 @@ Author: Mario Carneiro
 -/
 
 prelude
-import .lemmas init.meta.well_founded_tactics
+import init.data.nat.lemmas init.meta.well_founded_tactics
 
 universe u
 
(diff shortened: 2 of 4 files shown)


gebner:
Yes!  I'll merge it when the tests succeed.

gebner:
@digama0 The `simp_dif` test is failing.

Kha:
Which is the test added 15 days ago by @leodemoura when he made the reverse change :) 9a41f0f8996aeea90d833db4c34a89523cbfb3d8

digama0:
Ah, that's exactly the behavior I want, it's not a bug. If the decidable instance is replaced by a constructor, it definitionally reduces (as I've done here with the `; dsimp [ite]` in the tactic proofs that were broken by the change). It is a lot easier than what I used to write, `cases (show decidable p, by apply_instance)` (or similar using `match`).

digama0:
I don't know if it is possible to have `simp` automatically reduce `ite` applications like this, but `dsimp [ite]` is to me the correct way to simplify these goals (and it works well when the if statement is explicit). `simp [h]` is the alternative when the if is not explicit or appears later (in which case `if_pos` and `if_neg` should be triggered as simp rules). Both of these apply uniformly to both positive and negative branches after the `by_cases`.

leodemoura:
> Ah, that's exactly the behavior I want, it's not a bug. If the decidable instance is replaced by a constructor, it definitionally reduces (as I've done here with the ; dsimp [ite] in the tactic proofs that were broken by the change). It is a lot easier than what I used to write, cases (show decidable p, by apply_instance) (or similar using match).

The behavior you want was considered counterintuitive by other users. 
Commit 9a41f0f was based on a problem described by an user.
I took the time to figure out what was happening on his example and document what was happening. I will not merge this PR.

digama0:
@leodemoura What were the circumstances of this counterintuitive case? Even if we get rid of the substitution of the decidable constructors (which should really be iota reduced away anyway), the limitation to Prop is really artificial and can be alternately resolved. Please reopen this so we can consider solutions that resolve all the problems and satisfies whatever intuitions people have in this case.

leodemoura:
> @leodemoura What were the circumstances of this counterintuitive case? 

@digama0 I think the commit message at 9a41f0f is pretty clear.

digama0:
I see two solutions for this:

1. Don't rewrite `h` in the major premise. This would satisfy everyone, I think, since `if_pos h` would work in this case. The trick is to avoid the rewrite, since `cases` always rewrites the parameter being cased on.
2. Iota reduce `ite` and `dite` automatically after `by_cases`. This would immediately discharge any if-expressions, so `if_pos h` wouldn't even be necessary.

(2) is the natural extension of this PR to avoid any counterintuitive if statements in goals. (1) requires some kind of mechanism to avoid the rewrite. One hackish way to do it is to case on `id inst` instead of `inst`, since it probably won't find `id inst` in the goal. Another way is to apply `@dite p inst`, which is essentially nondependent `decidable.rec`. In this case there is no dependent motive, so no variables will be replaced.

digama0:
I found another solution to the issue in 9a41f0f which does not require this isolated semantics changing of `cases` in this situation, implemented at https://github.com/leanprover/lean/compare/master...digama0:decidable_cases2 . If the argument to `if_pos` and `if_neg` is implicit instead of a typeclass argument, then it will unify with anything in that slot, so the original proof in the test case works even though the instance is replaced after the cases.

PR #1753, opened 2017-07-18
refactor(init/meta/tactic): use dite instead of cases in by_cases

digama0:
This fixes the issue in PR #1752 by using `dite`, which is essentially nondependent cases, so that the decidable instances are not substituted.

### DIFF
--- a/library/init/meta/tactic.lean
+++ b/library/init/meta/tactic.lean
@@ -1004,8 +1004,10 @@ do tgt : expr ← target,
 meta def by_cases (e : expr) (h : name) : tactic unit :=
 do dec_e ← (mk_app `decidable [e] <|> fail "by_cases tactic failed, type is not a proposition"),
    inst  ← (mk_instance dec_e <|> fail "by_cases tactic failed, type of given expression is not decidable"),
-   em    ← mk_app `decidable.em [e, inst],
-   cases em [h, h]
+   t     ← target,
+   tm    ← (mk_mapp `dite [some e, some inst, some t] <|>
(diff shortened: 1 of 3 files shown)


library/init/meta/tactic.lean
```diff
@@ -1004,8 +1004,10 @@ do tgt : expr ← target,
 meta def by_cases (e : expr) (h : name) : tactic unit :=
 do dec_e ← (mk_app `decidable [e] <|> fail "by_cases tactic failed, type is not a proposition"),
    inst  ← (mk_instance dec_e <|> fail "by_cases tactic failed, type of given expression is not decidable"),
-   em    ← mk_app `decidable.em [e, inst],
-   cases em [h, h]
+   t     ← target,
+   tm    ← (mk_mapp `dite [some e, some inst, some t] <|>
+     fail "by_cases tactic failed, type of given expression is not decidable"),
```
gebner: I don't think we can ever hit this case.  If the proposition is not decidable, already the call to mk_instance will have failed.

gebner (approved):

gebner:
Once again, I really like the change that we now allow `by_cases` also in non-Prop goals.

leodemoura:
> Once again, I really like the change that we now allow by_cases also in non-Prop goals.

@gebner Ok, I will revise this PR carefully on Thursday. Tomorrow, I have to give a talk, and I didn't finish my slides yet.
I got really pissed with the previous PR when `simp_dif` was broken, and then @digama0 fixed the issue by rewriting the test. I think I was super clear on the PR that introduce the test.

PR #1756, opened 2017-07-20
refactor(*): move out stdlib

digama0:
This PR removes most of `library/data` from the repo. Don't merge this until the material has found a home in the soon-to-be created `stdlib` repo.


tests/lean/run/1659.lean
```diff
@@ -1,19 +0,0 @@
-import data.hash_map
```
digama0, line 1: This test was removed because it depends on some intimate details of the `hash_map` structure, and it wasn't clear that making a stub here would preserve the bugged behavior under test. If we really want to keep this bug test, we should probably try to make a better minimum working example using the old build or something.

tests/lean/run/hash_map1.lean
```diff
@@ -1,29 +0,0 @@
-import data.hash_map
```
digama0, line 1: This test was removed because obviously the core doesn't have hash_maps anymore. Perhaps we could create a test directory in stdlib as well, but since `hash_map` is now fully verified the test cases seem moot.

leodemoura:
@digama0 I will merge, but want to chat with you about it and how the others repos will be organized.
Do you have time this afternoon?

Kha:
There may also be some things in `init/` we can move out (`inductive_predicate`).

joehendrix:
Is there going to be some mechanism like submodules for checking out development builds of lean and the  standard library that should be guaranteed to work together?

leodemoura:
I created the stdlib repo: https://github.com/leanprover/stdlib
@digama0 You have write access.

leodemoura:
@digama0 Joe (@joehendrix) raised an important issue.
I think we have to address the following related issues:
- Distribution: we build binary distribution packages for each Lean release. We also have a `brew` formula.
The should include the standard library. 
- The `master` branch at stdlib will probably be incompatible with the Lean official release. We have a similar issue with packages. We discussed this issue in the past, and one idea was to specify the Lean version a package depends on.

gebner:
> Is there going to be some mechanism like submodules for checking out development builds of lean and the standard library that should be guaranteed to work together?

The current and official solution is to use leanpkg to add `stdlib` as a dependency.  For Lean files outside of packages, we can use `leanpkg install` to install the standard library.

> Distribution: we build binary distribution packages for each Lean release. We also have a brew formula.
The should include the standard library.

I think we should rather encourage people to become comfortable `leanpkg` and make it work smoothly with release versions.  Even if we bundle the standard library, people will still want to use `smt2_interface`, etc.

N.B. this is also the approach taken by Agda.

> We discussed this issue in the past, and one idea was to specify the Lean version a package depends on.

Concrete proposals:

1) We can add a `lean_version` field to `leanpkg.toml`:
```toml
[package]
name = "leanscratch"
version = "0.1"
lean_version = "3.2"
```

The `leanpkg` tool could give a warning if the used Lean version does not match the declared one.

The semantics of the `lean_version` would be as follows: right now, we use the `master` branch by default in the `install`, `add`, and `upgrade` commands to determine the latest git revision of a dependency.  With the `lean_version` field set, we'd use the `lean-v${version}` branch/tag instead.

2) In release builds (not the cmake build, but builds for the release tarballs and distribution packages), we modify leanpkg to use the `lean-v${version}` branch instead of `master` for the update operations.  This could be implemented by a LEANPKG_DEFAULT_BRANCH variable in cmake, that we change at the release version bump.  cmake would then generate a `leanpkg/version.lean` file from these variables like we already do for the C++ code.

leodemoura:
>  I think we should rather encourage people to become comfortable leanpkg and make it work smoothly with release versions. Even if we bundle the standard library, people will still want to use smt2_interface, etc.
N.B. this is also the approach taken by Agda.

I'm concerned about this one. Most programming languages are installed with their standard libraries.
Another issue is that `stdlib` will probably be quite big (@digama0 is really fast). It would be nice to have a single copy (per version) in our machines. If we use the approach above, `leanpkg` will create a copy of the stdlib in each project directory, right?

leodemoura:
Orthogonal issue: this week we were talking about having a native bit-blaster and a SAT solver Lean. 
I think this is a great idea, and it will open many new cool application scenarios.
Thus, we should keep the bit-vector library in this repo.

gebner:
> If we use the approach above, leanpkg will create a copy of the stdlib in each project directory, right?

Yes, that is true.  My concern about bundling is that this would be a big difference between release builds and git versions.

> this week we were talking about having a native bit-blaster and a SAT solver Lean.

I don't think this would block moving bitvectors out of the core library.  My current plan is to have a fairly modular bitblaster, that you can customize via custom lemmas.  (That is, you would have lemmas that tell you how to commute `vector.nth` with `bitvec.add` for example.)  Also even if we implement the CNF generation in C++ instead of in Lean due to performance issues, there should be no dependency on bitvectors either.

digama0:
We don't have to merge this yet, but it is no longer blocked on the `stdlib` import, which is now complete and compiles relative to this branch.

joehendrix:
I think for non-development builds, it would be good to eventually have version fields for projects and associated version requirement (similar to Haskell cabal).

Given Lean's rapid evolution however, it would be good if there was an easy way to check out development builds of Lean and the standard library, and know they were tested to work together.  Maybe if the standard-library repo included lean as a submodule or some field in the standard library's toml file).

For my current project involving Lean, I keep a submodule with Lean in the repo, so all the people on the project know which version of Lean this was intended to be used with.   We then update the submodule whenever it seems like there are new bug fixes or features to try out.  It seems like this patch just means that we'll need to add a reference to the standard library commit (via a commit id in the leanpkg.toml file).

leodemoura:
@digama0 I will merge this PR today.
However, I will move bitvector (bv) back to this repo based on our meeting on Monday.
Here is the motivation
1- Efficient code generation for bv operations
2- BV decision procedures and simplifiers that will be implemented in C++.

leodemoura:
> Yes, that is true. My concern about bundling is that this would be a big difference between release builds and git versions.

@gebner I agree. I'm still concerned about this, but I don't see a better solution.

leodemoura:
Merged.

Kha:
@leodemoura @digama0 

> I will merge, but want to chat with you about it and how the others repos will be organized.

For posterity, what was the answer to this? Especially the relation between `stdlib` and `library_dev`?

leodemoura:
> For posterity, what was the answer to this? Especially the relation between stdlib and library_dev?

Here is the summary of the meeting.
- @digama0 will manage `stdlib`, and will make sure it compiles with the latest Lean.
- Initially, `stdlib` will contain all Lean libraries we care about. This will make @digama0's life easier. Otherwise, he will have to keep patching multiple repos. If the stdlib gets too big, we will create smaller repos (e.g., one for analysis).
- If @digama0 is not available to fix problems with `stdlib`, we should just comment the broken parts.
- `library_dev` will continue to exist. We will use it for experiments.

Kha:
Sounds good, thanks for the summary

PR #1764, opened 2017-07-25
perf(util/rc): do not use fences

gebner:
(I'm not merging this directly, because I have no idea about the actual performance impact.)

As suggested by David Chisnall.  If I read the spec correctly, it would be unsafe to use the release ordering for dec_ref_core: then the following situation could happen:

```c++
// m_rc -> 2

// Thread 1:
unsigned x = atomic_fetch_sub_explicit(&m_rc, 1u, memory_order_release);

// Thread 2:
unsigned y = atomic_fetch_sub_explicit(&m_rc, 1u, memory_order_release);

// x = y = 1u, m_rc -> 2
```

That is, a release store operation is not required to be visible to another release operation.

Herb Sutter also recommends the acq_rel ordering for reference counters.

### DIFF
--- a/src/util/rc.h
+++ b/src/util/rc.h
@@ -14,12 +14,11 @@ Author: Leonardo de Moura
 private:                                                                \
 atomic<unsigned> m_rc;                                                  \
 public:                                                                 \
-unsigned get_rc() const { return atomic_load(&m_rc); }                  \
+unsigned get_rc() const { return atomic_load_explicit(&m_rc, memory_order_acquire); } \
 void inc_ref() { atomic_fetch_add_explicit(&m_rc, 1u, memory_order_relaxed); } \
 bool dec_ref_core() {                                                   \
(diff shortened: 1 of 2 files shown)


leodemoura:
@gebner The current implementation is also based on suggestions from C++ gurus.
Unfortunately, I did not add links to them. 
Herb is a C++ grand wizard. So, I will merge this PR as is. 
Could you please add a link to Herb's post where he suggested this approach?
I want to have it just in case I have to bug him in the future.

gebner:
> Could you please add a link to Herb's post where he suggested this approach?

It's from his "atomic weapons" talk: https://channel9.msdn.com/Shows/Going+Deep/Cpp-and-Beyond-2012-Herb-Sutter-atomic-Weapons-2-of-2

The reference counting example is on page 51 of the slides: https://1drv.ms/b/s!Aq0V7yDPsIZOgcI0y2P8R-VifbnTtw

PR #1776, opened 2017-07-29
Add documentation for the syntax changes in Lean 3

mk12:
These are the changes I have discovered so far while porting my own project to Lean 3. It would have been much easier if they were all documented in one place, so that's what I propose to do now.

Let me know if anything is inaccurate, or if it would be better to put this in changes.md rather than in a separate file.

cc @avigad

### DIFF
--- a/doc/changes.md
+++ b/doc/changes.md
@@ -1,3 +1,5 @@
+See [What's new in Lean 3](https://github.com/leanprover/lean/blob/master/doc/new_in_lean3.md) for the changes from Lean 2 to Lean 3.
+
 master branch (aka work in progress branch)
 -------------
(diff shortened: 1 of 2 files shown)


leodemoura:
Thanks for sending the PR, but I will not merge for the following reasons:
- It is incomplete.
- Lean 3 is a moving target (see `changes.md`).
- It will bit rot. 
- Lean 2 has very few users. The HoTT ones stayed with Lean 2, and the most of the other users have already moved to Lean 3.
- We have a `changes.md` file, so this kind of issue will not happen in the future. Note that, we did not try to document the changes from 3.0 to 3.2.

PR #1777, opened 2017-07-30
Backlog

digama0:
More small commits from stdlib development.

### DIFF
--- a/library/init/logic.lean
+++ b/library/init/logic.lean
@@ -41,8 +41,8 @@ assume hna : ¬a, absurd ha hna
 
 /- false -/
 
-lemma false.elim {c : Prop} (h : false) : c :=
-false.rec c h
+@[inline] def false.elim {C : Sort u} (h : false) : C :=
+false.rec C h
 
 /- eq -/
 
--- a/library/init/meta/interactive_base.lean
+++ b/library/init/meta/interactive_base.lean
@@ -59,6 +59,9 @@ meta def brackets (l r : string) (p : parser α) := tk l *> p <* tk r
 
 meta def list_of (p : parser α) := brackets "[" "]" $ sep_by (skip_info (tk ",")) p
 
+precedence `⊢` : 0
+precedence `|-` : 0
+
 /-- A 'tactic expression', which uses right-binding power 2 so that it is terminated by
     '<|>' (rbp 2), ';' (rbp 1), and ',' (rbp 0). It should be used for any (potentially)
     trailing expression parameters. -/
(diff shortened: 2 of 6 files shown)


library/init/meta/interactive_base.lean
```diff
@@ -59,6 +59,9 @@ meta def brackets (l r : string) (p : parser α) := tk l *> p <* tk r
 
 meta def list_of (p : parser α) := brackets "[" "]" $ sep_by (skip_info (tk ",")) p
 
+precedence `⊢` : 0
```
gebner, line 62: Why do we need this declaration?

digama0: Otherwise `rw at h1 ⊢` doesn't parse correctly. The scanner rejects it before it has a chance to be parsed.

library/init/meta/tactic.lean
```diff
@@ -101,6 +101,11 @@ meta def try_lst : list (tactic unit) → tactic unit
     end
   end
 
+meta def continue (t : tactic α) : tactic (option α) :=
+λ s, result.cases_on (t s)
+ (λ a, success (some a))
+ (λ e ref s', success none s')
+
```
leodemoura, line 108: We have a tactic called `try_core`
https://github.com/leanprover/lean/blob/master/library/init/meta/tactic.lean#L81-L84

digama0: @leodemoura Note the `s'` in the last line. This tactic is like `try_core` except that in the case of an error it allows you to continue from the failure state rather than backtracking (hence the name).

leodemoura:
@digama0 This PR breaks two tests.
One of them is easy to fix. The other one raises an issue with the new `subst`.
For example, we will not be able to write 
```lean
example (a b c : nat) : a = 0 → b = 1 → c = 0 → b + a + c = b :=
begin
  intros,
  repeat {subst ‹_ = 0›}, -- substitute all equalities of the form `_ = 0`
  guard_target b + 0 + 0 = b,
  reflexivity
end
```

library/init/logic.lean
```diff
@@ -41,8 +41,8 @@ assume hna : ¬a, absurd ha hna
 
 /- false -/
 
-lemma false.elim {c : Prop} (h : false) : c :=
-false.rec c h
+@[inline] def false.elim {C : Sort u} (h : false) : C :=
```
gebner, line 44: You don't need `inline` here, the body of `false.elim` is a "simple" application and should be inlined automatically.

gebner:
@digama0 I'm quite willing to review and merge these kind of PRs if they're 1) only about a single, motivated change and 2) all tests pass and there are no merge conflicts.  (From my point of view, half of this PR is pretty uncontroversial.  I'd merge that part, but I'm too lazy to separate the commits.)

PR #1779, opened 2017-07-30
Specialize Tactic

jroesch:
Add an implementation of Coq's specialize tactic, this covers much of the functionality I used in Coq. Suggestions and improvements welcome.


Kha:
You can use `tests/lean/run` for output-less test instead

library/init/meta/specialize_tactic.lean
```diff
@@ -0,0 +1,30 @@
...
+meta def specialize_core (loc : expr) (args : list expr) : tactic unit :=
+do let original_name := loc.local_pp_name,
+   new_name ← mk_fresh_name,
+   rename original_name new_name,
+   F ← get_local new_name,
+   note original_name none (expr.mk_app F args),
+   clear_lst [new_name],
+   pure ()
+
+meta def specialize (e : parse texpr) : tactic unit := do
```
Kha, line 21: Could you add a doc comment and a changelog entry? Also, I think this tactic is short enough that it can go into `init.meta.interactive`.

library/init/meta/specialize_tactic.lean
```diff
@@ -0,0 +1,30 @@
...
+
+open interactive interactive.types expr
+
+meta def specialize_core (loc : expr) (args : list expr) : tactic unit :=
+do let original_name := loc.local_pp_name,
+   new_name ← mk_fresh_name,
+   rename original_name new_name,
+   F ← get_local new_name,
+   note original_name none (expr.mk_app F args),
+   clear_lst [new_name],
```
digama0, line 18: If you use `clear F` directly, you don't have to rename anything. Is there a reason this isn't implemented simply as `have f := f x y z`?

library/init/meta/specialize_tactic.lean
jroesch: Sounds good, was just waiting to see if people liked the implementation first :)

PR #1784, opened 2017-08-02
[RFC] remove order pairs

gebner:
Replace `linear_strong_order_pair` and others by the more standard notions `pre_order`, `partial_order` and `linear_order`.  From what I gather the whole distinction between `weak_order` and `strict_order` was intended to be useful in constructive logic, but we've pretty much settled on classical logic in Prop now.

I prefer this change for three reasons:
 1. The current names and structures are fairly non-standard.  I never knew which classes represent linear orders, partial orders, or something else.  It was also not clear which type classes should be provided by (linear) orders, and which should be required for definitions that use them.
 2. There is a duplicate hierarchy between weak and strict orders, with no clear guidelines on when to use which.  (Both the weak and strict hierarchies are in use, with no clear way to convert between them.)
 3. Defining new linear orders required a lot of definitions and lemmas.  Ideally I'd just give a relation plus proofs for reflexivity, transitivity, antisymmetry, and totality.

With this change, defining a linear order is straightforward:
```lean
example : linear_order unit := {
    le := λ _ _, true,
    le_refl := λ _, trivial,
    le_trans := λ _ _ _ _ _, trivial,
    le_antisymm := by intros a b; intros; cases a; cases b; refl,
    le_total := λ _ _, or.inl trivial,
}
```
The less-then relation now comes for free.  It has a default value that is suitable for preorders in general.

There were some questions about how this change interacts with a potential implementation of constructive reals: as far as I can see, this change should pose no problem for constructive reals as long as you're willing to use classical logic to prove properties about them, or don't need order theorems that mix < and ≤.  You can still define your own transitivity lemmas for the constructive < and ≤, though.  Other options would be 1) to add a separate hierarchy (weak_pre_order, weak_partial_order, weak_linear_order) that is *only* used by the constructive reals, or 2) try adding enough negations so that you can constructively prove the relation between < and ≤.

Here is the definition of the type classes if you don't want to look through the diff:
```lean
class pre_order (α : Type u) extends has_le α, has_lt α :=
(le_refl : ∀ a : α, a ≤ a)
(le_trans : ∀ a b c : α, a ≤ b → b ≤ c → a ≤ c)
(lt := λ a b, a ≤ b ∧ ¬ b ≤ a)
(lt_iff_le_not_le : ∀ a b : α, a < b ↔ (a ≤ b ∧ ¬ b ≤ a) . order_laws_tac)

class partial_order (α : Type u) extends pre_order α :=
(le_antisymm : ∀ a b : α, a ≤ b → b ≤ a → a = b)

class linear_order (α : Type u) extends partial_order α :=
(le_total : ∀ a b : α, a ≤ b ∨ b ≤ a)
```
The `le` and `lt` fields are different so that we can have e.g. `n < m = n+1 ≤ m` as a definitional equality.


avigad:
I support this change.

PR #1795, opened 2017-08-17
feat(init/meta/interactive): multiple subst

digama0:
This allows `subst a, subst b, subst c` to be written `subst a b c`. In order to support `‹_›` notation without backtracking, a special parser is introduced for mixed identifier / assumption brackets, which can be listed without delimiters, making it possible to use them in other places where hypothesis names are expected (such as `at h1 h2`).

I attempted to use the `parser.pexpr` parser directly, and this almost works, but it breaks code such as `begin subst a end` because `end` is parsed as an expr and fails *after* consuming the token, so the parser cannot continue. The `ident` and `tk` parsers do not consume their input on failure, which is why the `identa` parser still works in this situation. This can be worked around by using a comma as in `begin subst a, end` but this PR instead special cases the `‹_›` notation to maintain backward compatibility.

This is a resubmission of one part of #1777.

### DIFF
--- a/library/init/meta/interactive.lean
+++ b/library/init/meta/interactive.lean
@@ -919,8 +919,8 @@ tactic.ac_refl
 meta def cc : tactic unit :=
 tactic.cc
 
-meta def subst (q : parse texpr) : tactic unit :=
-i_to_expr q >>= tactic.subst >> try (tactic.reflexivity reducible)
+meta def subst (l : parse parser.pexpr*) : tactic unit :=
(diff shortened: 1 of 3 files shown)


leodemoura:
I don't like this hack.
Moreover, this feature is not very useful: just check how often it would be used in the core and math libs.
In the next following months, this kind of PR is not welcome, and I don't want to discuss alternative syntax and hacks. This is not important to make the project successful.

PR #1796, opened 2017-08-17
feat(init/meta/injection_tactic): injection also does subst

digama0:
Now it should be about as smart as the internal injection tactic used for discharging index equalities in `cases`.

This is a resubmission of one part of #1777.


digama0:
Note, in case this PR is being delayed because of test failure: The tests are passing on this, but `mathlib` will need to be updated after the merge, and the Travis failure is due to this. AFAIK there is no way to resolve this "race condition" of testing so that the PR is submitted with a green checkmark already.

leodemoura:
I don't think this a good change, but I don't have time to discuss it.
https://github.com/leanprover/lean/blob/master/.github/CONTRIBUTING.md#opening-pull-requests

leodemoura:
will not be merged

PR #1797, opened 2017-08-17
feat(init/meta/interactive): cases with equality

digama0:
This is the equivalent of the `ginduction` tactic for cases, but rolled into the same syntax as `cases` itself. `cases h : term` is the syntax, and it will introduce a hypothesis `h : term = C a b...` demonstrating that the original term is equal to the current case.

I considered the possibility of calling `injection` on the generated equalities, but it's useless in the casaes when the equality carries some real information (such as `f x = C1 a`), and when the input term is a local constant, `injection` will do subst, which will undo the effect of the `cases`. If the input term is a constructor, then `injection` would do something interesting, but you would never want to call `cases` in this case because the constructor is already exposed.

### DIFF
--- a/library/init/meta/interactive.lean
+++ b/library/init/meta/interactive.lean
@@ -422,10 +422,6 @@ do e ← i_to_expr p,
      clear_lst (newvars.map local_pp_name),
      (e::locals).mmap' (try ∘ clear) }
 
-meta def cases (p : parse texpr) (ids : parse with_ident_list) : tactic unit :=
-do e ← i_to_expr p,
-   tactic.cases e ids
-
 private meta def find_case (goals : list expr) (ty : name) (idx : nat) (num_indices : nat) : option expr → expr → option (expr × expr)
 | case e := if e.has_meta_var then match e with
   | (mvar _ _ _)    :=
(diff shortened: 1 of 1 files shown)


library/init/meta/interactive.lean
```diff
@@ -521,6 +517,25 @@ do x ← mk_fresh_name,
    t ← get_local x,
    induction (to_pexpr t) rec_name hs ([] : list name)
 
+private meta def cases_arg_p : parser (option name × pexpr) :=
+with_desc "(id :)? expr" $ do
+  t ← texpr,
+  match t with
+  | (local_const x _ _ _) :=
```
Kha, line 524: Heh, that's quite clever. Should we do this for `have`/`generalize`/... too?

digama0: We can't do it for `have`, because the type is not mandatory (you can do just `have h` and it is okay). You wouldn't be able to distinguish it from `have t` where `t` is a local constant. We could use this trick for `generalize`, though, with a bit of modification to also parse the equality right.

Kha: The term-mode `have` will default to interpreting it as a type, so it should be reasonable to do the same in tactic mode.

digama0: Hm, I'm not sure I like that, because `have h` can always be possibly a type (before taking the actual tactic state into account), so we would have to assume this every time. This means that `let h := _` is no longer valid - it would have to be written `let h : _ := _` which I'm not so happy with. (It might be possible to eliminate the type entirely, as in `let h : := _`, but that's a bit weird to read.)

Kha: I'm not saying this is the best syntax I can think of, but I do believe that we should be as consistent as possible. Note that in term-mode `let` the name is not optional, which feels relatively intuitive to me.

In short, in term mode we have the syntaxes
```
"assume" binders
"assume" ":" expr -- anoymous
"have" ident ":" expr
"have" expr -- anonymous
"let" ident (":" expr)
```
With your trick, we emulate all of them in tactic mode.

leodemoura:
merged

PR #1800, opened 2017-08-22
Allow field notation for recursive calls

Kha:
Not a high-priority feature, but I kept writing code that failed because of this inconsistency.


src/frontends/lean/elaborator.cpp
```diff
@@ -2742,8 +2749,11 @@ elaborator::field_resolution elaborator::field_to_decl(expr const & e, expr cons
         if (is_structure_like(m_env, const_name(I)))
             if (auto p = find_field(m_env, const_name(I), fname))
                 return {const_name(I), *p, fname};
-        // TODO(sullrich): respect inheritance for extended field notation too?
         name full_fname = const_name(I) + fname;
+        if (auto ldecl = m_ctx.lctx().find_local_decl_from_user_name(full_fname.replace_prefix(get_namespace(env()), {}))) {
+            // recursive call
+            return {full_fname, ldecl};
+        }
```
leodemoura, line 2756: Please do not use anonymous constructors from now on.

Issue #722, opened 2015-07-05
Lean's convertibility is not transitive


maximedenes:
This is probably known since it is very related to issue #654. Following a remark by Matthieu Sozeau, I was investigating how Lean could deal with singleton elimination and proof irrelevance and noticed that convertibility is not transitive at the moment.

Of course, feel free to close the issue if it is considered as a feature of the formalism.

Here is a concrete example, using acc:

open nat

definition wf_nat (n : nat) : acc lt n :=
match lt.wf with
| well_founded.intro p := p n
end

/- f always returns 0 -/
definition f : ∀n : nat, acc lt n -> nat
| f 0 p := 0
| f (succ m) (acc.intro _ p) :=
    f m (p _ (le.refl (succ m)))

/- a is convertible to b -/
check (eq.refl _ : (λp, f 1 p) = (λp, f 1 (wf_nat 1)))

/- b is convertible to c -/
check (eq.refl _ : (λp, f 1 (wf_nat 1)) = (λp, 0))

/- but a is not convertible to c -/
check (eq.refl _ : (λp, f 1 p) = (λp, 0))

leodemoura:
Dear Maxime,

Thanks for posting the issue. Yes, this issue is very related to #654. 
The trick discussed at #654 could be used to address this issue. In the third `check`, it would replace `p` with `acc.intro 1 (fun y h, acc.inv p h)` and "unblock" the computation, and both sides would reduce to `λp, 0`. Unfortunately, this trick makes the convertibility checker non-terminating. We thought about limiting the number of times it can be applied, but we would be "sacrificing" transitivity again.

We also considered defining `acc` in `Type`, and removing singleton elimination from the kernel.
This is also not nice, if we do that, I don't see how to define 

``` lean
find_x : (∃ a : nat, p a) → {a : nat | p a}
```

https://github.com/leanprover/lean/blob/master/library/data/nat/choose.lean#L96

Comments about this issue and #654 would be very welcome. This is one of the things that keep me awake at night :-)
Are you considering supporting OTT in Coq?

Best wishes,
Leo

leodemoura:
@avigad @fpvandoorn any thoughts?
I'm planning to add a link from #654 and close it.

avigad:
[...]

PR #1803, opened 2017-08-24
Remove the requirement to reduce proofs

gebner:
Even though we already have definitional proof irrelevance, subsingleton reduction still requires the major premise (a proof!) to reduce to a constructor application.  This leads to a variety of practical and metatheoretical problems:
 * We need to δ-reduce theorems.
 * Definitional equality is not transitive.
 * Subject reduction fails.

The main application of this feature is the definitional reduction of well-founded recursion.  However, practically we make no essential use of it.  Hence I propose to completely remove subsingleton reduction except where it is justified by K-like reduction (and then the major premise is irrelevant).  Concretely, we change the elimination levels of inductive types to:

> The recursor eliminates only into Prop if and only if:
>  1. The level of the inductive type may be zero, and
>  2. There is more than one introduction rule, or an introduction rule has a non-parameter argument.

In the core library, this change only breaks the `acc` type, which now no longer has subsingleton elimination.  As suggested by @leodemoura and @digama0, we define an (a priori) noncomputable auxiliary recursor `acc.crec` ("computable recursor") using choice and function extensionality that can eliminate into Type.  We make `acc.crec` computable by fiat and provide a suitable VM implementation.

This change fixes all the problems listed in the introduction, that is:
 * We never need to reduce proofs.
 * We conjecture that definitional equality is transitive.
 * We conjecture that subject reduction holds.

Collateral damage:
 * `and.dcases_on` no longer eliminates into Type.  We could fix this by implementing `cases_on` using projections when possible.


gebner:
To illustrate, the following types still have large elimination:
 * `true` (K)
 * `false`
 * `eq` (K)
 * `heq` (K)

These no longer have it:
 * `acc`
 * `and`
 * `subsingleton`, etc.

leodemoura:
I will not merge this PR in the near future. I have to think more about it.
I don't see any pressing need for this PR. The problems it solves are not really affecting users. That is, it will not significantly improve user experience. Moreover, the problems described in the PR are well understood by the type theory community. That is, a hard core type theory person will not be surprised by any of them. They are annoying, but they are expected.

The change proposed by this PR will produce a system that is non standard. At least, I have never seen a system using this setup. I believe this will create confusion. For example, declarations defined using well founded recursion will not compute in the kernel. The use of classical axioms in the `wf.lean` module will also make people raise eyebrows. 

I am aware that it is inefficient to compute with `acc` in the Kernel in Lean and Coq. However, not being able to compute with them at all in the kernel will surprise users.

gebner:
I won't push this change.  I agree, this won't really change anything for users.  Should we at least revert the `acc.rec` special case in the type context, so that users don't accidentally unfold proofs?

> For example, declarations defined using well founded recursion will not compute in the kernel.

For what it's worth, in general they don't compute now, either.  We boldly use classical logic (or even just propositional extensionality) whenever it's useful, and I believe this is an important and distinctive feature of Lean.

Obviously, the use of either propext or choice can cause a term to be stuck now.  This is possible even if it does not use well-founded recursion.

> The change proposed by this PR will produce a system that is non standard. At least, I have never seen a system using this setup.

Well, the combination of in-kernel proof irrelevance and subsingleton elimination for `acc` is just as non-standard.  But yes, the kind of setup in this PR is normally only used by the HOL provers AFAIK.  I don't think any system based on type theory does that (although Coq with axioms + custom program extraction comes close).

I actually wanted to extend this approach to other "computable" functions as well.  Several people have requested support for more general quotient elimination, for example to convert `nat → quotient α` to `quotient (nat → α)`.  Just like the `acc.crec` function, this is easily definable using choice, and can be safely made computable on the VM.

```lean
def quot.lift''
    {γ : Sort w} {α : γ → Sort u} {r : ∀ i, α i → α i → Prop}
    {β : Sort v} (f : (∀ i, α i) → β)
    (h : ∀ a b : (∀ i, α i), (∀ i, r i (a i) (b i)) → f a = f b)
    (q : ∀ i, quot (r i)) : β
```

(EDIT: removed unrelated code snippet.  Sorry, I copy-pasted this without looking.)

leodemoura:
> Should we at least revert the acc.rec special case in the type context, so that users don't accidentally unfold proofs?

I'm happy to keep this commit. We can change if it becomes a problem.

> For what it's worth, in general they don't compute now, either. We boldly use classical logic (or even just propositional extensionality) whenever it's useful, and I believe this is an important and distinctive feature of Lean.

Only if they use our definitions. An user can create their own definitions and automation that do not use propositional extensionality and other axioms.

> Well, the combination of in-kernel proof irrelevance and subsingleton elimination for acc is just as non-standard.

Proof irrelevance, K axiom, and subsingleton elimination has been studied in the literature. The problems created by the combination (and listed in this PR) are also well known. 

> I actually wanted to extend this approach to other "computable" functions as well. Several people have requested support for more general quotient elimination, for example to convert nat → quotient α to quotient (nat → α). Just like the acc.crec function, this is easily definable using choice, and can be safely made computable on the VM.

If "several people" == @digama0, then it doesn't count :) 
I want to see a very strong proposal and motivation before we make any changes to the kernel.
Any change in the kernel must be applied to all reference type checkers, and we have to carefully check whether it is sound or not. Several of these requests are just "for fun", people want to play with the new feature, but don't have any concrete application for it. Sometimes, they have an application, but the application is just a minor convenience. That is, they could do it without the new feature.

avigad:
Is there a clear statement of subsingleton elimination somewhere? Right now, TPL and the reference manual describe singleton elimination as Gabriel portrayed it abov. I would like to correct them to describe subsingleton elimination (even if we change it later).

@gebner what is "large elimination" for `true`? Isn't it just `(true → α)  → α` for `α : Sort u`? What do we need K for?

avigad:
Correction: in TPL and the reference manual we say: "you are allowed to eliminate from an inductively defined Prop to an arbitrary Sort when there is only one constructor, and each argument to that constructor is either in Prop or an index." It seems to me that this is correct, except that we should say "each argument to that constructor that is not a parameter is either in Prop or an index." I will also describe the extra `acc` reduction hack.

gebner:
> subsingleton elimination vs. singleton elimination vs. large elimination

I think we've been using these terms virtually interchangeable.  I'm sorry if this caused any confusion.

> what is "large elimination" for true?

`true` has large/subsingleton elimination because it satisfies the condition that you've already posted.  That is, the eliminator `true.rec` has one more universe level parameter than the type former, allowing the motive to range over sorts other than Prop.  Concretely, it has the following type:
```lean
true.rec : ∀ {C : Sort l}, C → true → C
```

If `true` didn't have subsingleton elimination, then the type of the eliminator would be `∀ {C : Prop}, C → true → C`.

K-like reduction for `true` simply means that the following term always reduces definitionally, even if the major premise `x` is not `true.intro`:
```lean
true.rec m x   ==>   m
```

I don't think we use K for `true` anywhere though...

gebner:
Closing as we won't merge this.  @leodemoura Just for the record, do you know if the issues surrounding definitional proof irrelevance and subsingleton elimination are published somewhere?

leodemoura:
@gebner I don't remember the reference. I recall we had an issue about it submitted by one of the Coq guys. It was easy to find using the github search: https://github.com/leanprover/lean/issues/722
I will hopefully meet with them at ITP, and I can ask for the reference.

PR #1808, opened 2017-09-01
[RFC] feat(leanpkg): store Lean version in leanpkg.toml

gebner:
This RFC adds a field to `leanpkg.toml` indicating the intended Lean version of a package:
```toml
[package]
name = "hott3"
version = "0.1"
lean_version = "master"
```

The main intended use of this feature is to make it easier to use release versions of Lean:
 * You can use `leanpkg add` and `leanpkg upgrade` on release versions of Lean, and you should get versions that work with that release.
 * You get a warning if you don't have the right Lean version for a package.

Possible values for the `lean_version` field: `master`, `3.3.0`, etc.

Depending on the Lean version, `leanpkg` will then use the git branches `master` or `lean-3.3.0`, resp., when upgrading or adding dependencies.

I think we discussed this feature in Cambridge--did I miss anything?


avigad:
If I understand correctly, here is how it works. If someone develops a package `foo`, in addition to the `master` branch, they maintain branches for any release versions they want to support. Then when someone else uses `leanpkg` to build `foo` or any package that has `foo` in the chain of dependencies, the right version of `foo` is loaded. If `foo` doesn't have a branch for the release they are using, you fall back on `master` and they get a warning.

Have I got that right? If so, I love it.

gebner:
Pretty much, except for the fallback to master.  I'm not sure if we should do that--I don't think packages developed for Lean master will continue working on release versions, at least not in the near future.

Note that this change only affects `leanpkg add` and `leanpkg upgrade`, in all other cases we already have a good git hash.

avigad:
Got it. I agree, it is probably better to just fail if the package doesn't support the relevant release.

PR #1809, opened 2017-09-01
Parameterized user-defined attributes

Kha:
Add a parameter parser to `user_attribute`, store the parsed result in reflected form and provide a typed API to retrieve it. Implement a `[derive]` attribute on top of that.

/cc @rlewis1988, https://github.com/formalabstracts/formalabstracts/pull/49

### DIFF
--- a/library/init/meta/derive.lean
+++ b/library/init/meta/derive.lean
@@ -1,0 +1,64 @@
+/-
+Copyright (c) 2017 Microsoft Corporation. All rights reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: Sebastian Ullrich
+
+Attribute that can automatically derive typeclass instances.
+-/
+prelude
+import init.meta.attribute
+import init.meta.interactive_base
+import init.meta.mk_dec_eq_instance
+import init.meta.mk_has_sizeof_instance
+import init.meta.mk_inhabited_instance
+
+open lean
+open interactive.types
+open tactic
+
+/-- A handler that may or may not be able to implement the typeclass `cls` for `decl`.
+    It should return `tt` if it was able to derive `cls` and `ff` if it does not know
+    how to derive `cls`, in which case lower-priority handlers will be tried next. -/
+meta def derive_handler := Π (cls : pexpr) (decl : name), tactic bool
+
+@[user_attribute]
+meta def derive_handler_attr : user_attribute :=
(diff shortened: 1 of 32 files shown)


library/init/meta/derive.lean
```diff
@@ -0,0 +1,61 @@
...
+| (h::hs) :=
+do success ← h p n,
+   when (¬success) $
+     try_handlers hs
+
+@[user_attribute] meta def derive_attr : user_attribute unit (list pexpr) :=
+{ name := `derive, descr := "automatically derive typeclass instances",
+  parser := pexpr_list_or_texpr,
+  after_set := some (λ n _ _,
+  do ps ← derive_attr.get_param n,
```
Kha, line 39: Note the recursive use of `derive_attr` -- unbounded recursion can be helpful indeed.

gebner:
Nice!  Is there a way to set a parameterized user attribute from Lean?

gebner:
Hmm, the following would fail, right?
```lean
inductive foo | bar | baz
attribute [derive decidable_eq] foo
attribute [derive inhabited] foo
attribute [derive has_reflect] foo
```

Kha:
@gebner 
> Is there a way to set a parameterized user attribute from Lean?

Oh, I guess there should be! I was actually wondering whether we should change the system attributes to persist a single `expr` as well so that we can finally set them from Lean.

Kha:
@gebner That will actually work (apart from the missing derive handler for `has_reflect`), but only the last `[derive]` instance will be persisted (which I guess doesn't matter for that attribute).

Kha:
That is,

```lean
inductive foo | bar | baz
attribute [derive decidable_eq] foo
attribute [derive inhabited] foo

#check foo.decidable_eq
│ foo.decidable_eq : decidable_eq foo

#check foo.inhabited
│ foo.inhabited : inhabited foo

#print foo
│ @[derive list.cons.{0} pexpr ``(inhabited) (list.nil.{0} pexpr)]
│ inductive foo : Type                                            
│ constructors:                                                   
│ foo.bar : foo                                                   
│ foo.baz : foo                                                   
```

(We should probably use an `io_state_stream` for attribute printing...)

leodemoura:
@kha please do not merge PRs which are not small bug fixes.
I wanted to revise this PR before merging it.

PR #1818, opened 2017-09-08
Flesh out [derive]

Kha:
Handle deriving classes for inductive types with universe levels and parameters. Add a derive handler for `has_reflect`.

### DIFF
--- a/library/init/meta/mk_has_reflect_instance.lean
+++ b/library/init/meta/mk_has_reflect_instance.lean
@@ -27,10 +27,10 @@ do t    ← infer_type a,
    do {
      m    ← mk_app `reflected [a],
      inst ← mk_instance m
-   <|>
-   do {
-     f ← pp t,
-     fail (to_fmt "mk_has_reflect_instance failed, failed to generate instance for" ++ format.nest 2 (format.line ++ f)) },
+     <|> do {
+       f ← pp t,
+       fail (to_fmt "mk_has_reflect_instance failed, failed to generate instance for" ++ format.nest 2 (format.line ++ f))
+     },
      mk_app `reflect [a, inst] }
 
 /- Synthesize (recursive) instances of `reflected` for all fields -/
@@ -70,6 +70,8 @@ do I_name ← get_has_reflect_type_name,
    env ← get_env,
    v_name : name ← return `_v,
    F_name : name ← return `_F,
+   guard (env.inductive_num_indices I_name = 0) <|>
+     fail "mk_has_reflect_instance failed, indexed families are currently not supported",
    -- Use brec_on if type is recursive.
    -- We store the functional in the variable F.
    if is_recursive env I_name
--- a/src/frontends/lean/decl_cmds.cpp
+++ b/src/frontends/lean/decl_cmds.cpp
@@ -139,6 +139,7 @@ static environment declare_var(parser & p, environment env,
         if (meta.m_modifiers.m_is_protected)
             env = add_protected(env, full_n);
         env = ensure_decl_namespaces(env, full_n);
+        /* Apply attributes last so that they may access any information on the new decl */
         env = meta.m_attrs.apply(env, p.ios(), full_n);
         return env;
     }
--- a/src/frontends/lean/structure_cmd.cpp
+++ b/src/frontends/lean/structure_cmd.cpp
@@ -1304,6 +1304,7 @@ struct structure_cmd_fn {
             declare_no_confusion();
             declare_injective_lemmas();
         }
+        /* Apply attributes last so that they may access any information on the new decl */
         m_env = m_meta_info.m_attrs.apply(m_env, m_p.ios(), m_name);
         return m_env;
     }
--- a/doc/changes.md
+++ b/doc/changes.md
@@ -137,7 +137,7 @@ For more details, see discussion [here](https://github.com/leanprover/lean/pull/
   `⟨h1, h2⟩`. This change is motivated by the previous one. Without it, `and.intro h1 h2` would be
   pretty printed as `{left := h1, right := h2}`.
 
-* User attributes can no longer be set with `set_basic_attribute`.  You need to use `user_attribute.set_param` now.
+* User attributes can no longer be set with `set_basic_attribute`.  You need to use `user_attribute.set` now.
 
 *API name changes*
 
--- a/leanpkg/leanpkg/resolve.lean
+++ b/leanpkg/leanpkg/resolve.lean
@@ -87,11 +87,11 @@ deps ← monad.filter (not_yet_assigned ∘ dependency.name) d.dependencies,
 deps.mmap' (materialize relpath),
 deps.mmap' $ λ dep, do
   p ← resolved_path dep.name,
-  d ← manifest.from_file $ p ++ "/" ++ "leanpkg.toml",
-  when (d.name ≠ dep.name) $
-    io.fail $ d.name ++ " (in " ++ relpath ++ ") depends on " ++ d.name ++
+  d' ← manifest.from_file $ p ++ "/" ++ "leanpkg.toml",
+  when (d'.name ≠ dep.name) $
+    io.fail $ d.name ++ " (in " ++ relpath ++ ") depends on " ++ d'.name ++
       ", but resolved dependency has name " ++ dep.name ++ " (in " ++ p ++ ")",
-  solve_deps_core p d max_depth
+  solve_deps_core p d' max_depth
 
 def solve_deps (d : manifest) : io assignment := do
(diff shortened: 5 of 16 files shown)


library/init/meta/mk_has_reflect_instance.lean
```diff
@@ -0,0 +1,73 @@
+/-
+Copyright (c) 2017 Microsoft Corporation. All rights reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: Sebastian Ullrich
+
+Helper tactic for constructing has_reflect instance.
+-/
+prelude
```
leodemoura: Could you please add comments describing what is generated?
That is, the general form of the generated instances.

src/frontends/lean/decl_cmds.cpp
```diff
@@ -139,6 +139,7 @@ static environment declare_var(parser & p, environment env,
         if (meta.m_modifiers.m_is_protected)
             env = add_protected(env, full_n);
         env = ensure_decl_namespaces(env, full_n);
+        env = meta.m_attrs.apply(env, p.ios(), full_n);
         return env;
```
leodemoura: This is a bug fix, right? 
We were ignoring attribute declarations in this case.

src/frontends/lean/structure_cmd.cpp
```diff
@@ -1289,6 +1287,7 @@ struct structure_cmd_fn {
             declare_no_confusion();
             declare_injective_lemmas();
         }
+        m_env = m_meta_info.m_attrs.apply(m_env, m_p.ios(), m_name);
         return m_env;
```
leodemoura: Is this also a bug fix?

src/frontends/lean/decl_cmds.cpp
Kha: Yes. The test "broken" by the fix used "@[simp] axiom" and did one more `simp` than should be necessary.

src/frontends/lean/structure_cmd.cpp
Kha: Yes, or at least it is bringing `structure`'s behavior more in line with the other commands. Not sure where to document that behavior.

In my specific case, `@[derive has_reflect] structure` was trying to invoke `induction` before the structure was registered as an inductive.

leodemoura: Before `m_env = m_meta_info.m_attrs.apply(m_env, m_p.ios(), m_name);`, we can add
```cpp
/*
Remark: it is important to apply attributes after the structure has been registered as an inductive 
datatype. For example, suppose we have `@[derive has_reflect] structure`, if the attributes were
applied earlier, the application would fail.
*/
```

library/init/meta/mk_has_reflect_instance.lean
```diff
@@ -56,6 +56,9 @@ private meta def for_each_has_reflect_goal : name → name → list (list name)
   solve1 (has_reflect_case I_name F_name ns),
   for_each_has_reflect_goal I_name F_name nss
 
+/-- Solves a goal of the form `has_reflect α` where α is an inductive type.
+    Needs to synthesize a `reflected` instance for each inductive parameter type of α
+    and for each constructor parameter of α. -/
```
Kha: Oh, I see now I misread your comment. I will add comments to the implementation later.

src/frontends/lean/structure_cmd.cpp
Kha: The need for replicating such a comment across all commands (and for the two fixes) probably points to an underlying design issue. I think we can do better and enforce the convention by moving attribute application into `parse_command`. Commands return a list of declaration names they want attributes applied to (multiple ones for mutual definitions, none for non-declaration commands). We can probably factor out even more code that way like applying docstrings, registering namespaces, etc.

Not only should this streamline the C++ code, it would also help anyone implementing a user command. What do you think?

leodemoura: I don't like the idea of returning a "package of actions" to be executed in the parser.
This kind of "package" can get arbitrarily complicated: registering attributes, doc strings, etc.
It will get more complicated as we add/change features. Moreover, it is not clear whether the order we execute actions will always be the same for all commands. 
I think we can cleanup the code by providing a default "finalization procedure". This procedure can be built using smaller ones. If the default is not good for a specific command, then we assemble a new one using (some of) the smaller ones.

I'm also trying to minimize the amount of work on the parser since we have issue #1674. 
To be honest, I don't think it is feasible to attack issue #1674 this year. It is a major change.
First, redoing `parser`+`pp`+`elaborator` would be a major amount of work. Whenever, I fix bugs on the elaborator, it reminds me on how much stuff would have to be re-implemented. 
I think a more realistic path is to first write the new `parser` and a converter from `syntax` object to `pexpr`. In this way, we don't need to re-implement `pp` and the `elaborator`. If we succeed, then we try to implement the pretty printer. Then, we try the `elaborator`. 
Just a new `parser`, `syntax` object and `pp` would give us most of the benefits. At least, the ones that are visible to most users. These features are sufficient for implementing refactoring tools,  hole commands, and tactics that suggest replacements (e.g., sledgehammer-like tactics that run for awhile but produce a new tactic script that can be replayed efficiently). A new elaborator would be mainly useful for us and very sophisticated users.

Kha: > I don't like the idea of returning a "package of actions" to be executed in the parser.
This kind of "package" can get arbitrarily complicated: registering attributes, doc strings, etc.

Note that we already have such a package in form of the `cmd_meta` structure. Intuitively, it would make sense to me if this information that is parsed independent of the command would also be exactly the one we could apply independent of the command.

> I think we can cleanup the code by providing a default "finalization procedure".

Okay, that would also work. A small advantage of the cmd_meta-returning approach is that we could check centrally that a non-declaration command indeed doesn't take attributes or a doc string, which would have found the `@[simp] axiom` bug. Alternatively, we could pass the cmd_meta by non-const reference and check in the end whether all information has been "consumed".

> I think a more realistic path is to first write the new parser and a converter from syntax object to pexpr. In this way, we don't need to re-implement pp and the elaborator. If we succeed, then we try to implement the pretty printer. Then, we try the elaborator.

Yes, I believe that is the only realistic approach. I hope I can soon estimate how much time I will be able to put into that project.

library/init/meta/mk_has_reflect_instance.lean
Kha: Updated.

leodemoura:
I'm having problems with indexed families.
The following works:
```lean
universes u
inductive foo (α β : Type u) : nat → Type u
| bar : ℕ → foo 0
| baz (n : nat) : foo n → α → foo (nat.succ n)

instance (α β : Type u) (n : nat) [decidable_eq α] [decidable_eq β] : decidable_eq (foo α β n) :=
by tactic.mk_dec_eq_instance
```

but the following one doesn't

```lean
universes u

set_option trace.app_builder true

@[derive decidable_eq]
inductive foo (α β : Type u) : nat → Type u
| bar : ℕ → foo 0
| baz (n : nat) : foo n → α → foo (nat.succ n)
```

leodemoura:
`@[derive inhabited]` adds hypotheses assuming **all** parameters are inhabited. This produces sub-optimal instances.
Example:
```lean
universes u

@[derive inhabited]
inductive foo (α β : Type u) 
| mk  : foo
| mk1 : α → foo 
| mk2 : β → foo 

#print foo.inhabited -- assumes that α and β are inhabited

instance better_instance (α β : Type u) : inhabited (foo α β) :=
⟨foo.mk α β⟩
```
I think this will generate confusion. See the following example
```lean
@[derive inhabited]
inductive List (α : Type u)
| nil : List
| cons : α → List → List

#print List.inhabited -- assumes that α is inhabited
```

Kha:
@leodemoura I see, so `inhabited` does not quite follow the same pattern as the other two typeclasses. I think I'll remove its derive_handler for now since it's not a very important typeclass (I think).

Kha:
@leodemoura While addressing your first bug, I noticed that `mk_has_sizeof_instance` itself breaks on indexed families (and `mk_has_reflect_instance` as well since it's based on the same template):

```lean
inductive foo : nat → Type
| baz (n : nat) : foo n → foo (nat.succ n)

lemma foo.size (a : ℕ) : has_sizeof (foo a) :=
by (tactic.mk_has_sizeof_instance) -- get_local tactic failed, unknown '_p.1' local
```

Kha:
I have now removed the `inhabited` handler and explicitly disallowed indexed families for `has_reflect`. I pushed two unrelated but trivial fixes as well.

PR #1819, opened 2017-09-09
fix(frontends/lean/elaborator): go back to ignoring implicit args in quote patterns

Kha:
/cc @rlewis1988


src/frontends/lean/elaborator.cpp
```diff
@@ -3697,6 +3710,7 @@ void elaborator::report_error(tactic_state const & s, char const * state_header,
 
 void elaborator::ensure_no_unassigned_metavars(expr & e) {
     // TODO(gabriel): this needs to change e
+    if (!has_expr_metavar(e)) return;
```
Kha, line 3713: I had accidentally removed this line in a previous commit, sorry about that.

PR #1829, opened 2017-09-21
merge ginduction -> induction

digama0:
and also `by_cases h : p` in place of `by_cases p with h`. These changes are meant to improve consistency between cases, by_cases, and induction, and reduce the number of tactic names to remember.

### DIFF
--- a/library/init/algebra/functions.lean
+++ b/library/init/algebra/functions.lean
@@ -24,7 +24,9 @@ solve1 $ intros
 >> try `[apply le_of_not_le, assumption]
 
 meta def tactic.interactive.min_tac (a b : interactive.parse lean.parser.pexpr) : tactic unit :=
-`[by_cases (%%a ≤ %%b), repeat {min_tac_step}]
+interactive.by_cases (none, ``(%%a ≤ %%b)); min_tac_step
+-- TODO(Mario): why doesn't this work?
+--`[by_cases (%%a ≤ %%b), repeat {min_tac_step}]
 
 lemma min_le_left (a b : α) : min a b ≤ a :=
 by min_tac a b
--- a/library/init/meta/interactive.lean
+++ b/library/init/meta/interactive.lean
@@ -384,13 +384,17 @@ with_desc "(id :)? expr" $ do
   | _ := pure (none, t)
   end
 
-private meta def generalize_arg_p : pexpr → parser (pexpr × name)
-| (app (app (macro _ [const `eq _ ]) h) (local_const x _ _ _)) := pure (h, x)
-| _ := fail "parse error"
+private meta def generalize_arg_p : parser (pexpr × name) :=
+with_desc "expr = id" $ do
+  e ← parser.pexpr 0,
+  match e with
+  | (app (app (macro _ [const `eq _ ]) h) (local_const x _ _ _)) := pure (h, x)
+  | _ := fail "parse error"
+  end
 
 /-- `generalize : e = x` replaces all occurrences of `e` in the target with a new hypothesis `x` of the same type.
     `generalize h : e = x` in addition registers the hypothesis `h : e = x`. -/
-meta def generalize (h : parse ident?) (p : parse $ tk ":" *> with_desc "expr = id" (parser.pexpr 0 >>= generalize_arg_p)) : tactic unit :=
+meta def generalize (h : parse ident?) (_ : parse $ tk ":") (p : parse generalize_arg_p) : tactic unit :=
 do let (p, x) := p,
    e ← i_to_expr p,
    some h ← pure h | tactic.generalize e x >> intro1 >> skip,
@@ -412,7 +416,7 @@ meta def induction (hp : parse cases_arg_p) (rec_name : parse using_ident) (ids 
 do e ← match hp with
    | (some h, p) := do
      x ← mk_fresh_name,
-     generalize h (p, x),
+     generalize h () (p, x),
      get_local x
    | (none, p) := i_to_expr p
    end,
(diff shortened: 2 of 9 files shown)


library/init/algebra/functions.lean
```diff
@@ -24,7 +24,9 @@ solve1 $ intros
 >> try `[apply le_of_not_le, assumption]
 
 meta def tactic.interactive.min_tac (a b : interactive.parse lean.parser.pexpr) : tactic unit :=
-`[by_cases (%%a ≤ %%b), repeat {min_tac_step}]
+interactive.by_cases (none, ``(%%a ≤ %%b)); min_tac_step
+-- TODO(Mario): why doesn't this work?
+--`[by_cases (%%a ≤ %%b), repeat {min_tac_step}]
```
digama0, line 29: @gebner Do you know why the `expr.subst` application is not well typed here? From the output term, it seems like the modified `by_cases` parser is still able to correctly deduce the structure even though there are holes in the preterm, but the output term says it should be a `bool` instead of a `pexpr` which doesn't make any sense.

Kha: Ah, you've uncovered a dirty hack of mine: The instance `has_reflect pexpr` is a lie for preterms with antiquotations because, being outside of `tactic`, it cannot elaborate the antiquotations nor synthesize `to_pexpr` instances for them. Instead, for interactive parameters of type `pexpr`, we hackishly treat the reflected expression as a preterm after all: https://github.com/leanprover/lean/blob/d83b9ef3efbf2d9eafce32c26d047e1f69539a2f/src/library/vm/vm_parser.cpp#L69-L73 Because you're returning a `pexpr` in a tuple, this doesn't trigger.

Really, what we should do is changing `parse` from taking a `has_reflect` instance to some new typeclass `has_prereflect` that produces preterms. It would not be a big change (we can give a transitive instance from `has_reflect`), but I'm not sure it would make sense to tackle this before #1674.

library/init/meta/interactive.lean
```diff
@@ -975,13 +971,13 @@ private meta def delta_hyps : list name → list name → tactic unit
...
-                          delta_target new_cs,
-                          intron n
-| cs l              := do new_cs ← to_qualified_names cs, l.apply (delta_hyp new_cs) (delta_target new_cs)
+meta def delta (cs : parse ident*) : parse location → tactic unit
+| loc.wildcard := do ls ← tactic.local_context,
+                     n ← revert_lst ls,
+                     new_cs ← to_qualified_names cs,
+                     delta_target new_cs,
+                     intron n
+| l            := do new_cs ← to_qualified_names cs, l.apply (delta_hyp new_cs) (delta_target new_cs)
```
leodemoura: This looks like a random cosmetic change, and it is not related to the PR description.

library/init/meta/interactive.lean
```diff
@@ -375,14 +375,51 @@ do e_type ← infer_type e >>= whnf,
...
+  end
+
+private meta def generalize_arg_p : pexpr → parser (pexpr × name)
+| (app (app (macro _ [const `eq _ ]) h) (local_const x _ _ _)) := pure (h, x)
+| _ := fail "parse error"
+
+/-- `generalize : e = x` replaces all occurrences of `e` in the target with a new hypothesis `x` of the same type.
+    `generalize h : e = x` in addition registers the hypothesis `h : e = x`. -/
+meta def generalize (h : parse ident?) (p : parse $ tk ":" *> with_desc "expr = id" (parser.pexpr 0 >>= generalize_arg_p)) : tactic unit :=
+do let (p, x) := p,
```
leodemoura: This looks horrible. I want to die.

library/init/meta/interactive.lean
digama0: Sorry, I'll remove it.

library/init/meta/interactive.lean
digama0: I'm... sorry you feel that way. This is not new code, it was just moved earlier. Is this a bad time?

leodemoura: The bar for contributed code is higher now. The code should be crystal clear. My reaction was due to `p`'s type.

library/init/meta/interactive.lean
```diff
@@ -375,14 +375,55 @@ do e_type ← infer_type e >>= whnf,
...
+with_desc "expr = id" $ do
+  e ← parser.pexpr 0,
+  match e with
+  | (app (app (macro _ [const `eq _ ]) h) (local_const x _ _ _)) := pure (h, x)
+  | _ := fail "parse error"
+  end
+
+/-- `generalize : e = x` replaces all occurrences of `e` in the target with a new hypothesis `x` of the same type.
+    `generalize h : e = x` in addition registers the hypothesis `h : e = x`. -/
+meta def generalize (h : parse ident?) (_ : parse $ tk ":") (p : parse generalize_arg_p) : tactic unit :=
```
digama0, line 403: @leodemoura Is this better? I'm torn on whether to parse `:` in a separate arg like this or use `*>` like before; this way is a bit clearer but also requires an extra dummy arg when called directly a few times below.

leodemoura: Yes, it is better. It would be even better if you add comments to `generalize_arg_p` and other functions you have added.

digama0: I added some provisional descriptions of `by_cases` and `induction`, which were modified in this PR, but I will defer to @avigad who is also working on inline documentation of the tactics here.

leodemoura:
Changes have been incorporated to master branch.

PR #1833, opened 2017-09-27
lean-settings.el: Set rootdir automatically when installed.

shlevy:


Kha:
We are in the process of moving lean-mode into its own repository/MELPA package: https://github.com/leanprover/lean-mode
Note that the new version will not need `lean-rootdir` set if it can find `lean` in the system path. Hopefully that should subsume the functionality of your PR.

shlevy:
Awesome, sounds good! What's the time frame on the move, anything I can help with?

Kha:
You are welcome to give the packages (and the docs) a try as soon as https://github.com/melpa/melpa/pull/4813 is merged, otherwise we should be set. In general, I'm always happy about contributions from people knowing elisp, because I certainly don't :) .

PR #1837, opened 2017-09-29
has_coe_to_sort: Force coercion to actually be to a sort.

shlevy:
The 'S' field was not only superfluous, it also allowed nonsense
definitions like:

```lean
example : has_coe_to_sort ℕ := ⟨ℕ, id⟩
```

With this change, having a [has_coe_to_sort s] in the context actually
allows for coercing s to a type without needing to know any concrete
details about s.


shlevy:
Note that this will require a documentation update wherever type coercions are mentioned.

shlevy:
Updated with a new test that captures the issue.

PR #1840, opened 2017-10-04
feat(init/logic): mark `id` as reducible

Kha:
As noted by @skaslev, it would be nice if `simp` could match expressions like `(λ x, x) <$> x` with `id_map`. I used the same fix in https://github.com/Kha/system_f_sub/blob/master/system_f_sub.lean#L6 and I assume there's no reason not to do so globally.

### DIFF
--- a/library/init/logic.lean
+++ b/library/init/logic.lean
@@ -8,7 +8,7 @@ import init.core
 
 universes u v w
 
-@[inline] def id {α : Sort u} (a : α) : α := a
+@[inline, reducible] def id {α : Sort u} (a : α) : α := a
 
 def flip {α : Sort u} {β : Sort v} {φ : Sort w} (f : α → β → φ) : β → α → φ :=
 λ b a, f a b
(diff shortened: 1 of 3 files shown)


leodemoura:
I don't recall why `id` is not marked as reducible, but I remember it was considered in the past.
There are many ways to workaround this issue and avoid extra reducible annotations. 
Example: we can add the lemma `id_def : id = (fun x, x)` as a `simp` rule.

Kha:
> I don't recall why id is not marked as reducible, but I remember it was considered in the past.

Ah, I see now that it was removed in 3a4cd38ba906631d2d9cbd7b1d72506c70f08576.

> Example: we can add the lemma id_def : id = (fun x, x) as a simp rule.

For readability I would prefer `id`, but I can see that the reverse simp rule may be a bad idea.

leodemoura:
> For readability I would prefer id, but I can see that the reverse simp rule may be a bad idea.

We can also add a version of `id_map` that uses `(fun x, x)` instead of `id`.
If it is too inconvenient to add these redundant lemmas, we may also add a command to create them automatically.

PR #1847, opened 2017-10-19
Allow extending structures in the current context

Kha:
Reported by @thalesant and @rlewis1988. I only fixed it for the new structure command because that code path never actually has to inspect the parent structure.


src/frontends/lean/util.cpp
```diff
@@ -164,8 +164,6 @@ bool is_local_ref(expr const & e) {
     if (!is_as_atomic(e))
         return false;
     expr const & imp_arg = get_as_atomic_arg(e);
-    if (!is_app(imp_arg))
-        return false;
```
Kha, line 168: This check fails for local refs that capture only universe levels.

PR #1851, opened 2017-10-27
fix(util/utf8): drop leading one bit

gebner:
Currently strings containing ≠ are parsed incorrectly:
```lean
#eval "≠" -- "⋠"
```

@fpvandoorn noticed this in the HoTT library: https://github.com/gebner/hott3/commit/0543f235133ae0f4ea40122ff87c4ee562cf2022


gebner:
Closing in favor of #1852, which fixes character literals as well.

PR #1852, opened 2017-10-27
Unicode fixes

Kha:


src/library/string.cpp
```diff
@@ -29,16 +29,18 @@ static expr * g_fin_mk               = nullptr;
...
     } else if (c == '\t') {
         out << "\\t";
+    } else if (c == '\\') {
+        out << "\\\\";
     } else if (in_string && c == '\"') {
         out << "\\\"";
     } else if (!in_string && c == '\'') {
         out << "\\'";
-    } else if (32 <= c && c <= 126) {
+    } else if (c >= 32 && c != 127) {
```
Kha, line 43: It would of course be preferable if we escaped any non-printable Unicode character, but that is not exactly trivial to decide.

PR #1853, opened 2017-10-29
fix(interactive): fix focus tactic.

fpvandoorn:
Previously it would fail if there was more than one goal.

Notes: 
* <s>I haven't run the test suite. Could you add the instructions for doing that to [CONTRIBUTING.md](https://github.com/leanprover/lean/blob/master/.github/CONTRIBUTING.md)?</s>
* <s>The test https://github.com/leanprover/lean/blob/master/tests/lean/interactive/focus.lean failed before this commit. Is the test suite executed correctly? It succeeds after this commit (though I haven't checked if it produces the right output, see first note).</s> (I see that in the `expected.out` file it is supposed to raise an error)


leodemoura:
@fpvandoorn Could you please add a test that exposes the problem?

fpvandoorn:
Done

PR #1857, opened 2017-11-05
chore(init/algebra/group): use a user attr for transport to additive

digama0:
This makes it easier to transport more theorems after the original set in this file, without having to restate the map all over again.

### DIFF
--- a/library/init/algebra/group.lean
+++ b/library/init/algebra/group.lean
@@ -12,155 +12,7 @@ set_option default_priority 100
 
 set_option old_structure_cmd true
 
-universe u
-variables {α : Type u}
-
-class semigroup (α : Type u) extends has_mul α :=
-(mul_assoc : ∀ a b c : α, a * b * c = a * (b * c))
-
-class comm_semigroup (α : Type u) extends semigroup α :=
-(mul_comm : ∀ a b : α, a * b = b * a)
-
-class left_cancel_semigroup (α : Type u) extends semigroup α :=
-(mul_left_cancel : ∀ a b c : α, a * b = a * c → b = c)
-
-class right_cancel_semigroup (α : Type u) extends semigroup α :=
-(mul_right_cancel : ∀ a b c : α, a * b = c * b → a = c)
-
-class monoid (α : Type u) extends semigroup α, has_one α :=
-(one_mul : ∀ a : α, 1 * a = a) (mul_one : ∀ a : α, a * 1 = a)
-
-class comm_monoid (α : Type u) extends monoid α, comm_semigroup α
-
-class group (α : Type u) extends monoid α, has_inv α :=
-(mul_left_inv : ∀ a : α, a⁻¹ * a = 1)
-
-class comm_group (α : Type u) extends group α, comm_monoid α
-
-@[simp] lemma mul_assoc [semigroup α] : ∀ a b c : α, a * b * c = a * (b * c) :=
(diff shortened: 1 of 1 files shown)


library/init/algebra/group.lean
```diff
@@ -12,302 +12,312 @@ set_option default_priority 100
...
+meta def to_additive_attr : user_attribute (name_map name) name :=
+{ name      := `to_additive,
+  descr     := "Transport multiplicative to additive",
+  cache_cfg := ⟨λ ns, ns.mfoldl (λ dict n, do
+    val ← to_additive_attr.get_param n,
+    pure $ dict.insert n val) mk_name_map, []⟩,
+  parser    := lean.parser.ident,
+  after_set := some $ λ src _ _, do
+    env ← get_env,
+    dict ← to_additive_attr.get_cache,
```
digama0, line 56: It would have been nice to have a `set_cache` function here, rather than having to rebuild the entire name map from scratch on every new theorem. Is this possible in the current architecture (that is, is possible for the user data of a user attr to be read/write rather than just derived from the list of declarations with the attribute)?

Kha: That sounds like a weird API to me. The cache is supposed to transparently memoize the output of the caching function, explicitly setting it breaks that model. Or are you thinking of it as an "unsafe" "Trust me on this, I know what I'm doing" operation?
In any case, I would be surprised if it had a measurable performance impact for this use case.

digama0: Hm, in this case it's still pure-functional, but it's hard to tell that it is because the data is coming from tactics. I think it should just be treated as an "unsafe" function, as you say. I agree that this probably won't be a considerable performance impact, it's a minor thing, but if I understand correctly the current implementation is O(n^2 log n) over n definitions, where it could be O(n log n).

library/init/algebra/group.lean
```diff
@@ -55,8 +55,13 @@ meta def to_additive_attr : user_attribute (name_map name) name :=
     env ← get_env,
     dict ← to_additive_attr.get_cache,
     tgt ← to_additive_attr.get_param src,
-    (get_decl tgt >> skip) <|>
-      transport_with_dict dict src tgt }
+    let tgt' := match tgt, src with
+    | name.mk_string _ (name.mk_string _ _), _ := tgt
+    | _, name.mk_string _ pfx := tgt.update_prefix pfx
+    | _, _ := tgt
+    end,
```
digama0: It would be nice if there was a version of `resolve_constant` that worked on names that don't exist, or a way to get the current namespace prefix at the location of the containing top-level command. There are still a lot of things that the `def` command does that are not accessible to tactics (parameters and variables too).

This commit also suggests the usage of `_root_` to disable namespace resolution. How feasible would it be to have `def` support this as well? People have asked before about how to declare things outside the current namespace, and this seems like a nice workaround.

digama0: I undid this commit because it doesn't work without also adding the prefix management to the caching part, and I don't want to mix up parsing with the translation map. The best place to do the prefix management would be in a "before_set" handler that has time to preprocess the inputs before storing them in the definition, but this isn't available at the moment. I think I will add either a user command or a second user_attribute for handling this, but that part need not be done in the lean repo.

leodemoura:
It will not be merged since `transport_multiplicative_to_additive` will be deleted in the future.

PR #1858, opened 2017-11-05
refactor(init/meta/interactive): factor out replace_at from simp_core_aux

digama0:
The `replace_at` function is independently useful, this just moves the majority of the logic for `simp_core_aux` into a function that doesn't use simp


spl:
Note that, in the diff, the definition is named `replace_at`, whereas you otherwise refer to `rewrite_at`, in case you meant to use a different name than you did.

leodemoura:
It will not be merged, refactoring was not requested by developers.

PR #1864, opened 2017-11-12
Fix debug build

Kha:
I will be very happy when we can remove the old structure code, even if this issue only impacted the new code.

### DIFF
--- a/src/frontends/lean/structure_cmd.cpp
+++ b/src/frontends/lean/structure_cmd.cpp
@@ -473,24 +473,27 @@ struct structure_cmd_fn {
      *  `elaborate_header` will build the telescope `Pi (A : Type) (B : Type), s' A B -> Type`. */
 
     using tele_elab = std::function<expr(expr)>;
+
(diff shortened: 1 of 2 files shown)


src/frontends/lean/structure_cmd.cpp
```diff
@@ -490,9 +490,24 @@ struct structure_cmd_fn {
         } else {
             expr new_tmp = elab(mk_arrow(in_header ? parent : mk_as_is(parent), tmp));
             parent       = copy_tag(parent, expr(binding_domain(new_tmp)));
+            unsigned i   = static_cast<unsigned>(&parent - &m_parents[0]);
+
```
leodemoura: This cast and pointer arithmetic looks fishy to me. Could you please clarify what is going on here? It is not obvious that `parent` is in `m_parents`.

Kha: Yeah, this is unnecessarily clever. I will change the code so that the function is passed `i` instead of `parent`.

PR #1865, opened 2017-11-12
Add aliasing patterns `id@pat`

Kha:


johoelzl:
Do you really want to allow `x@y@p`? It looks like a error to me. Is there an application where you want to allow it?

digama0:
I can't see a use for `x@y@p`, but I think it follows the logical grammar with `patt ::= id@patt`. In particular, there is definitely a use for nested pattern identifiers such as `x@(a,b,c@(d,e))`. I do this all the time in Mathematica, and am happy to see it in Lean.

Kha:
> I can't see a use for x@y@p, but I think it follows the logical grammar with patt ::= id@patt.

Yeah. Basically, I decided to allow it because Haskell allows it.

src/frontends/lean/elaborator.cpp
```diff
@@ -2478,6 +2478,10 @@ class validate_and_collect_lhs_mvars : public replace_visitor {
     virtual expr visit_macro(expr const & e) override {
         if (is_inaccessible(e)) {
             return e;
+        } else if (is_as_pattern(e)) {
+            visit(get_as_pattern_lhs(e));
+            visit(get_as_pattern_rhs(e));
+            return e;
```
Kha: I misunderstood the return value, will fix tomorrow.

leodemoura: Ok.

src/frontends/lean/elaborator.cpp
```diff
@@ -2349,7 +2349,8 @@ static void mvar_dep_sort(type_context & ctx, buffer<expr> & mvars) {
    Before processing the right-hand-side, we must convert back these metavariables into
    pattern variables (i.e., local constants).
    This visitor collects metavariables that must be converted into pattern variables, and
-   validates the equation left-hand-side. */
+   validates the equation left-hand-side. It returns the input expression expanded to
+   expose its primitive patterns. */
```
Kha, line 2353: @leodemoura This should be correct, yes?

leodemoura: Correct.

PR #1878, opened 2017-11-26
refactor(algebra/ordered_field): move theorems from discrete

digama0:
This just proves a few `discrete_linearly_ordered_field` theorems under `linearly_ordered_field` instead. The theorems themselves are unchanged, although the proofs had to be reordered and changed a bit to accomodate.


leodemoura:
merged

PR #1881, opened 2017-12-09
Monad refactorings

Kha:
Next steps:
* (Maybe) refactor existing instances
* Add missing monads: reader, id, writer
* Add (lawful) mtl-like monad classes


src/frontends/lean/notation_cmd.cpp
```diff
@@ -237,16 +238,20 @@ static auto parse_mixfix_notation(parser & p, mixfix_kind k, bool overload, nota
 #pragma GCC diagnostic ignored "-Wmaybe-uninitialized"
 #endif
             return mk_pair(notation_entry(false, to_list(transition(tks, mk_expr_action(*prec), pp_tk)),
-                                          mk_app(f, Var(1), Var(0)), overload, priority, grp, parse_only), new_token);
+                                          head_beta_reduce(mk_app(f, Var(1), Var(0))), overload, priority,
+                                          grp, parse_only), new_token);
```
Kha: Not sure whether this commit is the correct solution yet

tests/lean/run/beta_reduce_mixfix.lean
```diff
@@ -0,0 +1,2 @@
+local infixr ` $> `:100  := λ a b, b <$ a
+example (m : Type → Type) [monad m] : m ℕ := pure 0 $> 1
```
Kha: The problem is that elaboration fails without the beta reduction

leodemoura: I think the main issue is the way we elaborate `(λ ...) (pure 0) 1`. We don't use information about the type of `pure 0` and `1` when elaborating `(λ ...)`. Actually, we elaborate `(λ ...)` without any expected type information. Note that we usually propagate typing information from left to right. That is, we elaborate `f a` by first elaborating `f` and then using information about type of `f` to elaborate `a`.

The `head_beta_reduce` trick above may backfire in the future and produce unintended behavior. Example: the user is using a lambda to create a let-like expression.
 
I think the following should work
1- We define a constant `map_const_rev` as ` λ a b, b <$ a`
2- We define ` $> ` as notation for `map_const_rev`
Benefit: we preserve the user intent. That is, `a $> b` will be pretty printed as `a $> b` instead of `b <$ a`.

Kha: > I think the main issue is the way we elaborate (λ ...) (pure 0) 1.

Do you propose we should change that way? I encountered a related issue where I have to write `has_map.map id` instead of `(id <$>)` because the latter introduces an unnecessary abstraction that again interferes with elaboration.

> The head_beta_reduce trick above may backfire in the future and produce unintended behavior.

We could restrict it to do either zero or exactly two reductions, that should hopefully not surprise anyone.

> We define a constant map_const_rev as λ a b, b <$ a

I really would like to avoid that - I already find it bothersome that I have to unfold `ge` to apply `le` simp lemmas (though an `add_key_equivalence ge le` may fix that?).
The new conservative pretty printer should help in preserving the user's intent. As one more data point, Isabelle does not distinguish between `<` and `>` at all and prints them both as `<`.

leodemoura: > Do you propose we should change that way? 

No, I fear it would make the elaborator even more complicated and unpredictable.
If we decide to add new elaboration strategies, we should at least provide annotations and/or attributes for letting users choose the one they want.

> I encountered a related issue where I have to write has_map.map id instead of (id <$>) because the latter introduces an unnecessary abstraction that again interferes with elaboration.

In both cases, the problem was introduced by an auxiliary lambda abstraction.
One option is to have an annotation for instructing the elaborator to apply beta reduction.
The annotation can be something simple like an identity function named `beta`.
Then, during elaboration `((beta f) a b)` is reduced to `head_beta_reduce (f a b)`. 

> We could restrict it to do either zero or exactly two reductions, that should hopefully not surprise anyone.

I think we need to let the user decide when to use the beta reduction. This is why I suggested the `beta` trick above. Otherwise, I bet someone will be surprised, and our explanation will sound weird: we do zero or exactly two reductions.

> I really would like to avoid that - I already find it bothersome that I have to unfold ge to apply le simp lemmas (though an add_key_equivalence ge le may fix that?).

I don't see a problem here. You can add a `simp` lemma that reduces `ge` into `le`. 
This kind of lemma would be a good candidate for the new "normalizing" `simp` set we discussed last week. Other users may decide to add this lemma to the main `simp` set. 

> The new conservative pretty printer should help in preserving the user's intent.

How would we preserve the user's intent after we apply tactics such as `simp`? It sound tricky.

> As one more data point, Isabelle does not distinguish between < and > at all and prints them both as <.

As I said above, we can easily simulate this behavior by adding a new lemma to a `simp` set.
We are providing more options to users: we can keep `>` or reduce it. 
I don't see how this can be a bad thing.
BTW, Coq doesn't print them both as `<`.

Kha: Okay, I think I should get used to the Coq way. I'll introduce `map_const_rev` then.

library/init/category/lift.lean
```diff
@@ -0,0 +1,78 @@
...
+class has_monad_lift (m : Type u → Type v) (n : Type u → Type w) :=
+(monad_lift {} : ∀ {α}, m α → n α)
+
+/-- The reflexive-transitive closure of `has_monad_lift`. -/
+class has_monad_lift_t (m : Type u → Type v) (n : Type u → Type w) :=
+(monad_lift {} : ∀ {α}, m α → n α)
+
+export has_monad_lift_t (monad_lift)
+
+@[reducible] def has_monad_lift_to_has_coe {m n} [has_monad_lift_t m n] {α} : has_coe (m α) (n α) :=
```
leodemoura: Does this generic coercion work reliably? 
We have tried similar coercions in the past, and they worked very poorly.
The problem is the coercion resolution procedure. 
Issue #1402 contains more info.

library/init/category/lift.lean
```diff
@@ -0,0 +1,78 @@
...
+instance has_monad_lift_t_refl (m) : has_monad_lift_t m m :=
+⟨λ α, id⟩
+
+@[simp] lemma monad_lift_refl {m : Type u → Type v} {α} : (monad_lift : m α → m α) = id := rfl
+
+
+/-- A functor in the category of monads. Can be used to lift monad-transforming functions.
+    Based on https://hackage.haskell.org/package/pipes-2.4.0/docs/Control-MFunctor.html,
+    but not restricted to monad transformers. -/
+class monad_functor (m m' : Type u → Type v) (n n' : Type u → Type w) :=
```
leodemoura: Do you have plans for this `class`? Why do we need it?
I found very few occurrences in the rest of the PR.
I also did not find any good motivation for it in the link above.

library/init/category/lift.lean
```diff
@@ -0,0 +1,78 @@
...
+
+export monad_run (run unrun)
+
+
+/-- Lift an impure scoping function.
+    The helper functions in init.util take a thunk of a pure computation and execute it under some side
+    effect. For monads like `state_t` that are "lazy", i.e. evaluate to a closure, we have to special
+    case these operations to go inside these binders and work on the strict "core". -/
+class has_scope_impure (m : Type u → Type v) :=
+(scope_impure_opt {} {α : Type u} : (∀ {β : Type u}, thunk β → option β) → thunk (m α) → m (option α))
```
leodemoura: It is not clear to me what `scope_impure_opt f t` is supposed to do when I read this definition.
I had to look at actual instances to try to understand it.
I also had to look at `init/util.lean` to (try to) understand why you use the `option` type.

leodemoura: BTW, does `{β : Type u}` buy us anything? I think it just forces us to use `@f` when implementing the instances. AFAICT, the only place where it was useful is at the definition of `has_scope_impure id`, where you could write `f (x ())` instead of `f _ (x ())`.

library/init/category/functor.lean
```diff
@@ -22,9 +19,4 @@ infixr ` <$ `:100  := has_map.map_const
 λ a b, b <$ a
 infixr ` $> `:100  := has_map.map_const_rev
 
-class functor (f : Type u → Type v) extends has_map f : Type (max (u+1) v) :=
-(map_const_eq : ∀ {α β : Type u}, @map_const α β = map ∘ const β . control_laws_tac)
--- `functor` is indeed a categorical functor
-(id_map       : Π {α : Type u} (x : f α), id <$> x = x)
-(map_comp     : Π {α β γ : Type u} (g : α → β) (h : β → γ) (x : f α), (h ∘ g) <$> x = h <$> g <$> x)
-end
+class functor (f : Type u → Type v) extends has_map f
```
leodemoura: It seems we don't need `has_map` anymore. We can keep `functor` only.
I believe we had both because `has_map` was the lawless `functor`.

library/init/category/lawful.lean
```diff
@@ -0,0 +1,223 @@
...
+
+class is_lawful_monad (m : Type u → Type v) [monad m] extends is_lawful_applicative m : Prop :=
+(bind_pure_comp_eq_map : ∀ {α β : Type u} (f : α → β) (x : m α), x >>= pure ∘ f = f <$> x  . control_laws_tac)
+(bind_map_eq_seq : ∀ {α β : Type u} (f : m (α → β)) (x : m α), f >>= (<$> x) = f <*> x  . control_laws_tac)
+-- monad laws
+(pure_bind : ∀ {α β : Type u} (x : α) (f : α → m β), pure x >>= f = f x)
+(bind_assoc : ∀ {α β γ : Type u} (x : m α) (f : α → m β) (g : β → m γ),
+  x >>= f >>= g = x >>= λ x, f x >>= g)
+-- helper lemma
+(bind_bind_pure_comp_eq_seq : ∀ {α β : Type u} (f : m (α → β)) (x : m α), (f >>= λ f, x >>= pure ∘ f) = f <*> x :=
```
leodemoura: Why do we need a field for a helper lemma?
This seems to be a leftover when these laws had to be defined without the tactic framework.
AFAICT, `bind_bind_pure_comp_eq_seq` is used only once. We can prove it there using `have ...`.

library/init/category/lawful.lean
```diff
@@ -0,0 +1,223 @@
...
+
+export is_lawful_functor (map_const_eq id_map comp_map)
+attribute [simp] id_map
+-- `comp_map` does not make a good simp lemma
+
+class is_lawful_applicative (f : Type u → Type v) [applicative f] extends is_lawful_functor f : Prop :=
+(seq_left_eq  : ∀ {α β : Type u} (a : f α) (b : f β), a <* b = const β <$> a <*> b . control_laws_tac)
+(seq_right_eq : ∀ {α β : Type u} (a : f α) (b : f β), a *> b = const α id <$> a <*> b . control_laws_tac)
+-- applicative laws
+(pure_seq_eq_map : ∀ {α β : Type u} (g : α → β) (x : f α), pure g <*> x = g <$> x) -- . control_laws_tac)
```
leodemoura: Is the comment `--  . control_laws_tac)` a leftover or a typo?

library/init/category/lawful.lean
```diff
@@ -0,0 +1,223 @@
...
+(bind_bind_pure_comp_eq_seq : ∀ {α β : Type u} (f : m (α → β)) (x : m α), (f >>= λ f, x >>= pure ∘ f) = f <*> x :=
+  by intros; rw ←bind_map_eq_seq; simp [(bind_pure_comp_eq_map _ _).symm])
+-- all applicative laws are derivable from the monad laws + id_map
+(pure_seq_eq_map := by intros; rw ←bind_map_eq_seq; simp [pure_bind])
+(map_pure := by intros; rw ←bind_pure_comp_eq_map; simp [pure_bind])
+(seq_pure := by intros; rw ←bind_map_eq_seq; simp [map_pure, bind_pure_comp_eq_map])
+(seq_assoc := by intros; simp [(bind_pure_comp_eq_map _ _).symm,
+                                (bind_bind_pure_comp_eq_seq _ _).symm, bind_assoc, pure_bind])
+
+export is_lawful_monad (bind_pure_comp_eq_map bind_bind_pure_comp_eq_seq bind_map_eq_seq pure_bind bind_assoc)
```
leodemoura: If `bind_bind_pure_comp_eq_seq` is helper lemma, why do we need to export it?

library/init/category/lawful.lean
```diff
@@ -0,0 +1,223 @@
...
+attribute [simp] pure_bind
+
+-- monad "law" derivable from other laws
+@[simp] theorem bind_pure {α : Type u} {m : Type u → Type v} [monad m] [is_lawful_monad m] (x : m α) : x >>= pure = x :=
+show x >>= pure ∘ id = x, by rw bind_pure_comp_eq_map; simp [id_map]
+
+lemma bind_ext_congr {α β} {m : Type u → Type v} [has_bind m] {x : m α} {f : α → m β} (g : α → m β) :
```
leodemoura: Why is `f` implicit and `g` explicit?

leodemoura: In principle, both can be inferred from `(∀ a, f a = g a)`

library/init/category/lift.lean
```diff
@@ -0,0 +1,78 @@
...
+
+instance monad_functor_t_refl (m m') : monad_functor_t m m' m m' :=
+⟨λ α f, f⟩
+
+@[simp] lemma monad_map_refl {m m' : Type u → Type v} (f : ∀ {α}, m α → m' α) {α} : (monad_map @f : m α → m' α) = f := rfl
+
+
+/-- Run a monad stack to completion. -/
+class monad_run (out : out_param $ Type u → Type v) (m : Type u → Type v) :=
+(run {} {α : Type u} : m α → out α)
```
leodemoura: What does "completion" mean here?
BTW, the signature of `has_monad_lift.monad_lift` and `monad_run.run` is essentially the same.
The main difference is that `out` is marked as `out_param` here.

library/init/category/monad_fail.lean
```diff
@@ -4,12 +4,19 @@ Released under Apache 2.0 license as described in the file LICENSE.
...
-import init.category.monad init.data.string.basic
+import init.category.lift init.data.string.basic
 
 universes u v
 
 class monad_fail (m : Type u → Type v) extends monad m :=
-(fail : Π {a}, string → m a)
+(fail {} : Π {a}, string → m a)
+
+-- deriving monad from monad_fail should happen only as a last resort
```
leodemoura: why?

library/init/category/monad_fail.lean
```diff
@@ -4,12 +4,19 @@ Released under Apache 2.0 license as described in the file LICENSE.
...
+attribute [instance, priority 1] monad_fail.to_monad
 
 def match_failed {α : Type u} {m : Type u → Type v} [monad_fail m] : m α :=
-monad_fail.fail m "match failed"
+monad_fail.fail "match failed"
+
+-- TODO(Sebastian): Could take `has_monad_lift_t`, except that the refl instances will make it loop
```
leodemoura: Is it clear to you when it is safe to use `has_monad_lift_t`?

library/init/category/monad.lean
```diff
@@ -20,57 +20,13 @@ do x, y
 infixl ` >>= `:55 := bind
 infixl ` >> `:55  := has_bind.and_then
 
-section
-set_option auto_param.check_exists false
-
 class monad (m : Type u → Type v) extends applicative m, has_bind m : Type (max (u+1) v) :=
```
leodemoura, line 23: Is there any reason to keep the `has_bind` type class?
It was originally added to make sure we could use the `do` notation for structures that were not monads.
This commit https://github.com/leanprover/lean/commit/cf8dd9e75e4d3b8572677c83244a0f6058e6a992 makes `has_bind` unnecessary, right?

library/init/category/reader.lean
```diff
@@ -0,0 +1,80 @@
...
+Authors: Sebastian Ullrich
+
+The reader monad transformer for passing immutable state.
+-/
+
+prelude
+import init.category.lift init.category.id
+universes u v w
+
+structure reader_t (r : Type u) (m : Type u → Type v) (α : Type u) : Type (max u v) :=
```
leodemoura: In this file, `r` is being used to name a type and elements of this type.
For example, `⟨λ r, pure a⟩` is `⟨λ r : r, pure a⟩`. I think this is really confusing.
Could you please use `env` or a greek letter to name `r`?
Or use something different from `r` to name variables of type `r`.

library/init/category/reader.lean
```diff
@@ -0,0 +1,80 @@
...
+  protected def lift (a : m α) : reader_t r m α :=
+  ⟨λ r, a⟩
+
+  instance (m) [monad m] : has_monad_lift m (reader_t r m) :=
+  ⟨@reader_t.lift r m _⟩
+
+  protected def monad_map {r m m'} [monad m] [monad m'] {α} (f : Π {α}, m α → m' α) : reader_t r m α → reader_t r m' α :=
+  λ x, ⟨λ r, f (x.run r)⟩
+
+  instance (r m m') [monad m] [monad m'] : monad_functor m m' (reader_t r m) (reader_t r m') :=
```
leodemoura: It is clear to me why you defined `has_monad_lift m (reader_t r m)`, but why do we need this `monad_functor` instance?

library/init/category/reader.lean
```diff
@@ -0,0 +1,80 @@
...
+  ⟨@reader_t.monad_map r m m' _ _⟩
+end
+end reader_t
+
+
+/-- A specialization of `monad_lift` to `reader_t` that allows `r` to be inferred. -/
+class monad_reader_lift (r : out_param (Type u)) (m : out_param (Type u → Type v)) (n : Type u → Type w) :=
```
leodemoura: Is this a trick/workaround to allow us to use `read` even when the environment `r` is not known? That is, if you have used `[has_monad_lift m (reader_t r m)]` instead of `[monad_reader_lift r m n]` to define `read`, then `r` would have to be known to trigger type class resolution. If this is the only motivation for the existence of `monad_reader_lift`, we should document it (describe the problem being solved, use examples, etc).

library/init/category/reader.lean
```diff
@@ -0,0 +1,80 @@
...
+[functor {} : monad_functor_t (reader_t r m) (reader_t r' m) n n']
+
+attribute [instance] monad_reader_functor.mk
+local attribute [instance] monad_reader_functor.functor
+
+def with_reader_t {r r' m} [monad m] {α} (f : r' → r) : reader_t r m α → reader_t r' m α :=
+λ x, ⟨λ r, x.run (f r)⟩
+
+def with_reader {r r'} {m n n'} [monad m] [monad_reader_functor r r' m n n'] {α : Type u} (f : r' → r) : n α → n' α :=
+monad_map $ λ α, (with_reader_t f : reader_t r m α → reader_t r' m α)
```
leodemoura: Ok, I found the motivation for the `monad_functor` above.

library/init/category/reader.lean
```diff
@@ -0,0 +1,80 @@
...
+attribute [instance] monad_reader_lift.mk
+local attribute [instance] monad_reader_lift.has_lift
+
+def monad_reader_lift.read {r : Type u} {m : Type u → Type v} {n : Type u → Type w} [monad m] [monad_reader_lift r m n] : n r :=
+@monad_lift _ _ _ _ (reader_t.read : reader_t r m _)
+export monad_reader_lift (read)
+
+
+/-- A specialization of `monad_map` to `reader_t` that allows `r` to be inferred. -/
+class monad_reader_functor (r r' : out_param (Type u)) (m : out_param (Type u → Type v)) (n n' : Type u → Type w) :=
```
leodemoura: You are using the trick above here too.

library/init/category/reader.lean
leodemoura: BTW, this definition is not used anywhere in the core lib or in the test suite.

library/init/category/reader.lean
```diff
@@ -0,0 +1,80 @@
...
+The reader monad transformer for passing immutable state.
+-/
+
+prelude
+import init.category.lift init.category.id
+universes u v w
+
+structure reader_t (r : Type u) (m : Type u → Type v) (α : Type u) : Type (max u v) :=
+(run : r → m α)
+
```
leodemoura: At `state.lean`, you have `attribute [pp_using_anonymous_constructor] state_t`.
Should we also have 
```
attribute [pp_using_anonymous_constructor] reader_t
```
here?

library/init/category/state.lean
```diff
@@ -1,124 +1,131 @@
...
- end,
- pure_bind := by intros; apply rfl,
- bind_assoc := begin
-   intros, funext,
-   simp [state_bind], cases x s,
-   apply rfl
- end}
-end
+structure state_t (σ : Type u) (m : Type u → Type v) (α : Type u) : Type (max u v) :=
+(run' : σ → m (α × σ))
```
leodemoura: You have used `run` at `reader_t`. So, why did you use `run'` here?

leodemoura: Forget question above. You want to swap the arguments.

library/init/category/state.lean
```diff
@@ -1,124 +1,131 @@
...
-@[inline] def write' {σ : Type u} : σ → state σ punit :=
-λ s' s, (punit.star, s')
-end state
+namespace state_t
+section
+  variables {σ : Type u} {m : Type u → Type v}
 
-def state_t (σ : Type u) (m : Type u → Type v) [monad m] (α : Type u) : Type (max u v) :=
-σ → m (α × σ)
+  -- swap `st` and `x`
```
leodemoura: Why do you want to swap `st` and `x`?
The only motivation I found is this term here `(state_t.run s ∘ c)`.
With `run'`, you would have to write `(λ a, state_t.run' (c a) s)`.

library/init/category/state.lean
```diff
@@ -1,124 +1,131 @@
...
+    state_t σ m α → state_t σ m' α :=
+  λ x, ⟨λ st, f (x.run st)⟩
+
+  instance (σ m m') [monad m] [monad m'] : monad_functor m m' (state_t σ m) (state_t σ m') :=
+  ⟨@state_t.monad_map σ m m' _ _⟩
+
+  -- TODO(Sebastian): uses lenses as in https://hackage.haskell.org/package/lens-4.15.4/docs/Control-Lens-Zoom.html#t:Zoom ?
+  protected def zoom {σ σ' α : Type u} {m : Type u → Type v} [monad m] (get : σ → σ')
+    (set : σ' → σ → σ) (x : state_t σ' m α) : state_t σ m α :=
+  ⟨λ st, (λ p : α × σ', (p.fst, set p.snd st)) <$> x.run (get st)⟩
```
leodemoura: Could you please replace this definition with 
```
  ⟨λ st, do (a, st') ← x.run' (get st),
            pure (a, set st' st)⟩
```
?
BTW, it would be nice to give examples why `zoom` is useful.
I think it may produce performance problems since the state is not being used linearly.
For example, suppose `st` is an array, then when executing `x` and `get` we will not be able to perform destructive updates.

library/init/category/state.lean
```diff
@@ -1,124 +1,131 @@
...
- end,
- bind_assoc := begin
-   intros, funext,
-   simp [state_t_bind, state_t_return, monad.bind_assoc],
-   apply congr_arg, funext r,
-   cases r, refl
- end}
+
+/-- A specialization of `monad_lift` to `state_t` that allows `σ` to be inferred. -/
+class monad_state_lift (σ : out_param (Type u)) (m : out_param (Type u → Type v)) (n : Type u → Type w) :=
```
leodemoura: Ok, it seems you love this trick :)

library/init/category/except.lean
```diff
@@ -0,0 +1,120 @@
...
+
+prelude
+import init.category.alternative init.category.lift
+universes u v w
+
+inductive except (ε : Type u) (α : Type v)
+| error {} : ε → except
+| ok {} : α → except
+
+class monad_except (ε : out_param (Type u)) (m : Type v → Type w) :=
```
leodemoura: I'm assuming the purpose of `monad_except` is to define a generic `throw` and `catch`.
This is useful because the interface is implemented by `except_t` and `io`.
Am I correct? If yes, we should add a comment motivating this type class.
Note that, we don't have `monad_state`. That is, `get` and `put` are tied to `state_t`.

library/init/category/cont.lean
```diff
@@ -0,0 +1,42 @@
...
+Authors: Sebastian Ullrich
+
+The continuation monad transformer.
+-/
+
+prelude
+import init.category.alternative init.category.combinators init.category.lift
+universes u v w
+
+structure cont_t (r : Type u) (m : Type u → Type v) (α : Type u) : Type (max u v) :=
```
leodemoura: Is this working in progress?
We don't have a `lift` like other transformers. 
AFAICT, this transformer is not used anywhere.
If you want to keep it, we need to document, add examples to test it, etc.

library/init/category/cont.lean
```diff
@@ -0,0 +1,42 @@
...
+(run : (α → m r) → m r)
+
+attribute [pp_using_anonymous_constructor] cont_t
+
+class monad_cont (m : Type u → Type v) :=
+(call_cc {α β : Type u} : ((α → m β) → m α) → m α)
```
leodemoura: Do you expect to have other implementations of `call_cc`?

cipher1024:
I noticed that you adopted the convention that monad transformers have type `Π m [monad m], Type u → Type v`. Would it be a problem to move `[monad m]` out of there? It is an idiom and a feature that has been abandoned in Haskell because it produces un-modular code. It's more modular to put that constraint as part of the instance declarations:

```
instance [functor m] : functor (my_monad_transformer m) := ...
instance [applicative m] : applicative (my_monad_transformer m) := ...
instance [monad m] : monad (my_monad_transformer m) := ...
```

This way, you can still use the monad transformer if the underlying type is only an applicative functor -- although part of the interface may be unavailable in this situation.

For reference: https://prime.haskell.org/wiki/NoDatatypeContexts

library/init/category/lift.lean
Kha: This is not a new definition; Github gets confused by the file renaming. The definition is used in super.

library/init/category/functor.lean
Kha: I assumed it was a "functor without pure", like `has_bind` was introduced for "monads without return"

library/init/category/lawful.lean
Kha: I didn't notice the field. Removed, thanks.

library/init/category/lawful.lean
Kha: Looks like a leftover from the original definition, removed.

library/init/category/lawful.lean
Kha: It was useful in the original commit, but now it doesn't matter anymore. Changed.

Kha: Btw, I will append new commits for changes for now and squash them in the end.

library/init/category/lift.lean
Kha: Extended the doc string. Btw, I added `unrun` because it seemed like it could be useful, but I haven't needed it yet.
There is an alternative design where the stack is not run "to completion" but stopped at some inner monad `n`. This may be useful for stacks based on io, e.g. `run : state_t st (except_t e io) a -> st -> except e (io a)`.

library/init/category/monad_fail.lean
Kha: I think I was mainly annoyed about the noise this instance introduced to the class_instance trace, since it's rarely the right thing to do. I can remove it.

cipher1024:
It might be a good time to change the `lift` functions here: https://github.com/Kha/lean/blob/b6978ed1274e2937e988c6eaf46c1cf9eecffb23/library/init/category/combinators.lean#L82-L98

They don't need to rely on `monad`. `applicative` is enough and in one case, `functor` is enough.

library/init/category/monad.lean
Kha: It may still be useful if you want to use `>>=` for a type without `pure`.

library/init/category/reader.lean
Kha: I now switched to rho for `reader_t` and `cont_t`

library/init/category/reader.lean
Kha: Probably :) . Fixed.

library/init/category/state.lean
Kha: For the parser proofs, I found the goal state to be much more readable when the program comes last. But I defined my own `run` there anyway, so I removed this one now.

tests/lean/run/state.lean
```diff
@@ -0,0 +1,52 @@
...
+
+def bistate_test : state_t ℕ (state_t bool io) unit :=
+do 0 ← get, -- outer state_t wins
+   -- manual
+   tt ← monad_lift (get : state_t bool io bool),
+   -- needs to mention inner monad
+   tt ← get_type io bool,
+   pure ()
+
+#eval (bistate_test.run 0).run tt
```
leodemoura: Why didn't you use `run` (i.e., `monad_run.run`)?

library/init/category/state.lean
```diff
@@ -1,124 +1,147 @@
...
-   simp [state_t_bind, state_t_return, monad.bind_assoc],
-   apply congr_arg, funext r,
-   cases r, refl
- end}
+/-- A specialization of `monad_lift` to lifting `state_t` that allows `σ` to be inferred.
+
+    This class is roughly equivalent to `MonadState` from https://hackage.haskell.org/package/mtl-2.2.2/docs/Control-Monad-State-Class.html,
+    with the important distinction that it is automatically derived via the generic
+    `has_monad_lift` class. -/
+class monad_state_lift (σ : out_param (Type u)) (m : out_param (Type u → Type v)) (n : Type u → Type w) :=
```
leodemoura: I think we should say this is an auxiliary class, and users should not define instances for it.
My understanding is that this is a workaround to address a limitation of the type class resolution procedure.

library/init/category/state.lean
```diff
@@ -1,124 +1,147 @@
...
+  λ x, ⟨λ st, f (x.run st)⟩
+
+  instance (σ m m') [monad m] [monad m'] : monad_functor m m' (state_t σ m) (state_t σ m') :=
+  ⟨@state_t.monad_map σ m m' _ _⟩
+
+  -- TODO(Sebastian): uses lenses as in https://hackage.haskell.org/package/lens-4.15.4/docs/Control-Lens-Zoom.html#t:Zoom ?
+  protected def zoom {σ σ' α : Type u} {m : Type u → Type v} [monad m] (get : σ → σ')
+    (set : σ' → σ → σ) (x : state_t σ' m α) : state_t σ m α :=
+  ⟨λ st, do (a, st') ← x.run (get st),
+            pure (a, set st' st)⟩
```
leodemoura: The state is not being used linearly here. I think this is an anti-idiom.
Suppose we use `zoom` for zooming into an array stored in the state. 
Then, when executing `x.run` we will have two references to the array, and destructive updates will not be performed.
Do you see a way to fix this? 
Do we really need `zoom`? AFAICT, all `monad_functor` machinery is only used to define `zoom` in this PR.

library/init/category/functor.lean
leodemoura: I didn't understand your answer. In this PR, `functor` and `has_map` are morally identical and have the same fields.
They would have the same fields if we were still flattening structures.

library/init/category/lift.lean
```diff
@@ -0,0 +1,100 @@
...
+⟨λ α, id⟩
+
+@[simp] lemma monad_lift_refl {m : Type u → Type v} {α} : (monad_lift : m α → m α) = id := rfl
+
+
+/-- A functor in the category of monads. Can be used to lift monad-transforming functions.
+    Based on pipes' [MFunctor](https://hackage.haskell.org/package/pipes-2.4.0/docs/Control-MFunctor.html),
+    but not restricted to monad transformers.
+    Alternatively, an implementation of [MonadTransFunctor](http://duairc.netsoc.ie/layers-docs/Control-Monad-Layer.html#t:MonadTransFunctor). -/
+class monad_functor (m m' : Type u → Type v) (n n' : Type u → Type w) :=
```
leodemoura, line 50: AFAICT, it has a single application in the PR: `zoom`.
Do you have other (more convincing) applications for it?

library/init/category/lift.lean
```diff
@@ -0,0 +1,100 @@
...
+
+/-- Run a monad stack to completion.
+    `run` should be the composition of the transformers' individual `run` functions.
+    `unrun` should be its inverse.
+    This class mostly saves some typing when using highly nested monad stacks:
+    ```
+    @[reducible] def my_monad := reader_t my_cfg $ state_t my_state $ except_t my_err id
+    -- def my_monad.run {α : Type} (x : my_monad α) (cfg : my_cfg) (st : my_state) := ((x.run cfg).run st).run
+    def my_monad.run {α : Type} (x : my_monad α) := monad_run.run x
+    ```
```
leodemoura, line 78: This example helped. However, I tried to use `monad_run.run` in one of the tests, and it didn't work :(
BTW, it is still not clear to me why `unrun` is useful. AFAICT, it is dead code.

library/init/category/lift.lean
leodemoura: This type class is still quite obscure to me. I am inclined to sacrifice `init.util` to avoid this extra complexity.
We can provide the functions defined at `init.util` as actions in the `tactic` and `io` monads.

library/init/category/lift.lean
leodemoura: Ok, it is also used to implement `with_reader`.

tests/lean/run/state.lean
leodemoura: I tried, but I got a weird error message.

library/init/category/cont.lean
```diff
@@ -0,0 +1,56 @@
...
+universes u v w
+
+/-- An implementation of [ContT](https://hackage.haskell.org/package/mtl-2.2.2/docs/Control-Monad-Cont.html#t:ContT) -/
+structure cont_t (ρ : Type u) (m : Type u → Type v) (α : Type u) : Type (max u v) :=
+(run : (α → m ρ) → m ρ)
+
+attribute [pp_using_anonymous_constructor] cont_t
+
+/-- An implementation of [MonadCont done right](https://wiki.haskell.org/MonadCont_done_right) -/
+class monad_cont (m : Type u → Type v) :=
```
leodemoura: Why do we need this type class?
I am assuming we have `monad_except` because we want to have a general `throw` and `catch` signature.
AFAICT, we will not have multiple implementations of `call_cc`. 
Note that we don't have a general `put`, `get` and `read` operations.

library/init/category/cont.lean
```diff
@@ -0,0 +1,56 @@
...
+The continuation monad transformer.
+-/
+
+prelude
+import init.category.alternative init.category.combinators init.category.lift
+universes u v w
+
+/-- An implementation of [ContT](https://hackage.haskell.org/package/mtl-2.2.2/docs/Control-Monad-Cont.html#t:ContT) -/
+structure cont_t (ρ : Type u) (m : Type u → Type v) (α : Type u) : Type (max u v) :=
+(run : (α → m ρ) → m ρ)
```
leodemoura: Do you see an application for `cont_t`? 
Code based on `call_cc` is usually quite cryptic, but some systems provide a native and reasonably efficient implementation. This is not the case here. Code using this `call_cc` will be very inefficient since closures are created all over the place.
My point is do we need it in the core lib? It feels it is in the category: "we can do it, but please don't use it".

library/init/category/cont.lean
```diff
@@ -0,0 +1,56 @@
...
+  instance : monad (cont_t ρ m) :=
+  { pure := @pure, bind := @bind }
+
+  protected def call_cc {α : Type u} (f : (∀ {β}, α → cont_t ρ m β) → cont_t ρ m α) : cont_t ρ m α :=
+  ⟨λ cc, (f (λ _ a, ⟨λ _, cc a⟩)).run cc⟩
+
+  instance : monad_cont (cont_t ρ m) :=
+  ⟨@call_cc⟩
+
+  protected def lift [_root_.monad m] {α : Type u} (x : m α) : cont_t ρ m α :=
```
leodemoura: You had to use `_root_.monad` because Lean has used `cont_t.monad` to name the instance `instance : monad (cont_t ρ m)`. This is quite unfortunate here. We can avoid this issue by marking all instances as protected.

tests/lean/run/state.lean
Kha: I just tried it, first we need the missing instance
```
instance : monad_run io io :=
⟨λ _, id, λ _, id⟩
```
(ie. "`io` cannot be run any further")
Then we can do
```
#eval let r := run bistate_test in r 0 tt
```
Note that we need the let to make class inference infer the function type before we can further apply it. So yes, it's a bit awkward to use. It's better when used [unapplied](https://github.com/Kha/lean/blob/tactic_trans/library/init/meta/interaction_monad.lean#L56), but I'm okay with removing it.

library/init/category/state.lean
Kha: I'm not sure I would call it a workaround; I can't think of a better solution in a cleverer typeclass system. I'm deriving instances of it for encapsulation https://github.com/Kha/lean/blob/tactic_trans/library/init/meta/interaction_monad.lean#L52

library/init/category/state.lean
Kha: I believe that in more complex monad-based code functions like `zoom` or `with_reader` for transforming monad stacks will be essential for gluing together different parts of the program. We can provide a specialized, linear version taking `(get : σ → σ') (set : σ' → σ)`, but the general `zoom` will be needed much more often I think. Usually you carry around some big state and want to give just a part of it to a subroutine.

Kha: There may also be even more exotic signatures like `(get : σ → σ' → σ) (set : σ' → σ → σ)` where `get` is supposed to return a new σ with a "hole" where σ' was extracted. I wonder if that is a thing, like "linear lenses"?

library/init/category/cont.lean
Kha: cont has some very special properties, like not having any possible implementation of monad_functor, or it being a monad without `m` having to be one. It therefore felt incomplete to omit it, but you are right of course that it probably won't be very useful. Should we move it to a test?

library/init/category/state.lean
leodemoura: > We can provide a specialized, linear version taking (get : σ → σ') (set : σ' → σ)

This version is great for the zoom-out case. You have one example in the test folder where this signature would work well.

library/init/category/cont.lean
Kha: They are protected, this is the `protected` + `parameter` name resolution issue

library/init/category/state.lean
cipher1024: @Kha I think you could accomodate that use case and manipulate the state linearly by combining the getter and setter into a lens: `lens : Π f [functor f], (σ' → f σ') → (σ → f σ)`. If that lens uses the state linearly -- I'm not sure how to make one that is computable but isn't linear -- the whole thing should be linear. And that lens is as expressive as a pair of getter and setter.

leodemoura: > There may also be even more exotic signatures like (get : σ → σ' → σ) (set : σ' → σ → σ)

It should be `get : σ → (σ' x σ)`, right? 
If yes, then I can see it working. For the array example, `get` would retrieve the array from the state and return the state `st : σ` updated with an empty array. Thus, we would have a single reference to the array. Then, `set` would update the state returned by `get` with the array produced by `x.run`.
We could also have `(get : σ → (σ' x σ'')) (set : σ' → σ'' → σ)`, right?

library/init/category/cont.lean
leodemoura: Oh no... :(

library/init/category/functor.lean
Kha: Sorry, I got my monad hierarchy mixed up. I'll remove it.

Kha:
@cipher1024 
> I noticed that you adopted the convention that monad transformers have type `Π m [monad m], Type u → Type v`.

No, all types have been changed in later commits :) .

> It might be a good time to change the lift functions here: https://github.com/Kha/lean/blob/b6978ed1274e2937e988c6eaf46c1cf9eecffb23/library/init/category/combinators.lean#L82-L98

These lifts are older than `functor`/`applicative`, do we still need them/do Haskell people still use them? The duplication and indices look messy compared to `<$>` and `<*>`.

library/init/category/cont.lean
leodemoura: > Should we move it to a test?

A test would be great.
In the future, we can have an example folder where we can put this kind of code.

leodemoura:
> These lifts are older than functor/applicative, do we still need them/do Haskell people still use them? The duplication and indices look messy compared to <$> and <*>.

@kha let's delete this crap. If we people need them, they can define themselves.

library/init/category/state.lean
Kha: @cipher1024 I'm afraid my optics are not up to par to imagine what that would look like. Anyway, this may not be the best PR to introduce a lense framework :) .

Kha: > It should be get : σ → (σ' x σ), right?

Yes, sorry. 

> We could also have (get : σ → (σ' x σ'')) (set : σ' → σ'' → σ), right?

Yes. So, which one(s) to choose? :)

leodemoura: > Yes. So, which one(s) to choose? :)

I think we can derive all variants from `(get : σ → (σ' x σ'')) (set : σ' → σ'' → σ)`, right?

cipher1024:
> These lifts are older than functor/applicative, do we still need them/do Haskell people still use them? The duplication and indices look messy compared to <$> and <*>.

> @Kha let's delete this crap. If we people need them, they can define themselves.

Thanks for the heads up! I'll bring it over to `mathlib`. @Kha, `lift₂` is the one I see most often and it's mostly useful curried when you're feeding it to a higher order function. It's also useful when monadic actions return monadic actions: with `x : m (m α), y : m (m β), f : α → β → γ`, you can write `lift₂ f <$> x <*> y` to combine the results of `x` and `y`.

cipher1024:
@Kha In `data.buffer.parser`, beside the use of `unit`, is there anything that constrains you to stay in universe 0?

Kha:
@cipher1024 I don't think so, but I don't want to touch that type more than necessary in this PR. It should be moved from buffer to the now efficient string type anyway.

tests/lean/run/reader.lean
```diff
@@ -0,0 +1,11 @@
+import system.io
+
+def lifted_test : state_t ℕ (reader_t ℕ io) unit :=
+do 0 ← get,
+   1 ← read,
+   with_reader (λ n, (n, 2)) ((do p ← read,
```
leodemoura: `with_reader`  ==> `adapt_reader`

library/init/category/reader.lean
```diff
@@ -0,0 +1,118 @@
...
+  instance [alternative m] : alternative (reader_t ρ m) :=
+  { failure := @reader_t.failure _ _ _ _,
+    orelse  := @reader_t.orelse _ _ _ _ }
+end
+end reader_t
+
+
+/-- An implementation of [MonadReader](https://hackage.haskell.org/package/mtl-2.2.2/docs/Control-Monad-Reader-Class.html#t:MonadReader).
+    It does not contain `local` because this function cannot be lifted using `monad_lift`.
+    Instead, the `monad_reader_adapter` class provides the more general `with_reader` function.
```
leodemoura: `with_reader` has been renamed (it is now called `adapt_reader`).

library/init/category/reader.lean
```diff
@@ -0,0 +1,118 @@
...
+
+section
+variables {ρ ρ' : Type u} {m m' : Type u → Type v}
+
+instance monad_reader_adapter_trans {n n' : Type u → Type v} [monad_functor m m' n n'] [monad_reader_adapter ρ ρ' m m'] : monad_reader_adapter ρ ρ' n n' :=
+⟨λ α f, monad_map (λ α, (adapt_reader f : m α → m' α))⟩
+
+instance [monad m] : monad_reader_adapter ρ ρ' (reader_t ρ m) (reader_t ρ' m) :=
+⟨λ α, reader_t.adapt⟩
+end
```
leodemoura, line 118: Should we add the following instance here?
```lean
instance (ε) [monad m] [monad_except ε m] : monad_except ε (reader_t ρ m) :=
{ throw := λ α, reader_t.lift ∘ throw,
  catch := λ α x c, ⟨λ r, catch (x.run r) (λ e, (c e).run r)⟩ }
```

leodemoura:
merged

PR #1894, opened 2017-12-21
fix(algebra/group): remove inherited [simp] from add_comm

digama0:
It looks like this is a bug in the implementation of `transport_multiplicative_to_additive` - at the time of the call, `mul_comm` has `[simp]` enabled locally, but when it gets copied to `add_comm` it becomes permanent.

A test:
 `#print add_comm`
```
@[no_rsimp, priority 100, simp, priority 100]
theorem add_comm : ∀ {α : Type u} [_inst_1 : add_comm_semigroup α] (a b : α), a + b = b + a :=
λ {α : Type u} [_inst_1 : add_comm_semigroup α], add_comm_semigroup.add_comm
```
I hope I'm not off base with this change, which is intended to finish the job of b7322e2. But be aware it has some fallout for dependent projects.


leodemoura:
I created an issue for the bug. I will fix it after I return from vacation. 
I'm closing this PR until the bug is fixed and we can avoid workarounds.

PR #1901, opened 2018-01-04
leanpkg: docs and tweaks

Kha:
Main feature is the `leanpkg help <command>` command, which should explain all important details of each command.


Kha:
Corresponding reference PR: https://github.com/leanprover/reference/pull/12

library/init/data/string/ops.lean
```diff
@@ -0,0 +1,48 @@
...
+  have it.next_to_string.length - 1 < it.next_to_string.length,
+    from nat.sub_lt (iterator.zero_lt_length_next_to_string_of_has_next h) dec_trivial,
+  if p it.curr then
+    let rest := it.next.next_to_string in
+    it.prev_to_string :: split_core rest.mk_iterator
+  else
+    split_core it.next
+else
+  [it.to_string]
+using_well_founded { rel_tac := λ _ _, `[exact ⟨_, measure_wf (λ it, it.next_to_string.length)⟩] }
```
Kha: @leodemoura This was an interesting test case for iterators. To achieve linear runtime, we would either need a function for extracting the string between two iterators, or make `iterator.next_to_string` O(1) (using sharing instead of copying).

leodemoura: I try to address this performance problem this weekend. As soon as I manage to add refl lemma support to `type_context`: https://github.com/leodemoura/lean/commit/2a7fed380450b8146acce7f57d600a243a94bdb8

PR #1902, opened 2018-01-04
Replace library_dev with mathlib

kbuzzard:
# Pull Request Description

It's been bugging me for months that the docs for leanpkg still refer to library_dev but now seems like a good time to change it.

### DIFF
--- a/leanpkg/README.md
+++ b/leanpkg/README.md
@@ -17,8 +17,8 @@ Dependencies can be fetched from git repositories; in this case `leanpkg` will r
 
 You need to run `leanpkg configure` first, in order to download dependencies and generate the `leanpkg.path` file.
 ```
-git clone https://github.com/leanprover/library_dev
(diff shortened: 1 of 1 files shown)


kim-em:
While you're fixing the docs, you should change the paragraph

> By default source files are assumed to be directly in the root directory of the package.  You can optionally add a `path = "src"` attribute to the `[package]` section that selects a different directory.

to 

> Source files are assumed to be in the `src/` directory. You can modify this by adding a `path = "other"` attribute to the `[package]` section of `leanpkg.toml`.

Sebastian just changed the behaviour of `leanpkg` in his recent pull request: https://github.com/leanprover/lean/pull/1901.

kbuzzard:
Closed after conversation with @Kha

Kha:
@semorrison see https://github.com/leanprover/reference/pull/12

PR #1903, opened 2018-01-05
fix(init/data/nat/lemmas): generalize nat.iterate to Sort*

digama0:


library/init/data/nat/lemmas.lean
```diff
@@ -874,7 +874,7 @@ begin
   apply le_div_iff_mul_le _ _ Hk
 end
 
-def iterate {α : Type} (op : α → α) : ℕ → α → α
+def iterate {α : Sort*} (op : α → α) : ℕ → α → α
```
Kha: I believe we try to use explicit universes in the core lib

digama0: This is the first I've heard of this, at least as regards polymorphic functions. Is there a situation where this causes a problem?

Kha: It's just about being explicit and making sure the universe parameter isn't accidentally fixed via unification. I checked, there is just a single use of `Sort/Type*`, in `init.native`.
Also, you don't have to close the PR. Just amend the commit and force-push to the same branch to update it.

digama0: I'm confused. I closed the PR because if a universe parameter isn't wanted here, then there's nothing to push. Are you saying that it should be `Sort u` instead of `Sort*`?

PR #1904, opened 2018-01-07
feat(util): add math alphanum symbol Latin letter blocks

ratmice:
Add the latin letter blocks from,
https://unicode.org/charts/PDF/U1D400.pdf
to is_letter_like() so they may be bound to variables.
-- 
Not sure if this would be desired, or the correct way to go about this.
If not, by all means close.

I was trying to match the notation of some variables used in a textbook,
But was unable to figure out a way to use unicode symbols using the `notation` command as variable names.
trying to figure out how alpha was handled for variable names, lead to the attached patch.

### DIFF
--- a/doc/changes.md
+++ b/doc/changes.md
@@ -105,6 +105,10 @@ master branch (aka work in progress branch)
 
 * Add interactive `subst_vars` tactic.
 
+* Allow the Script, Double-struck, and Fractur symbols from
+  Mathematical Alphanumeric Symbols: https://unicode.org/charts/PDF/U1D400.pdf
+  to be used as variables Example: `variables 𝓞 : Prop`.
+
 *Changes*
 
 * Replace `inout` modifier in type class declarations with `out_param` modifier.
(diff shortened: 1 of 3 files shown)


Kha:
Thank you for your contribution. This is the right way to implement it, but I think we should be more selective in the character sets. I can agree with adding script, double-struck and Fraktur symbols, but monospace symbols are probably not a good idea. The same may go for italic, bold, and sans-serif.

ratmice:
@Kha thanks, now that you mention it that I can see how those would be problematic.
I limited it to just the symbols you mentioned, and took the opportunity to add the unicode URL to the changes.md.

PatrickMassot:
Note (not only to @ratmice): if this nice PR goes through, someone should update https://github.com/leanprover/vscode-lean/blob/master/translations.json as well.

I'm sure @kbuzzard will be delighted to have 𝓞 available for his scheme formalization.

Kha:
(Just to avoid confusion, the PR was closed because it was merged manually :) )

PR #1909, opened 2018-01-22
fix Cygwin build

nunoplopes:
It compiles cleanly with this patch.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,6 +10,7 @@ GSYMS
 GTAGS
 .projectile
 .lean_options
+.vs
 src/emacs/dependencies
 compile_commands.json
 *.idea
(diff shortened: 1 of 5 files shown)


leodemoura:
Hi Nuno, I'm using msys2 instead of cygwin. It is simpler to install and use, and the result is a real Windows application. That is, the user doesn't need the cygwin DLL. We stopped the cygwin support last year. Of course, if you are willing to maintain the cygwin support, we can keep it :)

nunoplopes:
I'm happy to support Cygwin & MSVC!

leodemoura:
> I'm happy to support Cygwin & MSVC!

Great! Regarding MSVC, I tried a few years ago, but several C++11 features were still missing.
It would be great to be able to compile Lean using it. I think we would still need to install msys2 or cygwin to get some of the command line tools used to build lean. For sure, we need them to run the test suite.

nunoplopes:
These days MSVC has quite good support for C++11, 14 & 17.  We use MSVC 2017 to build our C++17 project (we only need a little hack for gcc 7, not for MSVC).

PR #1911, opened 2018-01-23
[WIP] Add initial MSVC support

nunoplopes:
Still doesn't build fully, but at least Intellisense sort of works now.

There are a bunch of compiler errors that we need to fix, but it is already able to compile a bunch of files.
Note: compilation with reduced warnings (/W1) for now, since higher levels are too noisy for now. Let's get it to build first :)


leodemoura:
I will keep this PR open until we can compile Lean using MSVC. I will also add the `[WIP]` modifier to the title.

leodemoura:
BTW, we should change the commit message to `feat(built): ...` the commit is not fixing anything, it is adding a feature.

src/CMakeLists.txt
```diff
@@ -164,7 +164,7 @@ if(VM_UNCHECKED)
   set(LEAN_EXTRA_CXX_FLAGS "${LEAN_EXTRA_CXX_FLAGS} -D LEAN_VM_UNCHECKED")
 endif()
 
-if(AUTO_THREAD_FINALIZATION)
+if(AUTO_THREAD_FINALIZATION AND NOT MSVC)
   set(LEAN_EXTRA_CXX_FLAGS "${LEAN_EXTRA_CXX_FLAGS} -D LEAN_AUTO_THREAD_FINALIZATION")
```
leodemoura, line 178: Why did you have to disable this option when using VS?
BTW, this option is relevant when creating the lean shared object (aka DLL on Windows). 
This DLL is not essential. It is only relevant for users that want to embed Lean into their applications.

src/cmake/Modules/FindGMP.cmake
```diff
@@ -4,7 +4,7 @@ if (GMP_INCLUDE_DIR AND GMP_LIBRARIES)
 endif (GMP_INCLUDE_DIR AND GMP_LIBRARIES)
 
 find_path(GMP_INCLUDE_DIR NAMES gmp.h )
-find_library(GMP_LIBRARIES NAMES gmp libgmp REQUIRED)
+find_library(GMP_LIBRARIES NAMES gmp libgmp mpir REQUIRED)
 #find_library(GMPXX_LIBRARIES NAMES gmpxx libgmpxx )
```
leodemoura, line 8: `mpir` is only needed on Windows, right?

src/kernel/normalizer_extension.cpp
```diff
@@ -54,7 +54,7 @@ class comp_normalizer_extension : public normalizer_extension {
     }
 };
 
-std::unique_ptr<normalizer_extension> compose(std::unique_ptr<normalizer_extension> && ext1, std::unique_ptr<normalizer_extension> && ext2) {
+std::unique_ptr<normalizer_extension> compose_ext(std::unique_ptr<normalizer_extension> && ext1, std::unique_ptr<normalizer_extension> && ext2) {
     return std::unique_ptr<normalizer_extension>(new comp_normalizer_extension(std::move(ext1), std::move(ext2)));
```
leodemoura, line 58: Is `compose` a reserved keyword on VS?

src/library/module.cpp
```diff
@@ -521,13 +521,13 @@ environment add_inductive(environment                       env,
                           inductive::inductive_decl const & decl,
                           bool                              is_trusted) {
     pair<environment, certified_inductive_decl> r = inductive::add_inductive(env, decl, is_trusted);
-    environment new_env            = r.first;
-    certified_inductive_decl cdecl = r.second;
+    environment new_env             = r.first;
+    certified_inductive_decl cidecl = r.second;
     module_ext ext = get_extension(env);
```
leodemoura, line 526: I imagine `cdecl` is a reserved keyword on VS, right? :)

src/library/vm/vm.h
```diff
@@ -48,7 +49,7 @@ class vm_obj_cell {
 };
 
 #define LEAN_VM_IS_PTR(obj) ((reinterpret_cast<size_t>(obj) & 1) == 0)
-#define LEAN_VM_BOX(num)    (reinterpret_cast<vm_obj_cell*>((num << 1) | 1))
+#define LEAN_VM_BOX(num)    (reinterpret_cast<vm_obj_cell*>((uintptr_t)((num << 1) | 1)))
 #define LEAN_VM_UNBOX(obj)  (reinterpret_cast<size_t>(obj) >> 1)
```
leodemoura: Why is this needed?

src/library/vm/vm_io.cpp
```diff
@@ -437,7 +447,8 @@ static vm_obj io_get_env(vm_obj const & k, vm_obj const &) {
 }
 
 static vm_obj io_get_cwd(vm_obj const &) {
-    auto cwd = getcwd(nullptr, 0);
+    char buffer[PATH_MAX];
+    auto cwd = getcwd(buffer, sizeof(buffer));
     if (cwd) {
```
leodemoura, line 452: I saw you included extra files for Linux, what about OSX?

src/util/debug.h
```diff
@@ -34,58 +34,66 @@ Author: Leonardo de Moura
...
-#define lean_cond_assert(TAG, COND, ARGS...) DEBUG_CODE({if (lean::is_debug_enabled(TAG) && LEAN_UNLIKELY(!(COND))) { lean::notify_assertion_violation(__FILE__, __LINE__, #COND); LEAN_DISPLAY(ARGS); lean::invoke_debugger(); }})
-#define lean_always_assert(COND, ARGS...) { if (LEAN_UNLIKELY(!(COND))) { lean::notify_assertion_violation(__FILE__, __LINE__, #COND); LEAN_DISPLAY(ARGS); lean_unreachable(); } }
+#define lean_assert(COND, ...) DEBUG_CODE({if (LEAN_UNLIKELY(!(COND))) { lean::notify_assertion_violation(__FILE__, __LINE__, #COND); LEAN_DISPLAY(__VA_ARGS__); lean::invoke_debugger(); }})
+#define lean_cond_assert(TAG, COND, ...) DEBUG_CODE({if (lean::is_debug_enabled(TAG) && LEAN_UNLIKELY(!(COND))) { lean::notify_assertion_violation(__FILE__, __LINE__, #COND); LEAN_DISPLAY(__VA_ARGS__); lean::invoke_debugger(); }})
+#define lean_always_assert(COND, ...) { if (LEAN_UNLIKELY(!(COND))) { lean::notify_assertion_violation(__FILE__, __LINE__, #COND); LEAN_DISPLAY(__VA_ARGS__); lean_unreachable(); } }
```
leodemoura, line 96: How confident are you about the code above? It was hard to understand before, now it is even more incomprehensible. 
The previous version was based on a "recipe" I found. I imagine that the `ARGS...` feature is not available on VS.

src/util/sexpr/format.h
```diff
@@ -302,7 +302,6 @@ bool get_pp_unicode(options const & o);
 format pp(name const & n);
 
 /** \brief Format a S-expression */
-format pp(sexpr const & s, options const & o);
 format pp(sexpr const & s);
```
leodemoura, line 305: why did you have to remove it?

leodemoura:
Are you rebasing this PR on top of the current master as you keep making progress?

src/cmake/Modules/FindGMP.cmake
nunoplopes: Yes, it's a replacement for gmp. It's available from vcpkg.

src/kernel/normalizer_extension.cpp
nunoplopes: No, but it clashes with the other compose() functions. VS gets confused in some files and can't infer which one to call, so I had to change the name.

src/library/module.cpp
nunoplopes: yep.

src/library/vm/vm.h
nunoplopes: because num is an unsigned, and thus 32-bits in VS (and others). When doing a reinterpret_cast between types of different sizes, VS gives an error. uinptr_t is the integer type with the same size as pointers, so the problem is fixed.

src/library/vm/vm_io.cpp
nunoplopes: I'm watching the build bots. I'll make sure they are all green.

src/util/debug.h
nunoplopes: Yes ARGS... is not available in VS.  I'm quite confident on these changes :)
Also, they pass all tests and compile just fine.

src/util/sexpr/format.h
nunoplopes: This one was just a "free" cleanup. This function is not defined anywhere; this declaration was left behind and VS flags declarations without definitions.

src/CMakeLists.txt
nunoplopes: AFAIU, when AUTO_THREAD_FINALIZATION is disabled, the code fallbacks to using __thread variables instead of using pthreads. That code was already written to be compatible with MSVC and compiles fine.
So nothing is disabled AFAICT; it's just it's using a non-pthread implementation.  I think the name of the cmake option is a bit misleading.

nunoplopes:
> Are you rebasing this PR on top of the current master as you keep making progress?

Yes, I've been rebasing it. It's synced with master as of this morning.

src/library/vm/vm.h
leodemoura: I see. We use `static_cast<size_t>(n)` for this purpose in the Lean code base.
Could you please replace it with this?

src/cmake/Modules/FindGMP.cmake
leodemoura: Will it affect Linux or OSX machines that do not have `mpir`?

src/library/vm/vm_io.cpp
leodemoura: Thanks!

nunoplopes: btw this also fixes a memory leak here. The buffer returned by getcwd wasn't being free().

src/cmake/Modules/FindGMP.cmake
nunoplopes: No. It's an OR.  If gmp is installed, it is preferred.

src/library/vm/vm.h
nunoplopes: ok, done

leodemoura: BTW, I have been using `size_t` instead of `uintptr_t`, and always assumed that a pointer would fit into `size_t`, but I just realized this is not true. https://stackoverflow.com/questions/1572099/size-t-is-pointer-size-in-practice
It is not a big issue since we don't support these architectures anyway, but I will replace the offending `size_t` instances.

leodemoura: https://stackoverflow.com/questions/1464174/size-t-vs-uintptr-t

leodemoura: This is another link showing that I should not assume a pointer fits into `size_t`

nunoplopes: yep, uintptr_t was created exactly for this situation:

http://en.cppreference.com/w/c/types/integer
uintptr_t: unsigned integer type capable of holding a pointer

nunoplopes:
Lean now builds successfully on x86 & x64.  There are still a few warnings, but at least it compiles fine.
Tests don't run because of missing bash.
I've tried a few simple examples and they work :)

leodemoura:
> I've tried a few simple examples and they work :)

Nice.

nunoplopes:
From my side, this is ready to be merged.  There are a few things left to do, like cleaning up a few warnings, increase warning level to at least /W2 would be nice, and enable testing from within VS, but otherwise seems to work fine.

leodemoura:
I'm getting a machine with the latest Visual Studio.
I will test, and merge.

Kha:
Would it make sense to add it to the AppVeyor test matrix? It looks like VS2017 is supported.

leodemoura:
> Would it make sense to add it to the AppVeyor test matrix? It looks like VS2017 is supported.

If you know how to do it, then Yes :)

Kha:
I've never even used MSVC before, but sure, I can give it a try :)
(probably best in a separate PR, where I can repeatedly force-push until it works)

doc/make/msvc.md
```diff
@@ -0,0 +1,40 @@
...
+        {
+          "name": "CMAKE_TOOLCHAIN_FILE",
+          "value": "C:\\path\\to\\vcpkg\\scripts\\buildsystems\\vcpkg.cmake"
+        }
+      ]
+```
+
+
+## Enable Intellisense
+
```
leodemoura, line 33: Which version of Visual Studio are you using?
I tried with VS 2017 and Win 10. Everything works up to this point.
`Ctrl+Q` does nothing on my version. `F7` also didn't work.
I had to select `CMake > Build all` to build Lean.

nunoplopes: I also have VS 2017 (15.5)
Ctrl+Q just places the focus of the cursor in the search box at the top right of the windows; it doesn't do anything on its own.
I found a bug in VS that makes this Ctrl+Q feature not always find these cmake json files. I've reported it already.

leodemoura: > Ctrl+Q just places the focus of the cursor in the search box at the top right of the windows; it doesn't do anything on its own.

Got it. I found the search box :)

BTW, it is still building, it is super slow :)

leodemoura:
merged

PR #1913, opened 2018-01-27
fix(library/init/meta/interactive): do not look up iff.intro in current namespace

gebner:


Kha:
Damn these silly unhygienic tactic languages :) .

gebner:
This was fun to debug since the error messages were just as unhygienic: "cannot unify `... ↔ ...` with `?m_1 ↔ ?m_2`.

Of course the first `↔` was `_root_.iff`, while the second one was `hott.iff` (`↔` is a local notation here, the pretty printer probably shouldn't use it for `_root_.iff`).  For extra frustration, the options were reset when calling the tactic, so `set_option pp.all true` did nothing at all.

leodemoura:
I am merging a different fix, where I add `_root_.` before `propext` and `iff.intro`, and document why the workaround is needed. In the new parser, we should fix how names in nested interactive tactic blocks
```
`[ ... ]
```
are resolved. These blocks are used to write non interactive tactics, and are very convenient.

PR #1914, opened 2018-01-27
Various improvements

Kha:
Accumulated commits from #1881 that do not actually depend on anything monadic.


tests/lean/run/subobject_from_source.lean
```diff
@@ -0,0 +1,8 @@
+structure foo :=
+(a b : ℕ := 2)
+
+structure bar :=
+(foo : foo)
+(c : ℕ)
+
+example (b : bar) : b.foo = {c := 2, ..b}.foo := rfl
```
leodemoura: I tried this test on the `master` branch, and it works.
I don't know what this commit is fixing. To be honest, I have a hard time reviewing commits that modify `visit_structure_instance`. This method has more than 300 lines. If you have time, please break it in smaller methods and add comments.

src/frontends/lean/definition_cmds.cpp
```diff
@@ -249,6 +249,9 @@ declare_definition(parser & p, environment const & env, decl_cmd_kind kind, buff
     } else {
         c_real_name = get_namespace(env) + c_name;
     }
+    if (env.find(c_real_name)) {
+        throw exception(sstream() << "invalid definition, a declaration named '" << c_real_name << "' has already been declared");
+    }
     if (val && !meta.m_modifiers.m_is_meta && !type_checker(env).is_prop(type)) {
```
leodemoura, line 255: Great! I hate that weird kernel exception too.

src/frontends/lean/definition_cmds.cpp
```diff
@@ -863,12 +863,15 @@ environment single_definition_cmd_core(parser & p, decl_cmd_kind kind, cmd_meta
     } catch (throwable & ex) {
         // Even though we catch exceptions during elaboration, there can still be other exceptions,
         // e.g. when adding a declaration to the environment.
-        try {
-            auto res = process(p.mk_sorry(header_pos, true));
+
+        // prefer elaboration error messages
+        if (!lt.get().has_entry_now(is_error_message))
             p.mk_message(header_pos, ERROR).set_exception(ex).report();
```
leodemoura: I'm not really sure why this change is needed.

src/frontends/lean/definition_cmds.cpp
```diff
@@ -863,12 +863,15 @@ environment single_definition_cmd_core(parser & p, decl_cmd_kind kind, cmd_meta
...
-            return res;
-        } catch (...) {}
-        throw;
+        try {
+            return process(p.mk_sorry(header_pos, true));
+        } catch (...) {
+            return env;
```
leodemoura: This is not a problem with this commit, the existing code was already using `catch (...)`. It is bad practice to use `catch (...)`. If we really need to use we should document why we need it here, and why `catch (exception &)` is not enough (i.e., only catch lean exceptions).

src/frontends/lean/pp.cpp
```diff
@@ -1460,7 +1460,7 @@ auto pretty_fn::pp_notation(notation_entry const & entry, buffer<optional<expr>>
                     result e_r   = pp_notation_child(e, token_lbp, a.rbp());
                     curr = tk_fmt + e_r.fmt();
                     if (last)
-                        last_rbp = a.rbp();
+                        last_rbp = std::min(a.rbp(), e_r.rbp());
                     break;
```
leodemoura: It would be nice to document why we need `std::min`, and the long discussion we had on Slack.
We discussed other possible fixes, including the "ideal one". I think all this information is important to keep.

leodemoura: Forget previous comment, I wrote it before I reviewed the following commit.

src/library/constructions/projection.cpp
```diff
@@ -190,8 +191,14 @@ environment mk_projections(environment const & env, name const & n, buffer<name>
...
-        expr param = mk_local(mk_fresh_name(), binding_name(intro_type), binding_domain(intro_type),
-            binding_info(intro_type).is_inst_implicit() ? mk_inst_implicit_binder_info() : binder_info());
+        auto bi = binding_info(intro_type);
+        if (!bi.is_inst_implicit())
+            // We reset implicit binders so that they are re-inferred by `infer_implicit_params` below
+            bi = binder_info();
+        if (is_class_out_param(binding_domain(intro_type)))
+            // out_params should always be implicit since they can be inferred from the later `c` argument
+            bi = mk_implicit_binder_info();
+        expr param = mk_local(mk_fresh_name(), binding_name(intro_type), binding_domain(intro_type), bi);
```
leodemoura: Great!

src/frontends/lean/structure_cmd.cpp
```diff
@@ -1016,6 +1016,14 @@ struct structure_cmd_fn {
...
     expr mk_intro_type() {
+        buffer<expr> params;
+        for (auto param : m_params) {
+            auto type = mlocal_type(param);
+            if (is_class_out_param(type))
+                // hide `out_param`
+                param = update_local(param, app_arg(type), local_info(param));
+            params.push_back(param);
+        }
         expr r = Pi(m_params, mk_intro_type_no_params(), m_p);
```
leodemoura: You remove the `out_param` here. So, why is the change at `projection.cpp` needed? I didn't check it.

tests/lean/run/subobject_from_source.lean
Kha: > I don't know what this commit is fixing.

Previously the RHS expanded to `(bar.mk (foo.mk b.a b.b) 2).foo`. Now it expands to `(bar.mk b.foo 2).foo`, which is definitionally equal to the LHS. I'll expand the comment with an example.

> To be honest, I have a hard time reviewing commits that modify visit_structure_instance.

Yeah, it's a mess. I dreaded splitting it into multiple functions because there are so many shared variables. Should I refactor it into a friend class of `elaborator` instead?

src/frontends/lean/definition_cmds.cpp
Kha: The other kernel exception I see regularly is "forgot `meta`", but that wasn't a simple 3-lines fix.

src/frontends/lean/definition_cmds.cpp
Kha: I'll improve the comment. The change hides kernel exceptions after error recovery, which do nothing but make the original error harder to see.

tests/lean/run/subobject_from_source.lean
leodemoura: > Previously the RHS expanded to (bar.mk (foo.mk b.a b.b) 2).foo. Now it expands to (bar.mk b.foo 2).foo, which is definitionally equal to the LHS. I'll expand the comment with an example.

I'm confused, in the current master (without this PR), I'm already getting `(bar.mk b.foo 2).foo`.
I just tried
```lean
def ex (b : bar) : b.foo = {c := 2, ..b}.foo := rfl

set_option pp.structure_instances false
#print ex
```
and I get
```lean
def ex : ∀ (b : bar), b.foo = (bar.mk (b.foo) 2).foo :=
λ (b : bar), rfl
```

leodemoura: > Yeah, it's a mess. I dreaded splitting it into multiple functions because there are so many shared variables. Should I refactor it into a friend class of elaborator instead?

Yes, I think a fried class is the best option is this kind of situation.

Kha: > in the current master (without this PR), I'm already getting (bar.mk b.foo 2).foo.

That's what I get for writing the test after the fix, thanks for catching that. `bar` should of course extend `foo`. Fixed.

src/frontends/lean/definition_cmds.cpp
```diff
@@ -860,12 +863,19 @@ environment single_definition_cmd_core(parser & p, decl_cmd_kind kind, cmd_meta
...
+        // If we have already logged an error during elaboration, don't
+        // bother showing the less helpful kernel exception
+        if (!lt.get().has_entry_now(is_error_message))
             p.mk_message(header_pos, ERROR).set_exception(ex).report();
-            return res;
-        } catch (...) {}
-        throw;
+        // As a last resort, try replacing the definition body with `sorry`.
+        // If that doesn't work either, just silently give up since we have
+        // already logged at least one error.
```
Kha, line 873: Comments added. Does this sufficiently motivate the `catch (...)` or should we replace it?

src/frontends/lean/structure_cmd.cpp
Kha: I removed this change (tweaking the intro rule type). It didn't work, and even if it did, it would have broken the preceding commit, where we look for `out_param`s in projection.cpp. If we still want to hide `out_param` when constructing structure instances, we can do it in the instance notation in the elaborator instead.

src/frontends/lean/definition_cmds.cpp
leodemoura: I think it is fine.

src/frontends/lean/elaborator.cpp
```diff
@@ -2884,350 +2869,405 @@ expr elaborator::instantiate_mvars(expr const & e, std::function<bool(expr const
...
+    buffer<expr> const & m_fvalues = m_info.m_field_values;
+
+    // set of fnames that have been used by `create_field_mvars`. Not using all of them is an error.
+    name_set m_fnames_used;
+    // elaborated sources
+    buffer<source> m_sources;
+
+    /* Because field default values may introduce arbitrary dependencies on other field values (not only within a structure,
+     * but even between a structure and its parent structures, in either direction), we cannot elaborate fields in a static
+     * order. Instead, we register a metavar and an elaboration function for each field (create_field_mvars).
```
Kha, line 2904: @leodemoura These registered elaboration functions were a major refactoring, but I'm very happy with the result. It reduces some horrible duplication and makes the overall approach much more general and less brittle. E.g. you could now have an inst implicit field whose resolution depends on the value of another field, if you really wanted to.

src/frontends/lean/elaborator.cpp
```diff
@@ -2884,350 +2869,405 @@ expr elaborator::instantiate_mvars(expr const & e, std::function<bool(expr const
...
-                        report_or_throw(elaborator_exception(e, sstream() << "invalid structure value {...}, field '"
-                                                                          << S_fname << "' is implicit and must not be provided"));
+                    /* implicit field */
+                    if (it != m_fnames.end()) {
+                        // avoid redundant error
+                        m_fnames_used.insert(S_fname);
+                        m_elab.report_or_throw(
+                                elaborator_exception(m_e, sstream() << "invalid structure value {...}, field '"
+                                                                  << S_fname
+                                                                  << "' is implicit and must not be provided"));
```
Kha: Btw, it wasn't clear to me why this restriction actually exists

src/frontends/lean/elaborator.cpp
```diff
@@ -2884,350 +2869,405 @@ expr elaborator::instantiate_mvars(expr const & e, std::function<bool(expr const
...
-                            return;
-                        new_new_fval = some_expr(fval);
+                    expr val_type = m_elab.infer_type(val);
+                    if (auto val2 = m_elab.ensure_has_type(val, val_type, expected_type, m_ref)) {
+                        /* Make sure mvar is assigned, even if val2 is a meta var as well.
+                         * This is important for termination and the `instantiate_mvars` call in `operator()`.
+                         * Note that `ensure_has_type` has already unified their types, so this should not result
+                         * in any missed unifications.
+                         */
+                        lean_always_assert(m_ctx.match(mvar, *val2));
```
Kha: @leodemoura I ended up using `match` after all because, as noted, the existing `instantiate_mvars` call below already depends on the helper mvars being unified away.

src/frontends/lean/elaborator.cpp
```diff
@@ -2884,350 +2869,405 @@ expr elaborator::instantiate_mvars(expr const & e, std::function<bool(expr const
...
+                                            m_elab.pp_type_mismatch(e2, c_type, *m_expected_type));
+        }
+
+        if (m_elab.m_in_pattern) {
+            /* Instantiate all helper mvars introduced by this function. This is important when elaborating patterns because
+             * all mvars left in the final expression are turned into pattern variables by `visit_equation` (see there).
+             * For example, the pattern `{a := a}` will result in the input `m_e = {a := ?a}` and result `e2 = foo.mk ?m` with
+             * the assignment `?m := ?a` from field elaboration. We want the return value to be `foo.mk ?a` regardless of
+             * whether ?a has an assignment (from a dependent pattern) or not. */
+            e2 = m_elab.instantiate_mvars(e2, [&](expr const & e) { return m_mvar2field.contains(mlocal_name(e)); });
```
Kha, line 3271: (this call)

src/frontends/lean/elaborator.cpp
leodemoura: I can't see the context. When I click "View changes", I go to a dead page. Is this still an issue?

Kha: Here is the line context: https://github.com/leanprover/lean/blob/6da6c143e696dc5a974cbc488dc531abfcca4ed9/src/frontends/lean/elaborator.cpp#L3068
I don't see an implementation reason not to support it

leodemoura: I believe this restriction is a leftover, and can be removed. AFAIR, before you wrote the new support for structure instance elaboration, structure instances would be elaborated by first converting the input pre-term into a temporary constructor application, where field values become arguments. This auxiliary pre-term would be constructed without `@` (reason: if `@` is used, then the elaborator uses a very simple elaboration procedure for the application). Thus, we did not have a "place" to put implicit arguments.

Kha: I see, that makes sense in context. I'll lift it then.

library/system/io_interface.lean
```diff
@@ -80,10 +80,8 @@ instance monad_io_is_monad (m : Type → Type → Type) (e : Type) [monad_io m]
...
 instance monad_io_is_monad_fail (m : Type → Type → Type) [monad_io m] : monad_fail (m io.error) :=
-{ fail := λ α s, monad_io.fail _ _ _ (io.error.other s),
-  ..monad_io.monad m io.error }
+{ fail := λ α s, monad_io.fail _ _ _ (io.error.other s) }
 
 instance monad_io_is_alternative (m : Type → Type → Type) [monad_io m] : alternative (m io.error) :=
 { orelse  := λ α a b, monad_io.catch _ _ _ a (λ _, b),
-  failure := λ α, monad_io.fail _ _ _ (io.error.other "failure"),
-  ..monad_io.monad m io.error }
+  failure := λ α, monad_io.fail _ _ _ (io.error.other "failure") }
```
Kha, line 87: @leodemoura Some quick examples of the change in action. The implementation is not perfectly straightforward (just like the other parts of the structure notation), but it should be worth it in practice in my opinion.

tests/lean/structure_instance_bug3.lean.expected.out
```diff
@@ -1,5 +1,26 @@
...
+
+α β : Type ?,
+f : α → β,
+x : set α
+⊢ Sort ?
+
+α β : Type ?,
+f : α → β,
+x : set α
+⊢ ?m_1
```
Kha: There is also some slight error messages regression, but the next commit removes the last one at least

library/system/io_interface.lean
leodemoura: Nice.

leodemoura:
merged

leodemoura:
I believe `visit_structure_instance_fn` is buggy.
In debug build, I'm getting an assertion violation.
On Windows, the release build crashes when compiling corelib.

Kha:
Ah, I see the Travis failure now. I usually only test the release mode locally. I'm building a debug binary now.

leodemoura:
I think I fixed it
https://github.com/leodemoura/lean/commit/327c96b46a43bb07caf87722ca1be5ae843736cb

PR #1923, opened 2018-02-08
fix warnings with VS

nunoplopes:
It's now /W2 clean.

We can't use \u with VS because the encoding is dependent on the locale. With the \xXX encoding we ensure it's UTF-8.

### DIFF
--- a/src/frontends/lean/pp.cpp
+++ b/src/frontends/lean/pp.cpp
@@ -124,7 +124,7 @@ bool pp_using_anonymous_constructor(environment const & env, name const & n) {
 }
 
 void initialize_pp() {
-    g_ellipsis_n_fmt  = new format(highlight(format("\u2026")));
+    g_ellipsis_n_fmt  = new format(highlight(format("\xE2\x80\xA6"))); //\u2026
     g_ellipsis_fmt    = new format(highlight(format("...")));
     g_placeholder_fmt = new format(highlight(format("_")));
     g_lambda_n_fmt    = new format(highlight_keyword(format("\xCE\xBB"))); //\u03BB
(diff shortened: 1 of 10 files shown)


Kha:
It's not much of a difference, but can't we use `u8"..."` literals?

nunoplopes:
> It's not much of a difference, but can't we use u8"..." literals?

In VS, yes. I thought it was not used due to some compatibility issue with older compilers?

src/frontends/lean/pp.cpp
```diff
@@ -127,13 +127,13 @@ void initialize_pp() {
     g_ellipsis_n_fmt  = new format(highlight(format("\u2026")));
     g_ellipsis_fmt    = new format(highlight(format("...")));
     g_placeholder_fmt = new format(highlight(format("_")));
-    g_lambda_n_fmt    = new format(highlight_keyword(format("\u03BB")));
+    g_lambda_n_fmt    = new format(highlight_keyword(format("\xCE\xBB"))); //\u03BB
     g_lambda_fmt      = new format(highlight_keyword(format("fun")));
```
leodemoura: @nunoplopes Can we put the unicode character (λ) instead? g++ and clang++ have support for it.
```
    g_lambda_n_fmt    = new format(highlight_keyword(format("λ")));		
```
BTW, if VS does not support it, we will have to fix other places where we have used unicode characters directly.

src/frontends/lean/pp.cpp
```diff
@@ -127,13 +127,13 @@ void initialize_pp() {
...
-    g_forall_n_fmt    = new format(highlight_keyword(format("\u2200")));
+    g_forall_n_fmt    = new format(highlight_keyword(format("\xE2\x88\x80"))); //\u2200
     g_forall_fmt      = new format(highlight_keyword(format("forall")));
     g_pi_n_fmt        = new format(highlight_keyword(format("Π")));
```
leodemoura: Here is a place where we used the unicode character directly.

nunoplopes: it works fine. I've updated the patch.

leodemoura:
merged

PR #1925, opened 2018-02-10
cleanup(init/meta/tactic) removing obsolete meta constants

kim-em:
These meta constants are not defined in the VM, and should be removed.


leodemoura:
merged

PR #1927, opened 2018-02-16
fix MSVC build

nunoplopes:
Visual Studio has a limitation on the size of thread-local storage.
This change compresses 3 thread-local strings into 1 by sharing those temporary buffers.

### DIFF
--- a/src/api/exception.h
+++ b/src/api/exception.h
@@ -11,7 +11,6 @@ Author: Leonardo de Moura
 #include "api/lean_macros.h"
 #include "api/lean_bool.h"
 #include "api/lean_exception.h"
-#include "api/exception.h"
 
 namespace lean {
 inline throwable * to_exception(lean_exception e) { return reinterpret_cast<throwable *>(e); }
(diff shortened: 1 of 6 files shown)


leodemoura:
@nunoplopes What is the limit?

BTW, I will remove these three thread local variables using a different approach.
I don't like a public `get_tls_str_buffer()`. Actually, I don't like the hack I used for the buffer, but it is a local hack. It is easier to check that it is used correctly.

leodemoura:
I merged only the first commit.
The second one was subsumed by:
3666fac95a1e400fdab459eb5c0a829632bc5c85
c35f2a2098d3e99be70729cb4418bea28d016351
a076f0003d54f293037761b146980ea0415f5f7f

nunoplopes:
LGTM, thanks!
The MSVC limit is 32 KBs (with the currently used scheme; we could switch to a more complicated scheme -- TlsAlloc() & friends, which have much higher limits).
If needed later there are a few other "easy" thread-local variables to remove, like the mpz instances.

PR #1932, opened 2018-02-22
fix memory leak in type_context.cpp

nunoplopes:


leodemoura:
This has been fixed in the `fresh_name` branch. 
https://github.com/leodemoura/lean/blob/fresh_name/src/library/abstract_context_cache.cpp#L133

So, I will not merge this PR since it is incompatible with the refactoring done there.

PR #1933, opened 2018-02-27
fix(library/tactic/simp_lemmas): avoid rewrite failure with more robust code

Kha:
The old code assumed `emetas` to be descendingly ordered by tmp idx, which is
not true for rfl lemmas.

### DIFF
--- a/src/library/tactic/simp_lemmas.cpp
+++ b/src/library/tactic/simp_lemmas.cpp
@@ -1227,7 +1227,7 @@ static bool instantiate_emetas(type_context & ctx, list<expr> const & _emetas, l
     lean_assert(emetas.size() == instances.size());
     for (unsigned i = 0; i < emetas.size(); ++i) {
         expr m = emetas[i];
-        unsigned mvar_idx = emetas.size() - 1 - i;
+        unsigned mvar_idx = to_meta_idx(m);
         expr m_type = ctx.instantiate_mvars(ctx.infer(m));
         // TODO(Leo, Daniel): do we need the following assertion?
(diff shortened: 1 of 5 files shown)


src/library/tactic/simp_lemmas.cpp
```diff
@@ -1227,7 +1227,7 @@ static bool instantiate_emetas(type_context & ctx, list<expr> const & _emetas, l
     lean_assert(emetas.size() == instances.size());
     for (unsigned i = 0; i < emetas.size(); ++i) {
         expr m = emetas[i];
-        unsigned mvar_idx = emetas.size() - 1 - i;
+        unsigned mvar_idx = to_meta_idx(m);
         expr m_type = ctx.instantiate_mvars(ctx.infer(m));
         // TODO(Leo, Daniel): do we need the following assertion?
         // lean_assert(!has_expr_metavar(m_type));
```
Kha, line 1233: @leodemoura Note, there is quite some duplication between this `instantiate_emetas` that is used in `refl_lemma_rewrite` and the "proper" one below.

Kha:
Well, this bug should only trigger for refl lemmas with unnecessary parameters (which I only noticed later for my actual example), so I suppose this is more of a refactoring than a fix.

src/library/tactic/simp_lemmas.cpp
leodemoura: I am aware of the duplication. It will be addressed when I get back to the simplifier. I had to stop the cleanup to work on the fresh name and cache problems.

leodemoura:
merged

PR #1934, opened 2018-02-27
More misc changes from the monad PR

Kha:


src/library/constructions/projection.cpp
```diff
@@ -169,13 +169,14 @@ void finalize_def_projection() {
 }
 
 environment mk_projections(environment const & env, name const & n, buffer<name> const & proj_names,
-                           implicit_infer_kind infer_k, bool inst_implicit) {
+                           buffer<implicit_infer_kind> infer_kinds, bool inst_implicit) {
     // Given an inductive datatype C A (where A represent parameters)
```
leodemoura: You are passing the buffer by value. It is missing `const &`

src/frontends/lean/elaborator.cpp
```diff
@@ -2983,7 +2979,10 @@ class visit_structure_instance_fn {
                 if (it != m_fnames.end()) {
                     /* explicitly passed field */
                     m_fnames_used.insert(S_fname);
-                    insert_field_preterm(S_fname, m_fvalues[it - m_fnames.begin()]);
+                    const expr & p = m_fvalues[it - m_fnames.begin()];
+                    m_field2elab.insert(S_fname, [=](expr const & d) {
```
leodemoura: We use `expr const &` instead of `const expr &` in the rest of the code base.

src/frontends/lean/structure_cmd.cpp
```diff
@@ -240,13 +240,15 @@ struct structure_cmd_fn {
...
-
-        field_decl(const expr & local, const optional<expr> & default_val, field_kind kind):
-                m_local(local), m_default_val(default_val), m_kind(kind),
+        expr                m_local; // name, type, and pos as an expr::local
+        optional<expr>      m_default_val;
+        field_kind          m_kind;
+        implicit_infer_kind m_infer_kind;
+        bool                m_has_new_default; // if true, (re-)declare default value in this structure
+
+        field_decl(const expr & local, const optional<expr> & default_val, field_kind kind,
```
leodemoura: `const type &` ==> `type const &`

Kha:
Fixed, thanks

PR #1937, opened 2018-03-01
AppVeyor: Support MSVC

Kha:
* I removed the mingw64 system upgrade since it's brittle and the AppVeyor base image should be good enough for our needs
* Releases are still built using mingw64
* @nunoplopes There are still some compiler errors, am I doing something wrong? https://ci.appveyor.com/project/Kha/lean/build/1.0.202/job/hyhj945xsmtq1e6c#L4205


nunoplopes:
@Kha Lean compiles fine with VS 2017. It seems AppVeyor is using the previous version (VS 2015 -- 14.0)?
Can you select a more recent version? (VS 2017 aka VS 15.x).

Kha:
@nunoplopes Thank you, now I'm down to [two errors](https://ci.appveyor.com/project/leodemoura/lean/build/1.0.2281):
```
Error copying file "C:/projects/lean/build/shell/lean.exe" to "C:/projects/lean/src/../bin/".
'C:\projects\lean\src\..\bin\lean' is not recognized as an internal or external command, operable program or batch file.
```
I'm not sure how to debug this further remotely. Have you ever tried building Lean directly from cmake?

nunoplopes:
no, I always run it from VS directly and the copy works fine.
I managed to find the problem:
The binary is created at C:\projects\lean\build\shell\Debug\lean.exe
and then the copy cmd tries to copy from C:/projects/lean/build/shell/lean.exe

What I don't know is why lean.exe ended up in a Debug sub-directory. It's not the behavior I see from VS..

Kha:
@nunoplopes  Thank you again for investigating. It looks like this is the expected behavior when using MSBuild with cmake: https://cmake.org/cmake/help/v3.0/variable/CMAKE_CFG_INTDIR.html
So I tried using NMake instead. Now it fails to run `bin\lean.exe`, apparently because the post-build event wasn't actually executed at all this time: https://ci.appveyor.com/project/leodemoura/lean/build/1.0.2285#L800

Kha:
Correction: The post-build step presumably was executed, just not logged (it's a dedicated target now). Apparently some DLL is missing:
`STATUS_DLL_NOT_FOUND             ((NTSTATUS)0xC0000135L)`
Windows helpfully shows such missing DLLs in an information dialog, which of course is not quite as helpful on AppVeyor.

nunoplopes:
I was able to reproduce the problem!
The problem is that we copy lean.exe to bin but not mpir.dll. I copied it by hand and the build continued just fine.

(FYI: VS uses ninja with cmake not msbuild)

nunoplopes:
The patch in https://github.com/leanprover/lean/pull/1941 should fix the test issue in MSVC.

Kha:
Thanks again, Nuno! The Ninja generator didn't seem available from the AppVeyor Windows cmake, so I'm sticking with NMake, which seems to work fine.

I see one last issue besides the test you fixed: Linking the tests with MSVC seems to be _reallly_ slow. We're very close to AppVeyor's 60 min time limit.

nunoplopes:
I've set up the build to do link-time optimization, that's why it's so slow :)  (incrementally is fast, but we don't have incrementally here..)
You can build in debug mode instead; should be a bit faster.

Kha:
Not sure it makes sense to do that for every single test executable :) . Well, I took the opportunity to add a debug build for mingw too.

nunoplopes:
> Not sure it makes sense to do that for every single test executable :)

Well, definitely not for the unit tests. But if you do it for the main binary, you need to do it for all..

Kha:
Okay, debug build is too slow using either compiler. Can't we use `set_target_properties(lean PROPERTIES LINK_FLAGS "/LTCG:INCREMENTAL")` (or `leanstatic`?) or something to optimize just the Lean executable?

nunoplopes:
Nop, you can't just do link-time optimization of a single binary.
The reason is that once you compile a .c file with /GL, then the .o is not binary code but rather compiler IR. So it's a all or nothing thing. We can add a flag to enable/disable LTCG if that helps..

src/CMakeLists.txt
```diff
@@ -253,8 +254,8 @@ elseif (MSVC)
     set(CMAKE_CXX_FLAGS_MINSIZEREL     "/Os /Zc:inline ${CMAKE_CXX_FLAGS_MINSIZEREL}")
     set(CMAKE_CXX_FLAGS_RELEASE        "/O2 /Oi /Oy /Zc:inline ${CMAKE_CXX_FLAGS_RELEASE}")
     set(CMAKE_CXX_FLAGS_RELWITHDEBINFO "/O2 /Oi /Zi ${CMAKE_CXX_FLAGS_RELWITHDEBINFO}")
-    set(LEAN_EXTRA_LINKER_FLAGS        "${LEAN_EXTRA_LINKER_FLAGS} /LTCG:INCREMENTAL")
```
nunoplopes, line 256: You need to remove /GL from CMAKE_CXX_FLAGS as well if you want to disable LTO.

Kha: I see, thanks. We're slowly getting there, hopefully :) .

Kha:
Looks good :tada:

PR #1941, opened 2018-03-05
make sure no one calls math.h's log2()

nunoplopes:
This fixes the bit_trick test with MSVC.
The problem happens because <math.h> gets included somehow in the test case when using MSVC headers.

### DIFF
--- a/src/tests/util/bit_tricks.cpp
+++ b/src/tests/util/bit_tricks.cpp
@@ -9,20 +9,20 @@ Author: Leonardo de Moura
 using namespace lean;
 
 static void tst1() {
-    lean_assert(log2(255) == 7);
-    lean_assert(log2(256) == 8);
-    lean_assert(log2(257) == 8);
-    lean_assert(log2(321) == 8);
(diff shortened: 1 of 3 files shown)


Kha:
The style test is failing:
```
/afs/info.uni-karlsruhe.de/user/ullrich/lean/src/util/bit_tricks.cpp:36:  Using C-style cast.  Use static_cast<int>(...) instead  [readability/casting] [4]
/afs/info.uni-karlsruhe.de/user/ullrich/lean/src/util/bit_tricks.h:12:  Using C-style cast.  Use static_cast<int>(...) instead  [readability/casting] [4]
```
Travis does not show an error message because the style_test cmdline is so huge: https://github.com/travis-ci/travis-ci/issues/8920
I added a new commit to #1937 that enables style_check under msys2, our fastest build.

nunoplopes:
ok, sorry, fixed it.  That warning is kinda bogus, though :)

Kha:
Ah, I didn't actually look at the offending line :)

leodemoura:
merged

PR #1948, opened 2018-03-11
Proper nightly builds

Kha:
Use Travis and AppVeyor cron jobs to publish properly versioned nightlies at https://github.com/leanprover/lean-nightly/releases
(cron jobs yet to be set up)

There is minimal leanpkg integration: It recognizes nightlies as separate versions, and will warn on a version mismatch. It will not help you with actually acquiring the correct release (yet).


leodemoura:
merged

PR #1960, opened 2018-06-25
fix(doc/export_format): correct #DEF example and typos

andrejtokarcik:
In 4f3e6035, the `#DEF` example was updated to reflect the command's new syntax. Unfortunately the resulting code -- being a mixture of the old example with a new updated line -- was incorrect. Hereby I replace the erroneous snippet with a proper export listing generated from the following Lean file:
```
prelude
universes l
def id {A : Sort l} (a : A) : A := a
```

I also include a few additional minor fixes (typos, style).

### DIFF
--- a/doc/export_format.md
+++ b/doc/export_format.md
@@ -20,7 +20,7 @@ Hierarchical names
 ------------------
 
 A hierarchical name is essentially a list of strings and integers.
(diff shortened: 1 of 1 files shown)


leodemoura:
This repo is frozen. We are working on Lean4

PR #1976, opened 2018-10-07
fix(lean/bin/leanpkg*): handle spaces in paths

### LINKED ISSUE #1973: leanpkg doesn't work on Windows when lean is copied to C:\Program Files\lean\
### Prerequisites
1. install/copy Lean in C:\Program Files\lean\
2. add it to PATH

### Description

leanpkg doesn't work on Windows when lean is copied to C:\Program Files\lean\

### Steps to Reproduce

1. run: leanpkg from PS or CMD

**Expected behavior:** [What you expect to happen]
leanpkg to run correctly on Windows under normal installation place for that OS

**Actual behavior:** [What actually happens]
```
C:\Users\Vassil>leanpkg
'Files\lean\bin\..' is not recognized as an internal or external command,
operable program or batch file.
C:\Program:1:0: error: file 'C:\Program' not found
<unknown>:1:1: error: file 'C:\Program' not found
```

**Reproduces how often:** [What percentage of the time does it reproduce?]

### Versions

You can get this information from copy and pasting the output of `lean --version`,
please include the OS and what version of the OS you're running.
```Lean (version 3.4.1, commit 17fe3decaf8a, Release)```

### Additional Information

Any additional information, configuration or data that might be necessary to reproduce the issue.

solson:
Fixes #1973. I've tested it locally and this appears to be enough to solve the problem.

I'm unsure if this falls within the scope of the "must include tests" contributing guideline, but I suppose we could make AppVeyor intentionally put Lean into a directory with spaces?

### DIFF
--- a/bin/leanpkg
+++ b/bin/leanpkg
@@ -14,12 +14,12 @@ else
     READLINK=readlink
 fi
 
-leandir=$(dirname $($READLINK -f $0))/..
-leandir=$($READLINK -f $leandir)
+leandir=$(dirname "$($READLINK -f "$0")")/..
+leandir=$($READLINK -f "$leandir")
 
 librarydir="$leandir/lib/lean"
 test -d "$librarydir" || librarydir="$leandir"
 
 LEAN_PATH=$librarydir/library:$librarydir/leanpkg \
   PATH=$leandir/bin:$PATH \
-  exec lean --run $librarydir/leanpkg/leanpkg/main.lean "$@"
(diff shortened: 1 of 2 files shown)


solson:
Since Lean 3.4.1 is described as the last release for Lean 3.x, @gebner has suggested we could add a patch for leanpkg.bat to [elan](https://github.com/Kha/elan) and/or the [VSCode extension](https://github.com/leanprover/vscode-lean) instead.

solson:
@rwbarton added a commit to also fix handling of spaces for the shell script.

PR #1981, opened 2018-11-12
feat(leanpkg): add git branch tracking for dependencies

khoek:
Add branch tracking for git dependencies. There is no functional change to currently existing `leanpkg.toml` manifests. In a git dependency an optional `branch="..."` (along with the required `git=...` and `rev=...`) is now permitted, which will cause subsequent invocations of `leanpkg upgrade` to track that branch, as opposed to a default determined by the Lean version.

We opt to store the branch name inside an `option string`, instead of just picking a default the first time `leanpkg.toml` is read with it missing, since then when `leanpkg.toml` is re-written unexpected `branch = "..."` entries will not appear (as was the case with the old behaviour).

### DIFF
--- a/leanpkg/leanpkg/git.lean
+++ b/leanpkg/leanpkg/git.lean
@@ -13,6 +13,10 @@ if lean.is_release then
 else
     "master"
 
+def git_default_revision : option string → string
+| none := upstream_git_branch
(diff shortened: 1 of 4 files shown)


Kha:
Looks great, thanks!

PR #1986, opened 2019-01-10
fix(library/module_mgr): ignore '\r' changes

gebner:
(As reported and complained about by @digama0, @ChrisHughes24, @kckennylau in Amsterdam.)

Apparently, on windows vscode changes the line-endings of a file when sending it to the server (as compared to the on-disk copy).  This seems to happen when the on-disk file has CR-LF endings.  Then the server thinks the file has changed and will recompile everything.

As a workaround, we just ignore CR when checking whether a file is changed.  (This doesn't affect command-line usage, which depends on timestamps.)

### DIFF
--- a/src/library/module_mgr.cpp
+++ b/src/library/module_mgr.cpp
@@ -353,6 +353,16 @@ module_mgr::build_lean_snapshots(std::shared_ptr<module_parser> const & mod_pars
     }
 }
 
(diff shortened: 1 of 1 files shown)


gebner:
(We've tested this workaround on @robertylewis's laptop and it works.)

kckennylau:
I second this PR.

PR #1994, opened 2019-03-12
Add inductive no confusion

digama0:
Adds the `add_ginductive` hook for producing nested/mutual inductive types, as well as regular inductive types with all the fancy features you normally get from the `inductive` command (by contrast to `add_inductive` which only gives you a really basic inductive with no lemmas that doesn't work with the equation compiler or `induction` tactic).


digama0:
wrong repo...

PR #1998, opened 2019-04-10
Fix a failing test (one more still fails, but you have a [TODO] in it)

khoek:


khoek:
(Sorry, this was meant for EdAyers.)

